Issue #5, opened 2018-06-12
Create manifests to deploy CCM as pod or daemonset


frapposelli:
For reference: https://github.com/kubernetes/cloud-provider-openstack/tree/master/manifests/controller-manager

frapposelli:
Initial stab at manifest under review as #14

dvonthenen:
Will deploy and give this a try and provide feedback.

dvonthenen:
I believe I have this working now for both a pod/deployment and DaemonSet with proper RBAC settings. Will have a PR in a bit.

dvonthenen:
PR: https://github.com/kubernetes/cloud-provider-vsphere/pull/20

PR #18, opened 2018-07-23
RBAC for DS and Pod Deployment of CCM

dvonthenen:
Adds a simple pod-based deployment instead of using a DaemonSet.
Provider roles and bindings for RBAC that support both the DS and POD deployment.

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/5


dvonthenen:
Will re-open a new PR with the correct GitHub email address.

PR #20, opened 2018-07-23
RBAC for DS and Pod Deployment of CCM

dvonthenen:
Adds a simple pod-based deployment instead of using a DaemonSet.
Provider roles and bindings for RBAC that support both the DS and POD deployment.

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/5

### DIFF
--- a/manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml
+++ b/manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml
@@ -21,7 +21,7 @@ spec:
       image: docker.io/vmware/vsphere-cloud-controller-manager:latest
       args:
         - /bin/vsphere-cloud-controller-manager
-        - --v=4
+        - --v=2
         - --cloud-config=/etc/cloud/vsphere.conf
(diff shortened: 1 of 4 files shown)


manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml
```diff
@@ -0,0 +1,71 @@
...
+    tier: control-plane
+  name: vsphere-cloud-controller-manager
+  namespace: kube-system
+spec:
+  containers:
+    - name: vsphere-cloud-controller-manager
+      image: docker.io/vmware/vsphere-cloud-controller-manager:latest
+      args:
+        - /bin/vsphere-cloud-controller-manager
+        - --v=4
```
frapposelli: Probably having `--v=2` is a better option for generic deployments as 4 is debug logging

frapposelli:
Great stuff @dvonthenen, just a minor nit in review, and can you please add a small doc under `docs/` with instructions on how to deploy? thanks!

Issue #17, opened 2018-07-20
Allow Pushing Images with Latest Tag


dvonthenen:
**Is this a BUG REPORT or FEATURE REQUEST?**:
 /kind feature

**What happened**:
When running `make upload-images`, docker images are pushed and tags based on the build hash.

**What you expected to happen**:
Having a tag on the build hash is fine, but I would also like to push with the tag `latest`. By default, I would be nice to push both the build tag and latest tag. Not ideal but acceptable would be optionally pushing a `latest` tag.

The reason why this is needed is because by not pushing with the `latest` tag means for each build I need to update my YAML to append the build tag (example `12a86808`) to the image name (ie `dvonthenen/vsphere-cloud-controller-manager:12a86808`).

**How to reproduce it (as minimally and precisely as possible)**:
Run `make upload-images`

**Anything else we need to know?**:
Nope

**Environment**:
- vsphere-cloud-controller-manager version: Based on https://github.com/kubernetes/cloud-provider-vsphere/pull/14/files
- OS (e.g. from /etc/os-release): CentOS 7.X
- Kernel (e.g. `uname -a`): meh
- Install tools:
- Others: Kubernetes 1.11.1

PR #22, opened 2018-07-24
Allow using latest tag for docker push

dvonthenen:
The reason why this is needed is by not pushing with the latest tag means for each build I need to update my YAML to append the build tag (example 12a86808) to the image name (ie dvonthenen/vsphere-cloud-controller-manager:12a86808). Having latest allows users to deploy the latest cloud provider without having to modify the YAML file in the manifest directory every single time.

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/17


frapposelli:
🐑 🇮🇹 

/lgtm
/approve

Issue #6, opened 2018-06-19
Enable the use of Kubernetes secrets to store passwords for CCM


frapposelli:
We should allow users to store infrastructure secrets in the right place.

/kind feature

dvonthenen:
I am actually looking at the code and the functionality is mostly there with a couple of (significant) caveats:
1. The functionality is disabled currently (hence the significant caveat).
2. I don't really agree with the order precedence for what credentials are used to establish a connection.

Going to re-enable it (see if what is implemented actually works) and change the precedence.

dvonthenen:
PR: https://github.com/kubernetes/cloud-provider-vsphere/pull/33

PR #33, opened 2018-08-03
Implement Kubernetes Secret for Storing vCenter Creds

dvonthenen:
This implements the ability to store vCenter credentials (username and password) in a Kubernetes Secret. 
- Multiple vCenter creds are supported within the secret.
- Configurable service account supported for multiple cloud-provider-vsphere instances.
- To maintain backward compatibility, credentials are obtained in the following order: secret, vCenter unique creds, global setting. If the secret doesn't exist, the behavior is to fall back on the old method of conf file.
- Documentation updated to help the user deploy using a secret.

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/6

Tested the following scenarios on a 1.11.1 Kubernetes cluster:
- Secret containing vCenter creds without setting creds in vsphere.conf
- Secret containing vCenter creds and also setting creds in vsphere.conf (made sure Secret is used)
- Secret does not contain vCenter creds but creds set in vsphere.conf (creds from config are used)
- No Secret set, vCenter creds set within [vcenter ip] block
- No Secret set, vCenter creds set using Global

**Release note**:
None

### DIFF
--- a/manifests/controller-manager/vccm-secret.yaml
+++ b/manifests/controller-manager/vccm-secret.yaml
@@ -4,5 +4,5 @@ metadata:
   name: vccm
   namespace: kube-system
 data:
-  1.2.3.4.username: "Replace with out from `echo -n YOUR_VCENTER_USERNAME | base64`"
-  1.2.3.4.password: "Replace with out from `echo -n YOUR_VCENTER_PASSWORD | base64`"
+  1.2.3.4.username: "Replace with output from `echo -n YOUR_VCENTER_USERNAME | base64`"
+  1.2.3.4.password: "Replace with output from `echo -n YOUR_VCENTER_PASSWORD | base64`"
--- a/manifests/controller-manager/vccm-secret.yaml
+++ b/manifests/controller-manager/vccm-secret.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: vccm
+  namespace: kube-system
+data:
+  1.2.3.4.username: "Replace with out from `echo -n YOUR_VCENTER_USERNAME | base64`"
+  1.2.3.4.password: "Replace with out from `echo -n YOUR_VCENTER_PASSWORD | base64`"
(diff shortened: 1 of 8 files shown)


frapposelli:
Awesome stuff, thanks @dvonthenen 

/ok-to-test

dvonthenen:
@figo I would have to investigate how to go about having unit tests for this change. The code in the PR interacts mostly with pulling a secret from kubernetes, initializing a Kubernetes client using a service account, deleting existing code in order to get backward compatibility, and documentation updates. Not sure if there is anything that mocks functionality of a kubernetes cluster like secret management.

manifests/controller-manager/vccm-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: vccm
+  namespace: kube-system
+data:
+  1.2.3.4.username: "Replace with output from `echo -n YOUR_VCENTER_USERNAME | base64`"
```
frapposelli: Any specific reasons why we should keep the username and password encoded in base64? I believe this was added originally to provide some obfuscation for credentials because they were stored in a configuration file on the host, ideally using k8s secrets should not require this obfuscation anymore.

dvonthenen: @frapposelli unless I missed the option for it (entirely possible), I thought secrets are still required to be base64 encoded. Using my 1.11.1 cluster, I attempted to create a secret without base64 encoding, this is the error you get:

```
[vonthd@k8master Documents]$ kubectl create -f vccm-secret-nobase64.yaml 
Error from server (BadRequest): error when creating "vccm-secret-nobase64.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 5, error found in #10 byte of ...|"XXXXX","10.160.2|..., bigger context ...|:"v1","data":{"10.160.221.107.password":"XXXXXX","10.160.221.107.username":"administrator@vsphere.|...
```

frapposelli: You're absolutely right :)

PR #44, opened 2018-08-23
Feature: Implement vcpctl create solution user for CCM

### LINKED ISSUE #15: Create vcpctl tool to facilitate CCM provisioning
**FEATURE REQUEST**:

>
> /kind feature

Deploying a cloud provider on vSphere is a task that has many prerequisites, from creating a user with correct roles on vCenter, to creating a correct configuration for the service. When migrating users from the in-tree version there's also a need to convert configuration files and to make sure sensitive credentials are stored safely.

The tool should fulfill these needs:

- [ ] Perform vSphere configuration health check:
  - disk.uuid set on esx
  - network reachability
- [ ] Create vSphere role with minimal set of permissions
- [x] Create vSphere solution user (generate keypair), to be used with CCM
- [ ] Convert old in-tree `vsphere.conf` configuration files to new configmap

fanzhangio:
- Add Client for delegating vim25 and ssoadmin
- Add Credential for client, session and sso
- Add CreateUserFunc for creating solution user and person user
- Implement creating default solution user and granting WSTrust
permission and Administrator role.

fixes #15 

Stage 2 for **vcpctl** implementation

**Release note**:
`NONE`

```release-note
none
```

### DIFF
--- a/cmd/vcpctl/provision/provision.go
+++ b/cmd/vcpctl/provision/provision.go
@@ -17,9 +17,12 @@ limitations under the License.
 package provision
 
 import (
+	"context"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
+	"k8s.io/cloud-provider-vsphere/pkg/cli"
 )
 
 var (
@@ -43,6 +46,10 @@ var (
 	vcUser string
 	// vCenter password in clear text.
 	vcPassword string
+	// vCenter certificate
+	vcCert string
+	// vcRole is role for solution user (Default is Administrator)
+	vcRole string
 )
 
 var provisionCmd = &cobra.Command{
@@ -60,16 +67,18 @@ var provisionCmd = &cobra.Command{
 }
 
 func AddProvision(cmd *cobra.Command) {
-	provisionCmd.Flags().StringVar(&configFile, "config", "", "vsphere cloud provider config file path")
-	provisionCmd.Flags().BoolVar(&oldInTreeConfig, "oldConfig", false, "old int-tree vsphere configuration file, true or false")
-	provisionCmd.Flags().BoolVar(&interactive, "interactive", true, "specify interactive mode (true) as default, set (false) for declarative mode for automation")
 
-	provisionCmd.Flags().StringVar(&vchost, "vchost", "", "specify vCenter IP ")
-	provisionCmd.Flags().StringVar(&vcport, "vcport", "", "specify vCenter Port ")
-	provisionCmd.Flags().StringVar(&vcUser, "vcuser", "", "specify vCenter user ")
-	provisionCmd.Flags().StringVar(&vcPassword, "vcpassword", "", "specify vCenter Password ")
+	provisionCmd.Flags().StringVar(&configFile, "config", "", "VSphere cloud provider config file path")
+	provisionCmd.Flags().BoolVar(&oldInTreeConfig, "oldConfig", false, "Old int-tree vsphere configuration file, true or false")
+	provisionCmd.Flags().BoolVar(&interactive, "interactive", true, "Specify interactive mode (true) as default, set (false) for declarative mode for automation")
+
+	provisionCmd.Flags().StringVar(&vchost, "host", "", "Specify vCenter IP")
+	provisionCmd.Flags().StringVar(&vcport, "port", "", "Specify vCenter Port")
+	provisionCmd.Flags().StringVar(&vcUser, "user", "", "Specify vCenter user")
+	provisionCmd.Flags().StringVar(&vcPassword, "password", "", "Specify vCenter Password")
 	provisionCmd.Flags().BoolVar(&insecure, "insecure", false, "Don't verify the server's certificate chain")
-	provisionCmd.Flags()
+	provisionCmd.Flags().StringVar(&vcCert, "cert", "", "Certificate for solution user")
+	provisionCmd.Flags().StringVar(&vcRole, "role", "Administrator", "Role for solution user (RegularUser|Administrator)")
 
 	cmd.AddCommand(provisionCmd)
 }
@@ -77,5 +86,14 @@ func AddProvision(cmd *cobra.Command) {
 func RunProvision(cmd *cobra.Command, args []string) {
 	// TODO (fanz): implement provision
 	fmt.Println("Perform cloud provider provisioning...")
-
+	o := cli.ClientOption{}
+	o.LoadCredential(vcUser, vcPassword, vcCert, vcRole)
+	ctx := context.Background()
+	client, err := o.NewClient(ctx, vchost)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
+	}
+	o.Client = client
+	cli.CreateSolutionUser(ctx, &o)
 }
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -17,10 +17,14 @@ limitations under the License.
 package cli
 
 import (
+	"context"
+	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 	"os"
 
-	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/ssoadmin"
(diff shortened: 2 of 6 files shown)


pkg/cli/cli.go
```diff
@@ -53,9 +57,34 @@ func CreateRole() error {
...
+	return u.Run(ctx, o, CreateUserFunc(func(c *ssoadmin.Client) error {
+		if cert, err := ReadContent(o.getCredential().getCertificate()); err == nil {
+			block, _ := pem.Decode([]byte(cert))
+			if block != nil {
+				u.solution.Certificate = base64.StdEncoding.EncodeToString(block.Bytes)
+			}
+			u.solution.Certificate = cert
+		}
+		if u.solution.Certificate == "" {
+			return fmt.Errorf("Need solution user certificate (--cert) to create solution user\n")
```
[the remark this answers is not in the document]
fanzhangio: A vCenter Solution User can only use public key authentication via SAML, a Solution User does not have a plain text password. So the SSO_TOKEN needs to be provided (by ENV currently). This `solution.Certificate` contains the cert data decoded from crt file. 
In command line, the cert file need to be passed to `--cert` flag. 
```bash
--cert  k8s-vcp.crt
```
It stores the raw data in ClientOption, and decoded into `User.solution.Certificate`

pkg/cli/util.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, v := range section {
+		if v == "https" || v == "http" {
+			continue
+		}
+		ip = net.ParseIP(v)
+		if ip != nil {
+			return "https://username:password@" + ip.String() + "/sdk"
+		}
+	}
+	return "https://username:password@host/sdk"
```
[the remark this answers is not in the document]
fanzhangio: Well, the basic idea is to initialize `UserInfo` with `username` and `password`, since the it is always required for calling rest api. In the following `processCredential()` call  the actual user and password will be loaded and overwrite. The `processCredential` call will handle the credentials from secret in the future.

pkg/cli/util.go
```diff
@@ -0,0 +1,58 @@
...
+	ip := net.ParseIP(host)
+	if ip != nil {
+		return "https://username:password@" + ip.String() + "/sdk"
+	}
+	section := strings.Split(host, ":")
+	for _, v := range section {
+		if v == "https" || v == "http" {
+			continue
+		}
+		ip = net.ParseIP(v)
```
fanzhangio: There is a small bug in `for range loop`, where token v should be trimmed by `//`
```golang
v = strings.Trim(v, "//")
```

pkg/cli/util.go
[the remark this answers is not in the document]
fanzhangio: The basic idea is to initialize `UserInfo` with `username` and `password`. The following 
 `processCredential()` call will overwrite the actual credentials based on the `Credential` strategy. 
```golang
type Credential struct {
	username string
	password string
	cert     string
	Secrete  VCCMSecret
}
```
1) declarative username and password
2) Secrete in k8s

cmd/vcpctl/provision/provision.go
```diff
@@ -60,22 +65,32 @@ var provisionCmd = &cobra.Command{
...
+
 	provisionCmd.Flags().StringVar(&configFile, "config", "", "vsphere cloud provider config file path")
 	provisionCmd.Flags().BoolVar(&oldInTreeConfig, "oldConfig", false, "old int-tree vsphere configuration file, true or false")
 	provisionCmd.Flags().BoolVar(&interactive, "interactive", true, "specify interactive mode (true) as default, set (false) for declarative mode for automation")
 
-	provisionCmd.Flags().StringVar(&vchost, "vchost", "", "specify vCenter IP ")
-	provisionCmd.Flags().StringVar(&vcport, "vcport", "", "specify vCenter Port ")
-	provisionCmd.Flags().StringVar(&vcUser, "vcuser", "", "specify vCenter user ")
-	provisionCmd.Flags().StringVar(&vcPassword, "vcpassword", "", "specify vCenter Password ")
+	provisionCmd.Flags().StringVar(&vchost, "host", "", "specify vCenter IP ")
```
dougm: Flag descriptions should start with an uppercase letter, and not have a trailing space.

pkg/cli/cli.go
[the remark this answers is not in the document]
dougm: We could add an option to generate the key pair too, can wait for a later PR though.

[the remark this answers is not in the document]
dougm: Trailing `\n` should be removed from the error message.

pkg/cli/cli.go
```diff
@@ -53,9 +57,34 @@ func CreateRole() error {
...
+		}
+		u.solution.Description = u.AdminPersonDetails.Description
+		if err := c.CreateSolutionUser(ctx, u.id, u.solution); err != nil {
+			return err
+		}
+		p := types.PrincipalId{Name: "k8s-vcp", Domain: c.Domain}
+
+		if _, err := c.SetRole(ctx, p, "Administrator"); err != nil {
```
dougm: Role should probably be a flag, we can default to `Administrator`.  A separate feature for vcpctl could be to create a role with the minimum set of privileges required by VCP.

pkg/cli/client.go
```diff
@@ -18,20 +18,131 @@ package cli
...
+
+func (o *ClientOption) getCredential() Credential {
+	return o.credential
+}
+
+type Credential struct {
+	username string
+	password string
+	cert     string
+	Secrete  VCCMSecret
```
dougm: `Secrete` -> `Secret`

pkg/cli/client.go
```diff
@@ -18,20 +18,131 @@ package cli
...
+func (o *ClientOption) LoadCredential(username, password, cert string) {
+	c := Credential{}
+	c.setUsername(username)
+	c.setPassowrd(password)
+	c.setCertificate(cert)
+	// TODO: (fanz) Secret of Credential
+	o.credential = c
+}
+
+func (c Credential) setPassowrd(p string) {
```
dougm: `setPassowrd` -> `setPassword`

dougm: Why the getters and setters here, rather than just use the fields directly?

pkg/cli/config.go
```diff
@@ -22,6 +22,10 @@ import (
 	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere"
 )
 
+// Config represents configuration of cpv
```
dougm: `cpv` -> `vcp`

pkg/cli/user.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+
+	ssoClient, err := ssoadmin.NewClient(ctx, vc)
+	if err != nil {
+		return err
+	}
+
+	// SSO admin server has its own session manager. There is no SSO token persistence yet, so just use an env
```
dougm: The persistence comment only applied to govc - sessions are persisted by default by govc in `~/.govmomi/sessions`

pkg/cli/util.go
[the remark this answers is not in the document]
dougm: `soap.ParseURL` can be used here instead?  ok to initialize `UserInfo`, but it should be with empty `""` username and password.

pkg/cli/cli.go
fanzhangio: Yes, added a flag for role (default is Administrator). 
`Create Role` could be in separate PR.

pkg/cli/client.go
fanzhangio: Done

pkg/cli/client.go
fanzhangio: Removed setters and getters.

pkg/cli/config.go
fanzhangio: Done.

cmd/vcpctl/provision/provision.go
fanzhangio: done

frapposelli:
/lgtm
/approve

👍

PR #45, opened 2018-09-10
Add NodeManager Unit Tests

dvonthenen:
Continue to add unit tests to the project.

Moved around Config initialization code to NodeManager to satisfy test order dependencies

**Release note**:
None

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager_test.go
+++ b/pkg/cloudprovider/vsphere/nodemanager_test.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsphere
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestRegUnregNode(t *testing.T) {
+	cfg, ok := configFromEnvOrSim()
+	defer ok()
+
+	vsphere, err := newVSphere(cfg)
+	if err != nil {
+		t.Fatalf("Failed to construct/authenticate vSphere: %s", err)
+	}
+
+	nm := NodeManager{
+		vsphereInstanceMap: vsphere.vsphereInstanceMap,
+		nodeNameMap:        make(map[string]*NodeInfo),
+		nodeUUIDMap:        make(map[string]*NodeInfo),
+		nodeRegUUIDMap:     make(map[string]*v1.Node),
+	}
+
+	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
+	name := vm.Config.GuestFullName
+	UUID := vm.Config.Uuid
+	k8sUUID := nm.convertK8sUUIDtoNormal(UUID)
+
+	node := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Status: v1.NodeStatus{
+			NodeInfo: v1.NodeSystemInfo{
+				SystemUUID: k8sUUID,
+			},
+		},
+	}
+
+	nm.RegisterNode(node)
+
+	if len(nm.nodeNameMap) != 1 {
+		t.Errorf("Failed: nodeNameMap should be a length of 1")
+	}
+	if len(nm.nodeUUIDMap) != 1 {
+		t.Errorf("Failed: nodeUUIDMap should be a length of  1")
+	}
+	if len(nm.nodeRegUUIDMap) != 1 {
+		t.Errorf("Failed: nodeRegUUIDMap should be a length of  1")
+	}
+
+	nm.UnregisterNode(node)
+
+	if len(nm.nodeNameMap) != 1 {
+		t.Errorf("Failed: nodeNameMap should be a length of  1")
+	}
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	nm := NodeManager{
+		vsphereInstanceMap: vsphere.vsphereInstanceMap,
+		nodeNameMap:        make(map[string]*NodeInfo),
+		nodeUUIDMap:        make(map[string]*NodeInfo),
+		nodeRegUUIDMap:     make(map[string]*v1.Node),
+	}
+
+	name := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine).Config.GuestFullName
```
dougm: Since you're grabbing multiple fields, should use the same vm for both:
```go
vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
name := vm.Config.GuestFullName
id := vm.Config.Uuid
```

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	if len(nm.nodeUUIDMap) != 1 {
+		t.Errorf("Failed: nodeUUIDMap should be a length of  1")
+	}
+	if len(nm.nodeRegUUIDMap) != 0 {
+		t.Errorf("Failed: nodeRegUUIDMap should be a length of 0")
+	}
+}
+
+//Sim error: SearchIndex:SearchIndex does not implement: FindByDnsName
```
dougm: We can implement that method in vcsim.  Can you open an issue in the govmomi repo?

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -58,105 +47,6 @@ bGWObufAR8Ni4QIgWpILjW8dkGg8GOUZ0zaNA6Nvt6TIv2UWGJ4v5PoV98kCIQDx
 rIiZs5QbKdycsv9gQJzwQAogC8o04X3Zz3dsoX+h4A==
 -----END RSA PRIVATE KEY-----`
 
-func configFromEnv() (cfg Config, ok bool) {
```
dougm: Looks like these configFrom* methods can stay in vsphere_test.go rather than move to node_manager_test.go.. They are also used in vsphere_test.go and that seems like a better place for the general purpose test helper functions.

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: Ahhhh I didn't realize that would grab multiple VM instances per call. Will update.

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: Will do!

dvonthenen: https://github.com/vmware/govmomi/issues/1231

pkg/cloudprovider/vsphere/vsphere_test.go
dvonthenen: I think I know what I did... I was testing just the unit tests within nodemanager_test.go and it was complaining that the methods were missing. Will move them back.

dougm:
@dvonthenen we can implement a `FindByDnsName` in the meantime, based on: https://godoc.org/github.com/vmware/govmomi/simulator#example-package

Here's an example for SearchIndex that should work in these tests:
```go
package simulator_test

import (
	"context"
	"testing"

	"github.com/vmware/govmomi"
	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/simulator"
	"github.com/vmware/govmomi/vim25/methods"
	"github.com/vmware/govmomi/vim25/soap"
	"github.com/vmware/govmomi/vim25/types"
)

type SearchIndex struct {
	*simulator.SearchIndex
	vm *simulator.VirtualMachine
}

func (s *SearchIndex) FindByDnsName(req *types.FindByDnsName) soap.HasFault {
	res := &methods.FindByDnsNameBody{Res: new(types.FindByDnsNameResponse)}
	if req.VmSearch && req.DnsName == s.vm.Name {
		res.Res.Returnval = &s.vm.Self
	}
	return res
}

func TestFindByDNS(t *testing.T) {
	ctx := context.Background()
	model := simulator.VPX()

	defer model.Remove()
	err := model.Create()
	if err != nil {
		t.Fatal(err)
	}

	s := model.Service.NewServer()
	defer s.Close()

	c, err := govmomi.NewClient(ctx, s.URL, true)
	if err != nil {
		t.Fatal(err)
	}

	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)

	search := object.NewSearchIndex(c.Client)
	si := simulator.Map.Get(search.Reference()).(*simulator.SearchIndex)
	simulator.Map.Put(&SearchIndex{si, vm})

	for _, name := range []string{"nope", vm.Name} {
		ref, err := search.FindByDnsName(ctx, nil, name, true)
		if err != nil {
			t.Fatal(err)
		}
		t.Logf("FindByDnsName(%s): %s", name, ref)
	}
}
```

dougm:
@dvonthenen do you want to integrate with this PR or a follow up PR?  Up to you.  Should just need to copy the `SearchIndex` type + method, then add these lines to your test:
```go
        search := object.NewSearchIndex(c.Client)
	si := simulator.Map.Get(search.Reference()).(*simulator.SearchIndex)
	simulator.Map.Put(&SearchIndex{si, vm})
```

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -0,0 +1,174 @@
...
+		nodeRegUUIDMap:     make(map[string]*v1.Node),
+	}
+
+	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
+	name := vm.Config.GuestFullName
+
+	var cancel context.CancelFunc
+	for _, instance := range vsphere.vsphereInstanceMap {
+		var ctx context.Context
+		ctx, cancel = context.WithCancel(context.Background())
```
dougm: No need for `WithCancel` here.  Also note in general you'd want to `defer cancel()` here, rather than outside the loop.

dougm: You can also use the instance directly as the other tests do:
```go
	instance, ok := vsphere.vsphereInstanceMap[cfg.Global.VCenterIP]
	if !ok {
		t.Fatalf("Couldn't get vSphere instance: %s", cfg.Global.VCenterIP)
	}
```

dvonthenen:
My bad... I assumed the update to that file was due to the addition of vcsim stuff and I didn't check the changes since the file in the review isn't automatically expanded to show the changes. I will be more cautious in the future with changes made to that file.

PR #46, opened 2018-09-11
Feature: Create role and check vms on vSphere for enabling UUID (Stage 3 and 4)

fanzhangio:
- Add create default role with minimal privileges
- Traversal vms on vsphere, checking UUID and enable it if not.

#15

### DIFF
--- a/cmd/vcpctl/provision/provision.go
+++ b/cmd/vcpctl/provision/provision.go
@@ -58,7 +58,7 @@ var provisionCmd = &cobra.Command{
 	Long: `Starting prerequisites for deploying a cloud provider on vSphere, in cluding :
 	[x] vSphere configuration health check.
 	[x] Create vSphere solution user.
-	[x] Create vSohere role with minimal set of premissions.
+	[x] Create vSphere role with minimal set of permissions.
   `,
 	Example: `# Specify interaction mode or declaration mode (default)
 	vcpctl provision --interactive=false
@@ -94,6 +94,13 @@ func RunProvision(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		os.Exit(1)
 	}
+	defer client.Logout(ctx)
 	o.Client = client
+	fmt.Println("Create solution user...")
 	cli.CreateSolutionUser(ctx, &o)
+	r := cli.Role{RoleName: "k8s-vcp-default"}
+	fmt.Println("Create default role with minimal permissions...")
+	cli.CreateRole(ctx, &o, &r)
+	fmt.Println("Checking vSphere Config on VMs...")
+	cli.CheckVSphereConfig(ctx, &o)
 }
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -23,8 +23,12 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/ssoadmin"
 	"github.com/vmware/govmomi/ssoadmin/types"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25/mo"
+	vimType "github.com/vmware/govmomi/vim25/types"
 	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere"
 )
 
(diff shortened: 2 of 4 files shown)


cmd/vcpctl/provision/provision.go
```diff
@@ -94,6 +94,13 @@ func RunProvision(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		os.Exit(1)
 	}
+	defer client.Logout(ctx)
 	o.Client = client
+	fmt.Println("Create solution user...")
 	cli.CreateSolutionUser(ctx, &o)
+	r := cli.Role{RoleName: "default"}
+	fmt.Println("Create default role with minimal premissions...")
```
dougm: "permissions"

cmd/vcpctl/provision/provision.go
```diff
@@ -94,6 +94,13 @@ func RunProvision(cmd *cobra.Command, args []string) {
...
+	cli.CreateRole(ctx, &o, &r)
+	fmt.Println("Check vSphere Config on VMs...")
```
dougm: "Checking"

pkg/cli/cli.go
```diff
@@ -47,14 +52,61 @@ func ParseConfig(configFile string) (vsphere.Config, error) {
...
+
+	f := find.NewFinder(c, true)
+	err = cv.Retrieve(ctx, []string{"VirtualMachine"}, []string{"summary"}, &vms)
+	if err != nil {
+		return err
+	}
+	for _, v := range vms {
+		if v.Summary.Config.Uuid == "" {
+			name := v.Summary.Config.Name
+			vm, err = f.VirtualMachine(ctx, name)
```
dougm: No need to use `Finder` here, instead you can use: `object.NewVirtualMachine(c, v.Reference())`

pkg/cli/cli.go
```diff
@@ -47,14 +52,61 @@ func ParseConfig(configFile string) (vsphere.Config, error) {
...
+			if err != nil {
+				return err
+			}
+
+			spec := vimType.VirtualMachineConfigSpec{
+				ExtraConfig: config,
+			}
+			task, err := vm.Reconfigure(ctx, spec)
```
dougm: With the ContainerView selecting all VMs in the inventory, this means we're Reconfiguring every VM when we should only do this for k8s nodes.  I believe the `NodeManager` related code can help to find the k8s VMs, @dvonthenen may have pointers.

cmd/vcpctl/provision/provision.go
```diff
@@ -94,6 +94,13 @@ func RunProvision(cmd *cobra.Command, args []string) {
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		os.Exit(1)
 	}
+	defer client.Logout(ctx)
 	o.Client = client
+	fmt.Println("Create solution user...")
 	cli.CreateSolutionUser(ctx, &o)
+	r := cli.Role{RoleName: "default"}
```
dougm: The role name should probably have a `k8s-vcp-` prefix or similar, so it is easier to identify from the UI, etc.

cmd/vcpctl/provision/provision.go
```diff
@@ -58,7 +58,7 @@ var provisionCmd = &cobra.Command{
 	Long: `Starting prerequisites for deploying a cloud provider on vSphere, in cluding :
 	[x] vSphere configuration health check.
 	[x] Create vSphere solution user.
-	[x] Create vSohere role with minimal set of premissions.
+	[x] Create vSphere role with minimal set of premissions.
```
dougm: Also "permissions"

pkg/cli/user.go
```diff
@@ -80,3 +82,27 @@ func (u *User) Run(ctx context.Context, c *ClientOption, fn CreateUserFunc) erro
 
 	return fn(ssoClient)
 }
+
+type RolePermission struct {
+	Roles       object.AuthorizationRoleList `json:",omitempty"`
+	Permissions []vimType.Permission         `json:",omitempty"`
```
dougm: Doesn't look like `Permissions` is set anywhere?

cmd/vcpctl/provision/provision.go
fanzhangio: I am thinking if we only need to create a "k8s-vcp-default" role with limited permissions in vcpctl. Do we need feature to add more roles or update role by vcpctl in the future ?

dougm: Yes one role should be fine for now.  It may need to be updated in the future, if vcp expands the method set it currently uses.

pkg/cli/cli.go
fanzhangio: I have a first look at the `NodeManager` code. Basically all methods are not exposed to other packages, it's seems like difficult to leverage code from NodeManager. I will look into it to see what's the better solution. Please feel free to share any ideas. @dvonthenen @dougm .
Thanks so much!

dougm: We can refactor the NodeManager code and export the functionality needed, probably into a package that can be shared by both.

pkg/cli/user.go
dougm: I was thinking of the "Minimal set of vSphere roles/privileges" documented here: https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/vcp-roles.html
Ideally the role created by vcpctl would contain those, ok for that to be in a separate PR though.

pkg/cli/cli.go
fanzhangio: One concern is (might not be critical for this case) that keeping multi-connection to vc by different clients (`NodeManager` sets up a `vSphereConnection`, and vcpctl uses its own `vim25.Client`) here could cause performance issue or even race condition on vm operations.

dougm: How about for now, change the PR to check if the property `disk.enableUUID == 1` and log a message if it isn't? The message could even include the govc command to set the property :)

We should check if the property is already set to avoid unneeded `Reconfigure` anyhow.  We can add the node check and `Reconfigure` in a later PR.

pkg/cli/util.go
```diff
@@ -36,3 +39,18 @@ func ReadContent(path string) (string, error) {
 	}
 	return string(content), nil
 }
+
+type extraConfig []types.BaseOptionValue
```
dougm: This type implements [flag.Value](https://golang.org/pkg/flag/#Value), which is useful as the `govc vm.change -e` flag.  In vcpctl, the code can just use `types.OptionValue` directly.

pkg/cli/cli.go
fanzhangio: @dvonthenen I am thinking if `NodeManager` can expose a method to return a nodes list (just name is fine for now). By this kind of method, vcpctl can perform a node check for only `Reconfigure` these particular k8s nodes.

dougm: Separate PR sounds good to me.

pkg/cli/util.go
dougm: Yes, but no need for `types.AnyType()`.  And extra config is always string values, so just `"1"` is fine.

pkg/cli/cli.go
dvonthenen: @fanzhangio @dougm yea, that shouldn't be a problem and it's fairly easy to do. if we do expose a call, we should go through the process of making this the start of a formal API that the provider exposes. I can start working on that on Monday.

PR #50, opened 2018-09-11
report node hostname

figo:
both ip address and hostname needed be reported in node status

out-of-tree CCM reporting node hostname in node status.
this is necessary after the upstream change at #67714,
if not reporting, in the node status, it will only has node ip address, it will break tests and applications.

https://github.com/kubernetes/kubernetes/issues/67714

we need to add test later similar as this one: 
https://github.com/kubernetes/kubernetes/pull/68017,
but since we are building tests for nodemanager, this can be part of nodemanager tests.

**Release note**:

```release-note
None
```
/cc @frapposelli @dougm @dvonthenen

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -232,6 +232,9 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy FindVM) error {
 								}, v1.NodeAddress{
 									Type:    v1.NodeInternalIP,
 									Address: ip,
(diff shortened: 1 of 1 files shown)


dvonthenen:
Is oVM.Guest.HostName the FQDN? or just the hostname on its own?

dougm:
The value comes from the guest (via vmware-tools), depends how the guest OS is configured.  Could be a hostname or fqdn.

figo:
@dvonthenen @dougm  agree, this is the same as run os.HostName() inside guest.

PR #56, opened 2018-09-20
Expose gRPC call to obtain k8s VMs

dvonthenen:
Implements a gRPC endpoint to obtain a list of k8s nodes (master+workers) in a vSphere environment. Implemented based on the comment found here: https://github.com/kubernetes/cloud-provider-vsphere/pull/46#discussion_r217589906

These nodes returned are only nodes belonging to a kubernetes cluster. The gRPC call allows obtaining all nodes across all configured vCenters, nodes that belong within a given vCenter, and nodes in a specific Datacenter within a vCenter.

Added go tests for new functionality in node manager and also to test the gRPC server/client functionality.


pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -320,3 +326,109 @@ func (nm *NodeManager) convertK8sUUIDtoNormal(k8sUUID string) string {
...
+		for _, dc := range nm.vcList[vcenter].dcList {
+			nm.datacenterToNodeList(dc.vmList, nodeList)
+		}
+	} else {
+		for _, vc := range nm.vcList {
+			for _, dc := range vc.dcList {
+				nm.datacenterToNodeList(dc.vmList, nodeList)
+			}
+		}
+	}
```
frapposelli: disclaimer: I'm nit picking here...

I would suggest to use `switch` instead of `if-else` chains, they're easier to read and more idiomatic.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -320,3 +326,109 @@ func (nm *NodeManager) convertK8sUUIDtoNormal(k8sUUID string) string {
...
+			Dnsnames:   make([]string, 0),
+			Addresses:  make([]string, 0),
+			Uuid:       node.UUID,
+		}
+		for _, address := range node.NodeAddresses {
+			if address.Type == v1.NodeExternalIP {
+				pbNode.Addresses = append(pbNode.Addresses, address.Address)
+			} else if address.Type == v1.NodeHostName {
+				pbNode.Dnsnames = append(pbNode.Dnsnames, address.Address)
+			}
```
frapposelli: Same as above

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: Good feedback. I agree. Will make changes.

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: Same as above

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: I just took a look and the first if statement is conditional on 2 parameters.

pkg/cloudprovider/vsphere/proto/cloudprovidervsphere.proto
```diff
@@ -0,0 +1,47 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "com.vmware.cloudprovider.vsphere";
```
figo: just for my understanding, why we need to define the a java_package name? thanks, maybe you can add an comment here.

pkg/cloudprovider/vsphere/server/server.go
```diff
@@ -0,0 +1,90 @@
...
+type GRPCServer interface {
+	Start()
+}
+
+type server struct {
+	s       *grpc.Server
+	nodeMgr NodeManagerInterface
+}
+
+// SayHello implements helloworld.GreeterServer
```
figo: comment need be updated, :)

pkg/cloudprovider/vsphere/server/server_test.go
```diff
@@ -0,0 +1,85 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	r, err := c.ListNodes(ctx, &pb.ListNodesRequest{})
+	if err != nil {
+		t.Fatalf("could not greet: %v", err)
+	}
+
+	found := false
+	for _, node := range r.Nodes {
+		if node.Uuid == "422e4956-ad22-1139-6d72-59cc8f26bc90" {
```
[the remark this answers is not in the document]
dvonthenen: The gRPC client doesnt need to run on the same node. That is just the go test. I did all my testing running the kubernetes cluster on Nimbus and the client was running from my laptop.

pkg/cloudprovider/vsphere/proto/cloudprovidervsphere.proto
dvonthenen: Because gRPC supports multiple languages. We just happen to be using Go. Someone out there might be using Java. It's my understanding this is required to generate a Java client.

pkg/cloudprovider/vsphere/server/server.go
dvonthenen: Ooops. Will change that.

dvonthenen:
The gRPC client doesnt need to run on the same node. That is just the go test. I did all my testing running the kubernetes cluster on Nimbus and the client was running from my laptop.

dvonthenen:
You are correct. Only the insecure connection is implemented. This was done in order so that work on vcpctl can move forward. I will need to implement functionality and a mechanism for allowing secure/validated connections. Will have a follow up PR for that. For now, removing the k8s service definition will effectively disable the functionality.

PR #57, opened 2018-09-23
Default to gcr.io image registry

akutz:
This PR changes the behavior of building and pushing images to use the gcr.io image registry.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: NA

This PR includes a number of small fixes discovered while implementing the primary feature. However, the changes **are** isolated into individual commits.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/make.sh
+++ b/hack/make.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Copyright 2018 The Kubernetes Authors.
 #
@@ -13,6 +13,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# posix complaint
+# verified by https://www.shellcheck.net
 
 # Simple script to offer the ability to run the make directives in a Docker Container.
 # In other words you don't need a Go env on your system to run Make (build, test etc)
--- a/hack/make.sh
+++ b/hack/make.sh
@@ -18,6 +18,13 @@
 # In other words you don't need a Go env on your system to run Make (build, test etc)
 # This script will just bind-mount the source directory into a container under the correct
 # GOPATH and handle all of the Go ENV stuff for you.  All you need is Docker
-docker run -i -v "$PWD":/go/src/k8s.io/cloud-provider-vsphere:z \
+
+# When in an interactive terminal add the -t flag so Docker inherits
+# a pseudo TTY. Otherwords SIGINT does not work to kill the container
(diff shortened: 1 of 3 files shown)


hack/make.sh
```diff
@@ -14,10 +14,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# posix complaint
```
frapposelli: s/complaint/compliant

hack/make.sh
```diff
@@ -14,10 +14,21 @@
...
+# When in an interactive terminal add the -t flag so Docker inherits
+# a pseudo TTY. Otherwords SIGINT does not work to kill the container
+# when running this script interactively.
+TERM_FLAGS="-i"
+echo "${-}" | grep -q i && TERM_FLAGS="${TERM_FLAGS}t"
+
+docker run ${TERM_FLAGS} \
+  -v "$(pwd)":/go/src/k8s.io/cloud-provider-vsphere:z \
+  -w /go/src/k8s.io/cloud-provider-vsphere \
+  golang:1.11 make "${@}"
```
frapposelli: can we parametrize the go version somewhere? it's going to be easier to source it from a known location across all the scripts in the repo.

hack/make.sh
akutz: Fixed

hack/make.sh
akutz: Hi @frapposelli,

Is it used elsewhere? I kind of feel the version of Go should be fixed. We don't necessarily want it to be easy for people to produce binaries with different Go versions. I think it's good that it's fairly assured that the image is the same each time.

akutz:
Hi @frapposelli,

Any other changes I should make?

hack/make.sh
```diff
@@ -14,10 +14,21 @@
...
 # Simple script to offer the ability to run the make directives in a Docker Container.
 # In other words you don't need a Go env on your system to run Make (build, test etc)
 # This script will just bind-mount the source directory into a container under the correct
 # GOPATH and handle all of the Go ENV stuff for you.  All you need is Docker
-docker run -i -v "$PWD":/go/src/k8s.io/cloud-provider-vsphere:z \
-	-w /go/src/k8s.io/cloud-provider-vsphere \
-	golang:1.10 make "$@"
+
+# When in an interactive terminal add the -t flag so Docker inherits
+# a pseudo TTY. Otherwords SIGINT does not work to kill the container
```
figo: Otherwords -> Otherwise?

PR #58, opened 2018-09-23
Adds CCM user; fixes "watch secret" errors

akutz:
This patch binds the user `cloud-controller-manager` to the role `system:cloud-controller-manager` in order to enable the use of a custom Kubernetes user and thus no longer needing to provide the CCM with the controller-manager's kubeconfig.

This patch also fixes the annoying "watch secret" failures that populate the logs. Adding the `watch` permission to the `secrets` resource for the `system:cloud-controller-manager` role and using the `cloud-controller-manager` user's `kubeconfig` will result in these errors being a thing of the past.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: NA

**Release note**:
```release-note
NONE
```

### DIFF
--- a/manifests/controller-manager/cloud-controller-manager-role-bindings.yaml
+++ b/manifests/controller-manager/cloud-controller-manager-role-bindings.yaml
@@ -36,5 +36,7 @@ items:
   - kind: ServiceAccount
     name: cloud-controller-manager
     namespace: kube-system
+  - kind: User
+    name: cloud-controller-manager
 kind: List
 metadata: {}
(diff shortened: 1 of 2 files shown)


frapposelli:
I seem to recall a change to the pod manifest, from the debugging session we did, was that required at all?

akutz:
Hi @frapposelli,

Any other changes I should make?

frapposelli:
All good, as per offline conversation

/lgtm
/approve

PR #59, opened 2018-09-23
Fix insecure as true by default, pass flag to client option

fanzhangio:
Quick fix about enable insecure by default. Pass insecure flag to client option

### DIFF
--- a/cmd/vcpctl/provision/provision.go
+++ b/cmd/vcpctl/provision/provision.go
@@ -87,7 +87,7 @@ func RunProvision(cmd *cobra.Command, args []string) {
 	// TODO (fanz): implement provision
 	fmt.Println("Perform cloud provider provisioning...")
 	o := cli.ClientOption{}
-	o.LoadCredential(vcUser, vcPassword, vcCert, vcRole)
+	o.LoadCredential(vcUser, vcPassword, vcCert, vcRole, insecure)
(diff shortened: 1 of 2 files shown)


cmd/vcpctl/provision/provision.go
```diff
@@ -76,7 +76,7 @@ func AddProvision(cmd *cobra.Command) {
 	provisionCmd.Flags().StringVar(&vcport, "port", "", "Specify vCenter Port")
 	provisionCmd.Flags().StringVar(&vcUser, "user", "", "Specify vCenter user")
 	provisionCmd.Flags().StringVar(&vcPassword, "password", "", "Specify vCenter Password")
-	provisionCmd.Flags().BoolVar(&insecure, "insecure", false, "Don't verify the server's certificate chain")
+	provisionCmd.Flags().BoolVar(&insecure, "insecure", true, "Don't verify the server's certificate chain")
```
frapposelli: isn't this setting insecure `true` by default?

fanzhangio:
Done. Thanks @fabianofranz @akutz

cmd/vcpctl/provision/provision.go
fanzhangio: Done. Thanks @fabianofranz @akutz

PR #62, opened 2018-09-25
Disable the vSphere CCM API

dvonthenen:
This disables the vSphere CCM API (default is disabled) until a mechanism to securely connect to the gRPC endpoint exists. This will be implemented in a future PR.

This was originally implemented in order to facilitate developed for a feature in vcpctl.

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -21,9 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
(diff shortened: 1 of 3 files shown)


dvonthenen:
Some additional detail from discussions I had with @figo 

Moving the configFromEnv() from vsphere_test.go to cloud.go is mainly due to the helm chart work I am doing. In helm charts, the prefer mechanism for configuring components is using ENV/configmaps as they are easier to override. Note the config file will override any ENV setting which happens to preserve backwards compatibility.

Issue #48, opened 2018-09-11
Add generating certs (key-pair) in vcpctl for provisioning CCM


fanzhangio:
`vcpctl` requires user to provide certificate by `--cert` flag in command line. There needs to add an option feature to generate key-pairs  

> Uncomment only one, leave it on its own line: 
>

> /kind feature

fanzhangio:
There are four Solution User Certificates:
- machine
- vpxd
- vpxd-extension
- vsphere-webclient

To generate the cert key-pair for vcpctl, the process is that:
1) Generate `machine.csr` and `machine.key` and store locally. Note: VMCA requires that the certificate have a valid date of at least 24 hours prior
2) Provide the preceding csr to CA to generate a solution user cert and name the files as `machine_name.cer`, `vpxd.cer`, `vpxd-extension.cer`, `vsphere-webclient.cer` and store these files locally (maybe user can specify the path by flags/ENV)
3) Provide the full path to each of these certificates and keys (including CA certs):
- machine.cer
- machine.key
- vpxd.cer
- vpxd.key
- vpxd-extension.cer
- vpxd-extension.key
- vsphere-webclient.cer
- vsphere-webclient.key
- Root64.cer

dougm:
We don't want to use the existing solution user certs.  vCenter doesn't allow multiple solution users to use the same cert anyhow.

govc has some code to generate a new key pair: https://github.com/vmware/govmomi/blob/2eb1d11bc266cc2b0f1f5477ed2dec6ffdcaa5b0/govc/extension/setcert.go#L84

But is only meant for use in tests at the moment, such as: https://github.com/vmware/govmomi/blob/2eb1d11bc266cc2b0f1f5477ed2dec6ffdcaa5b0/govc/test/session.bats#L80

The solution user key pair is only used for signing login requests, they are not used for TLS client authentication.  No need for a CSR or the cert to be signed by a known CA.

dvonthenen:
Closing because PR https://github.com/kubernetes/cloud-provider-vsphere/pull/65 was merged.

/close

PR #65, opened 2018-09-25
Add generating solution user key pair for login purpose

fanzhangio:
#48 

**vcpctl** allows`--cert` flag to receive solution user certificate like`/path/to/certificate/k8s-vcp.crt`

1) if the cert exists, directly return error since VC does not allow multiple solution users use the same cert.
2) if the cert does not exist, invoke creating certificate function to generate key and cert file with the same id k8s-vcp in the same path directory as `--cert` specified.
3) the generated `k8s-vcp.crt` file will be decoded and assigned back to u.cert again.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -110,8 +110,19 @@ func CreateSolutionUser(ctx context.Context, o *ClientOption) error {
 	u.role = o.getCredential().role
 	u.cert = o.getCredential().cert
 	return u.Run(ctx, o, CreateUserFunc(func(c *ssoadmin.Client) error {
-		// TODO (fanz): By far, cert data is provided from crt file (by --cert flag)
-		//  Will add an option to generate key-pairs in separate PR.
+
+		_, err := os.Stat(u.cert)
+		if err == nil {
+			fmt.Errorf("cert file already exists (%s). Please delete the cert and key files", u.cert)
+		} else if os.IsNotExist(err) {
+			err = u.createCert()
+			if err != nil {
+				fmt.Errorf("Create solution user certificate (%s) error: %s", u.cert, err)
+			}
+		} else {
+			fmt.Errorf("Invalid cert file or directory (%s), create solution user error : %s", u.cert, err)
+		}
+
 		if cert, err := ReadContent(u.cert); err == nil {
 			block, _ := pem.Decode([]byte(cert))
 			if block != nil {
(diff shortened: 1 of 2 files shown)


fanzhangio:
@dougm

pkg/cli/user.go
```diff
@@ -106,3 +116,67 @@ type Role struct {
...
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return err
+	}
+
+	template := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			Organization: []string{"com.vmware.vsphere-cloud-provider"},
```
fanzhangio: What should the org be, I just make an sample here..

pkg/cli/cli.go
```diff
@@ -110,8 +110,14 @@ func CreateSolutionUser(ctx context.Context, o *ClientOption) error {
 	u.role = o.getCredential().role
 	u.cert = o.getCredential().cert
 	return u.Run(ctx, o, CreateUserFunc(func(c *ssoadmin.Client) error {
-		// TODO (fanz): By far, cert data is provided from crt file (by --cert flag)
-		//  Will add an option to generate key-pairs in separate PR.
+
+		if _, err := os.Stat(u.cert); os.IsNotExist(err) {
```
dougm: We should return an error if `u.cert` already exists.

pkg/cli/user.go
```diff
@@ -42,6 +50,8 @@ type CreateUserFunc func(c *ssoadmin.Client) error
 
 func (u *User) Run(ctx context.Context, c *ClientOption, fn CreateUserFunc) error {
 
+	u.id = "k8s-vcp"
```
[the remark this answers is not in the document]
dougm: :+1:

pkg/cli/user.go
dougm: How about just `"sig-vmware"`?

pkg/cli/user.go
```diff
@@ -106,3 +116,67 @@ type Role struct {
 	RoleName   string
 	Privileges []string
 }
+
+// createCert creates a key pair for login purpose
+func (u *User) createCert() error {
+	id := strings.TrimeSuffix(u.cert, ".crt")
```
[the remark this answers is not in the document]
dougm: typo/compile error: `TrimeSuffix`

pkg/cli/user.go
frapposelli: I would suggest a more generic 'kubernetes'

pkg/cli/cli.go
fanzhangio: I am thinking if we could do this:
1) Pass the cert file in command line, like `--cert /path/to/certificate/files/k8s-vcp.crt`
2) The file `k8s-vcp.crt` will be assigned to `u.cert`, if it exists, will use this certificate, so the generation key-pair will not be invoked. Otherwise, will reuse the name of `k8s-vcp.crt` to generate key and cert in the same directory `/path/to/certificate/files/`.

In other way, if we want to create key-pair every time, so there is no need to pass the cert file in `--cert`, just directly point to the path and id name like `--cert /path/to/certificate/files/k8s-vcp` 

Please let me know your ideas, thanks. @dougm

PR #67, opened 2018-09-26
General housekeeping/cleanup

dvonthenen:
This PR contains the following changes:
- Changed the default port number on the API server to 43001. This was done in order to avoid application protocol collision. Old port was used in the example.
- Provided a mechanism to change the default binding for the API server. This is required for Helm Chart development
- Changed ENV variables for API disabling and binding to standardize the names. Now they are VSPHERE_API_DISABLE and VSPHERE_API_BINDING respectively.
- Fixed logging to standardize on glog. There were some places where fmt.XXXX and log.XXXX were used
- Fixed the glog.V(X) logging level to appropriate levels
- Fixed various variable names to standardize on naming in the Config object.

Tested on 1.11.2

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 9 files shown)


pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -152,8 +160,11 @@ func (vs *VSphere) Initialize(clientBuilder controller.ControllerClientBuilder)
 			DeleteFunc: vs.nodeDeleted,
 		})
 
-		if !vs.cfg.Global.DisableAPI {
+		if !vs.cfg.Global.APIDisable {
+			glog.V(1).Info("Startng the API Server")
```
frapposelli: `/s/Startng/Starting/`

Issue #89, opened 2018-10-23
csi: read same config from file as is used with cloud provider


codenrhoden:
When the config map is mounted to the container running the CSI plugin, the plugin should be capable of reading this file. This will allow the same config map to be used for the cloud provider and for CSI.

Since this is just a k/v flat file in the end, this does not make reading that config file K8s specific.

As much as we can, we want to be able to configure the plugin with env vars, but more advanced configs involving multiple vCenter's, for example, would be complicated through that method.

dvonthenen:
PR submitted
https://github.com/kubernetes/cloud-provider-vsphere/pull/91

PR #91, opened 2018-10-25
Refactor configuration gathering for all 3 projects

dvonthenen:
This PR refactors the vsphere.conf format and parsing in the CCM project. The CSI implementation should leverage the same config file that the CCM project is using. It turns out that the vcpctl project was also using the same config file but was directly importing the entire CCM package just to parse the config. The config parsing code was isolated and this new package was placed in a common area that can be used by all three projects.

This refactor does also include basic capabilities to use ENVARs to configure vCenters. Although this method exists, it is highly discouraged not to be used in a production-like deployment because it can be very error prone. It's only to be used for dev/test purposes if at all.

For CSI to read the config:
```
import (
       vcfg "k8s.io/cloud-provider-vsphere/pkg/config"
)

cfg, err := vcfg.ReadConfig(config)
if err != nil {
	return nil, err
}
```

There is also a helper function to setup up the connection objects used by vclib:
```
import (
       vcfg "k8s.io/cloud-provider-vsphere/pkg/config"
)

//genrate connection map
vsphereInstanceMap := vcfg.GenerateInstanceMap(cfg)
```

If more things can be moved over, then we can address in a follow on PR. This is a decent foundation though.

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/89

Tested CCM on 1.11.3
Tested CCM Helm Chart on 1.11.3
make test passes

**Release note**:
Special note that the [Network] section in the vsphere.conf file was removed since it was no longer being referenced anywhere. Documentation was updated to reflect this change. Existing vsphere.conf will need to remove that section otherwise you will get a parse error in the config file.


divyenpatel:
@dvonthenen for CSI, I guess `vsphere.conf` file should not be present on any node?

I see https://github.com/kubernetes/cloud-provider-vsphere/blob/master/manifests/controller-manager/README.md, we will be creating `configmap` from the conf file.

Since we have vCenter's username and password in our config file, We should use `secret` for keeping such credentials safe.

CC: @SandeepPissay @abrarshivani

dvonthenen:
Hi @divyenpatel,

`vsphere.conf` is only present as a configmap and the cloud controller is only deployed on master nodes. The current implementation and recommendation is that creds are stored in secret and it is not recommended to use `vsphere.conf` cred storage.

Please see the documentation for the cloud provider here:
https://github.com/kubernetes/cloud-provider-vsphere/blob/master/docs/deploying_cloud_provider_vsphere_with_rbac.md

If you have anymore questions, please let me know.

PR #97, opened 2018-11-07
Move CredMgr to pkg/common, add DCOS/generic-CO support

dvonthenen:
Moved credential manager into pkg/common for CSI to reuse. Also, adds support for other container orchestrators like DCOS.

Since this is a refactor plus additional implementation to support DCOS and other generic COs, the k8s workflow, docs, and user experience is unchanged.

Tested using k8s on 1.11.3 using k8s secret listener
Tested by mounting the k8s secret to the filesystem which simulates a DCOS deploy of CSI
`make test` functions correctly

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -35,6 +35,7 @@ const (
 	DefaultAPIBinding        string = ":43001"
 	DefaultK8sServiceAccount string = "cloud-controller-manager"
 	DefaultVCenterPort       string = "443"
+	DefaultSecretDirectory   string = "/etc/cloud/secrets"
 )
 
 // Error Messages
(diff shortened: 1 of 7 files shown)


pkg/common/config/config.go
```diff
@@ -137,6 +138,17 @@ func ConfigFromEnv() (cfg Config, ok bool) {
 	}
 	cfg.Global.APIBinding = APIBinding
 
+	var SecertsDirectory string
```
dougm: Typo "Secerts" -> "Secrets".  Could also:
```go
secretsDirectory := os.Getenv("VSPHERE_SECRETS_DIRECTORY")
if secretsDirectory == "" {
    secretsDirectory = DefaultSecretDirectory
}

dvonthenen: Fixed!

dvonthenen:
Thanks for your review @dougm!! Fixed!!

CC: @frapposelli @codenrhoden

PR #102, opened 2018-11-19
Remove dead/stale code

dvonthenen:
This removes dead/stale code that was carried/copied over from the in-tree provider which is no longer needed. This removes the networking section from the configuration file because it isn't needed by the out-of-tree provider. 

There was an original attempt to remove this dead/stale code but it broke a couple of projects that were using/deploying the CCM as apart of these other projects. Sorry about that! This is a second attempt to remove code that is no longer needed before we cut a first version/tag of the CCM.

The behavior of `gcfg` was changed such that if extra or missing sections/keyvalues are parsed, they will not produce a fatal error. Only invalid config file syntax are errors. This was tested with the legacy `network` section in the config file, did not produce a fatal error and the CCM was able to run successfully.

When these lines of code were reverted/added back into the CCM, the updates to the documentation, example YAML, the outstanding Helm chart, etc were not reverted along with it. It would be very nice to remove this code/property from the config so we don't create technical debt or legacy issues before we cut the first version.

If we decided not to merge this PR, we can close this PR but we would need to create a new PR to "revert" the documentation, example YAML, Helm chart, etc to match the existence of this property in the config file.

CC: @frapposelli @dougm @codenrhoden @akutz

### DIFF
--- a/pkg/cloudprovider/vsphere/vsphere_test.go
+++ b/pkg/cloudprovider/vsphere/vsphere_test.go
@@ -93,7 +93,6 @@ func configFromSimWithTLS(tlsConfig *tls.Config, insecureAllowed bool) (vcfg.Con
 	cfg.Global.User = s.URL.User.Username()
 	cfg.Global.Password, _ = s.URL.User.Password()
 	cfg.Global.Datacenters = vclib.TestDefaultDatacenter
-	cfg.Network.PublicNetwork = vclib.TestDefaultNetwork
 	cfg.VirtualCenter = make(map[string]*vcfg.VirtualCenterConfig)
 	cfg.VirtualCenter[s.URL.Hostname()] = &vcfg.VirtualCenterConfig{
 		User:         cfg.Global.User,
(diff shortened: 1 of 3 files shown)


frapposelli:
@dvonthenen I agree with the sentiment here, but right now the default behavior for unmanaged properties in the config file is to panic, and this is not ideal for many that are transferring their old configurations over.

We have discussed this during one of the cloud-provider-vsphere meetings, the consensus was to explore if the `gcfg` has a parsing method that can handle unexpected fields gracefully (as opposed to panic), if that's the case, I would support merging this change and drop `[network]` altogether.

akutz:
Hi @frapposelli,

I agree. This close to KubeCon there is simply no good reason to introduce a possible panic.

frapposelli:
/lgtm

Waiting on @akutz lgtm before approving

akutz:
Hi @frapposelli / @dvonthenen,

I'm an idiot. The above comment's build proves nothing as the build is using the latest CCM image which doesn't include @dvonthenen's change. This PR /lgtm as long as the CCM can stand up without a panic when the `network` section is still present.

akutz:
Tested thoroughly. /lgtm. I do wonder if we should add the deprecation warning back into the code?

PR #108, opened 2018-12-03
Initial Implementation for CSI ListVolumes using FCD

dvonthenen:
This PR implements ListVolumes for First Class Disks (FCD) on DatastoreClusters and Datastores.
- Datastore has a new function to ListDisks which returns all FCD on that datastore
- New class called DatastoreCluster which gets Properties, Child Datastores comprising the DatastoreCluster, and also ListDisks for FCD on the Datastores comprising the DatastoreCluster
- Datacenter has a function ListDisks which returns only unique FCDs on all DatastoreClusters and Datastores then sorts them if pagination is used.

NodeManager:
- Moves DiscoverNode core functionality into ConnectionManager

ConnectionManager:
- Provides a new function that finds the VC/DC combo that owns a given VM called WhichVCandDCByNodeId
- Provides a "mock" function that finds a VC/DC comobo based on a zone called WhichVCandDCByZone. *NOTE: this is a placeholder until we support multiple VC/DC*

Misc:
- Moved various constants and etc to a proper location in code.

**Which issue this PR fixes**: NA (does not have an issue for ListVolumes)

- tested CCM because of refactor of nodeManager.DiscoverNode now uses connectionManager. WhichVCandDCByNodeId
- tested CSI ListVolumes using FCD on a mix of DatastoreClusters and traditional Datastores. Tested: 1) returning all FCD, 2) tested the first pagination of FCD getting 10 out of 11, 3) then tested the remainder pagination getting 11 of 11.
- tested on k8s 1.11.3
- tested on vSphere 6.7U1

**Release note**:


pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -39,6 +39,9 @@ const (
 	FindVMByUUID FindVM = iota // 0
 	FindVMByName               // 1
 
+	POOL_SIZE  int = 8
+	QUEUE_SIZE int = POOL_SIZE * 10
```
codenrhoden: These two consts are also defined as `cm.POOL_SIZE` and `cm.QUEUE_SIZE`. I think you can use them directly rather than redefining them here since you are importing `pkg/common/connectionmanager`.

pkg/csi/service/fcd/constants.go
```diff
@@ -0,0 +1,32 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fcd
+
+const (
+	// MbInBytes... just like it sounds but as aint64
```
codenrhoden: s/a int64/aint64

dougm:
I was thinking the same regarding tests.  I can put together the preamable that populates some datastores, ds cluster and some disks.  I'll just drop into a gist and link here.

dougm:
Test preamble creates 2 DCs, 2 standalone Datastores per DC, 1 Datastore cluster per DC with 2 datastores and some disks: https://gist.github.com/dougm/1ce1797faef9a63d961c303f49781419

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: Not needed. Has been removed.

pkg/csi/service/fcd/constants.go
dvonthenen: Fixed.

dvonthenen:
> Test preamble creates 2 DCs, 2 standalone Datastores per DC, 1 Datastore cluster per DC with 2 datastores and some disks: https://gist.github.com/dougm/1ce1797faef9a63d961c303f49781419

@dougm Currently we only support 1 VC with 1 DC. Any test preamble should reflect that.

dougm:
@dvonthenen supporting 1 DC in the CSI configuration, yes.  But the code should tolerate > 1 DC in the VC environment, having 2 in the test can help ensure that.

dvonthenen:
Added the `LIST_VOLUMES` capability

pkg/common/vclib/utils.go
```diff
@@ -206,5 +206,16 @@ func VerifyVolumePathsForVMDevices(vmDevices object.VirtualDeviceList, volPaths
 			}
 		}
 	}
+}
 
+func ExistsInList(needle string, haystack []string, caseSensitive bool) bool {
+	for _, straw := range haystack {
+		if caseSensitive && straw == needle {
+			return true
+		} else if strings.ToLower(straw) == strings.ToLower(needle) {
```
dougm: `strings.EqualFold(straw, needle)`

akutz: See an [explanation](https://github.com/akutz/sortfold) as to why `EqualFold` should almost always be used for case-insensitive string comparisons

dvonthenen: Will fix

dvonthenen:
Thanks @akutz for the help with the CI stuff!

Issue #78, opened 2018-10-09
csi: support for datastore cluster(s)


codenrhoden:
Current in-tree solution does not support datastore clusters, only accessing a specific datastore.

dougm:
Starting with support for placement (https://github.com/vmware/govmomi/issues/740)

dougm:
Other than placement, we'll need to have a `DatastoreCluster` config option or add support to treat existing `Datastore` config option as a cluster.  cc @dvonthenen

dvonthenen:
> Other than placement, we'll need to have a `DatastoreCluster` config option or add support to treat existing `Datastore` config option as a cluster. cc @dvonthenen

Yep, I have another PR to follow the current one open. I will tackle this when I remove the network section.

jingxu97:
@dvonthenen wondering whether you have any new update for this support? Thank you!

Issue #104, opened 2018-11-28
csi: Implement volume create/delete


codenrhoden:
As part of implementing CSI spec functionality, first step is adding volume create and delete.

Right now this is simplistic, as it just goes to the identified datastore, rather than a datastore cluster.

dvonthenen:
I think we can close based on this PR https://github.com/kubernetes/cloud-provider-vsphere/pull/110

PR #110, opened 2018-12-04
Initial implementation for CSI create/delete

dvonthenen:
Initial implementation for Create/Delete for vSphere CSI.
- Handles DatastoreClusters, VSAN datastores, and "traditional" Datastores
- added a comment/doc idea for how to implement zones

Other:
- connectiomanager implements WhichVCandDCByFCDId to retrieve which VC/DC owns a given FCD
- reverted recent changes to nodemanager.DiscoverNode() which was causing the go tests to fail
- refactored some ctx stuff on nodemanager and connectionmanager to handle some intermittent go test failures
- changed a few functions on Datacenter to return DatastoreInfo instead of Datastore
- updated go tests to reflect the above
- removed some unnecessary loops carried over from the in-tree version

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/78 and https://github.com/kubernetes/cloud-provider-vsphere/issues/104

- tested CSI Create/Delete FCD on DatastoreClusters, VSAN datastores, and "traditional" Datastores
- tested on k8s 1.11.3
- tested on vSphere 6.7U1

**Release note**:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1088,7 +1088,7 @@
   revision = "e3762e86a74c878ffed47484592986685639c2cd"
 
 [[projects]]
-  digest = "1:f7f7777d09307c2c606e4779fd7810e53bb18d07ea2f8860a60abe67c4973b6b"
+  digest = "1:cba624958aaf538386f59ff24426f41d2521ea215c55274dba0a6271d4556876"
   name = "k8s.io/kubernetes"
   packages = [
     "cmd/cloud-controller-manager/app",
@@ -1140,15 +1140,24 @@
     "pkg/util/file",
     "pkg/util/flag",
     "pkg/util/hash",
+    "pkg/util/io",
     "pkg/util/metrics",
+    "pkg/util/mount",
     "pkg/util/net/sets",
     "pkg/util/node",
+    "pkg/util/nsenter",
     "pkg/util/parsers",
     "pkg/util/pointer",
     "pkg/util/taints",
     "pkg/version",
     "pkg/version/prometheus",
     "pkg/version/verflag",
+    "pkg/volume",
+    "pkg/volume/util",
+    "pkg/volume/util/fs",
+    "pkg/volume/util/recyclerclient",
+    "pkg/volume/util/types",
+    "pkg/volume/util/volumepathhandler",
   ]
   pruneopts = "UT"
   revision = "bb9ffb1654d4a729bb4cec18ff088eacc153c239"
@@ -1161,6 +1170,14 @@
   packages = ["pkg/signals"]
   pruneopts = "UT"
   revision = "be98dc6210ab86895825bb9ed212f9f60557c33e"
+
+[[projects]]
+  branch = "master"
+  digest = "1:381323c2fe2e890a3dd3b5d6dc6f2199068408cca89b24f6b7ca1c60f32644a5"
+  name = "k8s.io/utils"
+  packages = ["exec"]
+  pruneopts = "UT"
+  revision = "0d26856f57b32ec3398579285e5c8a2bfe8c5243"
 
 [solve-meta]
   analyzer-name = "dep"
@@ -1198,7 +1215,9 @@
     "github.com/vmware/govmomi/vslm",
     "golang.org/x/net/context",
     "google.golang.org/grpc",
+    "google.golang.org/grpc/codes",
     "google.golang.org/grpc/reflection",
+    "google.golang.org/grpc/status",
     "gopkg.in/gcfg.v1",
     "k8s.io/api/core/v1",
     "k8s.io/apimachinery/pkg/api/errors",
@@ -1225,6 +1244,7 @@
     "k8s.io/kubernetes/pkg/util/flag",
     "k8s.io/kubernetes/pkg/version/prometheus",
     "k8s.io/kubernetes/pkg/version/verflag",
+    "k8s.io/kubernetes/pkg/volume/util",
     "k8s.io/sample-controller/pkg/signals",
   ]
   solver-name = "gps-cdcl"
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/golang/glog"
 	"k8s.io/api/core/v1"
@@ -31,6 +33,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
 )
 
 type FindVM int
@@ -93,48 +96,199 @@ func (nm *NodeManager) removeNode(node *v1.Node) {
 }
 
 func (nm *NodeManager) DiscoverNode(nodeID string, searchBy FindVM) error {
-	vm, err := nm.connectionManager.WhichVCandDCByNodeId(nodeID, cm.FindVM(searchBy))
-	if err != nil {
-		return err
+	if nodeID == "" {
+		glog.V(3).Info("DiscoverNode called but nodeID is empty")
+		return vclib.ErrNoVMFound
+	}
+	type vmSearch struct {
+		vc         string
+		datacenter *vclib.Datacenter
 	}
 
-	var oVM mo.VirtualMachine
-	err = vm.VM.Properties(context.Background(), vm.VM.Reference(), []string{"config", "summary", "summary.config", "guest.net", "guest"}, &oVM)
-	if err != nil {
-		glog.Errorf("Error collecting properties for vm=%+v in vc=%s and datacenter=%s: %v",
-			vm.NodeName, vm.VcServer, vm.DataCenter.Name(), err)
-		return err
+	var mutex = &sync.Mutex{}
+	var globalErrMutex = &sync.Mutex{}
+	var queueChannel chan *vmSearch
+	var wg sync.WaitGroup
+	var globalErr *error
+
+	queueChannel = make(chan *vmSearch, cm.QUEUE_SIZE)
+
+	myNodeID := nodeID
+	if searchBy == FindVMByUUID {
+		glog.V(3).Info("DiscoverNode by UUID")
+		myNodeID = strings.ToLower(nodeID)
+	} else {
+		glog.V(3).Info("DiscoverNode by Name")
+	}
+	glog.V(2).Info("DiscoverNode nodeID: ", myNodeID)
+
+	vmFound := false
+	globalErr = nil
+
+	setGlobalErr := func(err error) {
+		globalErrMutex.Lock()
+		globalErr = &err
+		globalErrMutex.Unlock()
 	}
 
-	addrs := []v1.NodeAddress{}
-	for _, v := range oVM.Guest.Net {
-		for _, ip := range v.IpAddress {
-			if net.ParseIP(ip).To4() != nil {
-				v1helper.AddToNodeAddresses(&addrs,
-					v1.NodeAddress{
-						Type:    v1.NodeExternalIP,
-						Address: ip,
-					}, v1.NodeAddress{
-						Type:    v1.NodeInternalIP,
-						Address: ip,
-					}, v1.NodeAddress{
-						Type:    v1.NodeHostName,
-						Address: oVM.Guest.HostName,
-					},
-				)
+	setVMFound := func(found bool) {
+		mutex.Lock()
+		vmFound = found
+		mutex.Unlock()
+	}
+
+	getVMFound := func() bool {
+		mutex.Lock()
+		found := vmFound
+		mutex.Unlock()
+		return found
+	}
+
+	ctx := context.Background()
+
+	go func() {
+		var datacenterObjs []*vclib.Datacenter
+		for vc, vsi := range nm.connectionManager.VsphereInstanceMap {
+
+			found := getVMFound()
+			if found == true {
+				break
+			}
+
+			var err error
+			for i := 0; i < 3; i++ {
+				err = nm.connectionManager.Connect(ctx, vc)
+				if err == nil {
+					break
+				}
+				err = nm.connectionManager.ConnectByInstance(ctx, vsi)
+				if err == nil {
+					break
+				}
+				time.Sleep(1 * time.Second)
+			}
+
+			if err != nil {
+				glog.Error("Discovering node error vc:", err)
+				setGlobalErr(err)
+				continue
+			}
+
+			if vsi.Cfg.Datacenters == "" {
+				datacenterObjs, err = vclib.GetAllDatacenter(ctx, vsi.Conn)
+				if err != nil {
+					glog.Error("Discovering node error dc:", err)
+					setGlobalErr(err)
+					continue
+				}
+			} else {
+				datacenters := strings.Split(vsi.Cfg.Datacenters, ",")
+				for _, dc := range datacenters {
+					dc = strings.TrimSpace(dc)
+					if dc == "" {
+						continue
+					}
+					datacenterObj, err := vclib.GetDatacenter(ctx, vsi.Conn, dc)
+					if err != nil {
+						glog.Error("Discovering node error dc:", err)
+						setGlobalErr(err)
+						continue
+					}
+					datacenterObjs = append(datacenterObjs, datacenterObj)
+				}
+			}
+
+			for _, datacenterObj := range datacenterObjs {
+				found := getVMFound()
+				if found == true {
+					break
+				}
+
+				glog.V(4).Infof("Finding node %s in vc=%s and datacenter=%s", myNodeID, vc, datacenterObj.Name())
+				queueChannel <- &vmSearch{
+					vc:         vc,
+					datacenter: datacenterObj,
+				}
 			}
 		}
+		close(queueChannel)
+	}()
+
+	for i := 0; i < cm.POOL_SIZE; i++ {
+		go func() {
+			for res := range queueChannel {
+				var vm *vclib.VirtualMachine
+				var err error
+				if searchBy == FindVMByUUID {
+					vm, err = res.datacenter.GetVMByUUID(ctx, myNodeID)
+				} else {
+					vm, err = res.datacenter.GetVMByDNSName(ctx, myNodeID)
+				}
+
+				if err != nil {
+					glog.Errorf("Error while looking for vm=%+v in vc=%s and datacenter=%s: %v",
+						vm, res.vc, res.datacenter.Name(), err)
+					if err != vclib.ErrNoVMFound {
+						setGlobalErr(err)
+					} else {
+						glog.V(2).Infof("Did not find node %s in vc=%s and datacenter=%s",
+							myNodeID, res.vc, res.datacenter.Name())
+					}
+					continue
+				}
+
+				var oVM mo.VirtualMachine
+				err = vm.Properties(ctx, vm.Reference(), []string{"config", "summary", "guest"}, &oVM)
+				if err != nil {
+					glog.Errorf("Error collecting properties for vm=%+v in vc=%s and datacenter=%s: %v",
+						vm, res.vc, res.datacenter.Name(), err)
+					continue
+				}
+
+				addrs := []v1.NodeAddress{}
+				for _, v := range oVM.Guest.Net {
+					for _, ip := range v.IpAddress {
+						if net.ParseIP(ip).To4() != nil {
+							v1helper.AddToNodeAddresses(&addrs,
+								v1.NodeAddress{
+									Type:    v1.NodeExternalIP,
+									Address: ip,
+								}, v1.NodeAddress{
+									Type:    v1.NodeInternalIP,
+									Address: ip,
+								}, v1.NodeAddress{
+									Type:    v1.NodeHostName,
+									Address: oVM.Guest.HostName,
+								},
+							)
+						}
+					}
+				}
+
+				glog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
+					nodeID, vm, res.vc, res.datacenter.Name())
+				glog.V(2).Info("Hostname: ", oVM.Guest.HostName, " UUID: ", oVM.Summary.Config.Uuid)
+
+				nodeInfo := &NodeInfo{dataCenter: res.datacenter, vm: vm, vcServer: res.vc,
+					UUID: oVM.Summary.Config.Uuid, NodeName: oVM.Guest.HostName, NodeAddresses: addrs}
+				nm.addNodeInfo(nodeInfo)
+				setVMFound(true)
+				break
+			}
+			wg.Done()
+		}()
+		wg.Add(1)
+	}
+	wg.Wait()
+	if vmFound {
+		return nil
+	}
+	if globalErr != nil {
+		return *globalErr
 	}
 
-	glog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
-		nodeID, vm.NodeName, vm.VcServer, vm.DataCenter.Name())
-	glog.V(2).Info("Hostname: ", oVM.Guest.HostName, " UUID: ", oVM.Summary.Config.Uuid)
-
-	nodeInfo := &NodeInfo{dataCenter: vm.DataCenter, vm: vm.VM, vcServer: vm.VcServer,
-		UUID: oVM.Summary.Config.Uuid, NodeName: oVM.Guest.HostName, NodeAddresses: addrs}
-	nm.addNodeInfo(nodeInfo)
-
-	return nil
+	glog.V(4).Infof("Discovery Node: %q vm not found", myNodeID)
+	return vclib.ErrNoVMFound
 }
 
 // Reformats UUID to match vSphere format
--- a/pkg/cloudprovider/vsphere/nodemanager_test.go
+++ b/pkg/cloudprovider/vsphere/nodemanager_test.go
@@ -115,6 +115,7 @@ func TestDiscoverNodeByName(t *testing.T) {
 		t.Errorf("Failed to construct/authenticate vSphere: %s", err)
 	}
 	vsphere.connectionManager = cm.NewConnectionManager(&cfg, nil)
+	defer vsphere.connectionManager.Logout()
 
 	nm := NodeManager{
 		nodeNameMap:       make(map[string]*NodeInfo),
@@ -126,7 +127,6 @@ func TestDiscoverNodeByName(t *testing.T) {
 
 	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
 	name := vm.Config.GuestFullName
-
 	err = nm.connectionManager.VsphereInstanceMap[cfg.Global.VCenterIP].Conn.Connect(context.Background())
 	if err != nil {
 		t.Errorf("Failed to Connect to vSphere: %s", err)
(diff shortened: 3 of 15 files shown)


pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -34,10 +36,20 @@ import (
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
 )
 
+func (s *SearchIndex) FindByUUID(req *types.FindByUuid) soap.HasFault {
```
dougm: vcsim already implements `FindByUUID`, you shouldn't need to override as we did for `FindByIp` a while back.

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -34,10 +36,20 @@ import (
...
+		res.Res.Returnval = &s.vm.Self
+	}
+	return res
+}
+
 func TestRegUnregNode(t *testing.T) {
 	cfg, ok := configFromEnvOrSim()
 	defer ok()
 
+	time.Sleep(5 * time.Second)
```
dougm: Is this sleep needed?

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -110,11 +134,14 @@ func TestDiscoverNodeByName(t *testing.T) {
...
```
dougm: needed?

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -126,6 +153,7 @@ func TestDiscoverNodeByName(t *testing.T) {
 
 	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
 	name := vm.Config.GuestFullName
+	fmt.Println("Name:", name)
```
dougm: `t.Logf` instead, gives filename:linenumber and only shows up with `go test -v`

pkg/common/connectionmanager/connectionmanager.go
```diff
@@ -268,6 +274,9 @@ func (cm *ConnectionManager) WhichVCandDCByNodeId(nodeID string, searchBy FindVM
 		return found
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
```
dougm: It's from the original code, but don't think there's a need for using `WithCancel` here, just `ctx` as-is should be fine.

pkg/common/vclib/datacenter.go
```diff
@@ -383,6 +396,135 @@ func (dc *Datacenter) GetAllDatastoreClusters(ctx context.Context, child bool) (
...
+
+	task, err := m.CreateDisk(ctx, spec)
+	if err != nil {
+		glog.Errorf("CreateDisk failed. Err: %v", err)
+		return err
+	}
+
+	err = task.Wait(ctx)
+	if err != nil {
+		glog.Errorf("Wait failed. Err: %v", err)
```
dougm: More context here would be good, `"CreateDisk(%s) failed. Err: %v", diskName, err)`

pkg/common/vclib/datacenter_test.go
```diff
@@ -133,7 +134,8 @@ func TestDatacenter(t *testing.T) {
 		t.Error(err)
 	}
 
-	diskPath := ds.Path(avm.Name + "/disk1.vmdk")
+	fmt.Println("ds.Datastore:", ds.Datastore.Name())
```
dougm: `t.Logf`

pkg/common/vclib/datastore.go
```diff
@@ -133,6 +133,41 @@ func (ds *Datastore) ListFirstClassDisks(ctx context.Context) ([]*FirstClassDisk
...
+func (ds *Datastore) GetFirstClassDiskByName(ctx context.Context, diskName string) (*FirstClassDisk, error) {
+	m := vslm.NewObjectManager(ds.Client())
+
+	oids, err := m.List(ctx, ds.Reference())
+	if err != nil {
+		glog.Errorf("Failed to list disks. Err: %v", err)
+		return nil, err
+	}
+
+	var ids []string
```
dougm: Could drop this var and loop, just use `oids` as-is in the main loop.

pkg/common/vclib/datastore.go
```diff
@@ -170,3 +205,42 @@ func (dsi *DatastoreInfo) ListFirstClassDiskInfos(ctx context.Context) ([]*First
...
+func (dsi *DatastoreInfo) GetFirstClassDiskInfoByName(ctx context.Context, diskName string) (*FirstClassDiskInfo, error) {
+	m := vslm.NewObjectManager(dsi.Datacenter.Client())
+
+	oids, err := m.List(ctx, dsi.Reference())
+	if err != nil {
+		glog.Errorf("Failed to list disks. Err: %v", err)
+		return nil, err
+	}
+
+	var ids []string
```
dougm: Could drop this var and loop, just use `oids` as-is in the main loop.

pkg/common/vclib/storagepod.go
```diff
@@ -134,13 +133,96 @@ func (spi *StoragePodInfo) ListFirstClassDisksInfo(ctx context.Context) ([]*Firs
...
+	m := vslm.NewObjectManager(spi.Datacenter.Client())
+
+	for _, child := range spi.DatastoreInfos {
+		oids, err := m.List(ctx, child)
+		if err != nil {
+			glog.Errorf("Failed to list disks. Err: %v", err)
+			return nil, err
+		}
+
+		var ids []string
```
dougm: Could drop this var and loop, just use `oids` as-is in the main loop.

pkg/common/vclib/storagepod.go
```diff
@@ -202,3 +284,46 @@ func (sp *StoragePod) ListFirstClassDisks(ctx context.Context) ([]*FirstClassDis
...
+	m := vslm.NewObjectManager(sp.Datacenter.Client())
+
+	for _, child := range sp.Datastores {
+		oids, err := m.List(ctx, child)
+		if err != nil {
+			glog.Errorf("Failed to list disks. Err: %v", err)
+			return nil, err
+		}
+
+		var ids []string
```
dougm: Could drop this var and loop, just use `oids` as-is in the main loop.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,107 @@ func (c *controller) CreateVolume(
...
+	zone := "TODO"
+	discoveryInfo, err := c.connMgr.WhichVCandDCByZone(ctx, zone)
+	if err != nil {
+		log.Errorf("Failed to retrieve VC/DC based on zone %s. Err: %v", zone, err)
+		return nil, err
+	}
+
+	err = discoveryInfo.DataCenter.DeleteFirstClassDisk(ctx, datastoreName, datastoreType, req.VolumeId)
+	if err != nil {
+		log.Errorf("GetAllFirstClassDisks failed. Err: %v", err)
```
dougm: s/GetAllFirstClassDisks/DeleteFirstClassDisk/g

Gopkg.lock
```diff
@@ -360,6 +360,14 @@
   revision = "4b7aa43c6742a2c18fdef89dd197aaae7dac7ccd"
   version = "1.0.1"
 
+[[projects]]
+  digest = "1:8f41bbca0d6435bb54d7c090336f7693942cda30f9126eaf3959b8df48721ac3"
+  name = "github.com/myesui/uuid"
```
dougm: There are already 2 vendored uuid packages, `github.com/google/uuid` and `github.com/pborman/uuid`, can we use one of those?

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: Sorry, I was debugging last night trying to figure out that intermittent go test failure and had to step away as it getting late.

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: See first comment above.

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: See first comment above.

pkg/cloudprovider/vsphere/nodemanager_test.go
dvonthenen: See first comment above.

pkg/common/connectionmanager/connectionmanager.go
dvonthenen: Will remove

pkg/common/vclib/datastore.go
dvonthenen: Can do... this is just a copy and paste from an existing function. Will get rid of the loop.

pkg/common/vclib/datacenter_test.go
dvonthenen: See first comment above.

pkg/common/vclib/datastore.go
dvonthenen: Can do... this is just a copy and paste from an existing function. Will get rid of the loop.

pkg/common/vclib/storagepod.go
dvonthenen: See previous comment

pkg/common/vclib/storagepod.go
dvonthenen: See previous comment

pkg/common/vclib/datacenter.go
dvonthenen: Agreed. Will update.

pkg/csi/service/fcd/controller.go
dvonthenen: Will fix

Gopkg.lock
dvonthenen: Will use google

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
+			return nil, err
+		}
+		volName = myUUID.String()
+	}
+
+	// Volume Size - Default is 10 GiB
+	volSizeBytes := int64(DefaultGbDiskSize * GbInBytes)
+	if req.GetCapacityRange() != nil {
+		volSizeBytes = int64(req.GetCapacityRange().GetRequiredBytes())
+	}
```
codenrhoden: So this gets slightly complicated. I'd have to dig into the generated protobuf code, but I think `req.GetCapacityRange()` can return non-nil, but still have zero values for required and limit fields.

So within this `if` statement, we need to see if `GetRequiredBytes` returns 0. If it does, use the default size. And *then* we also need to see if the the capacity limit has been set, and ensure we are not over the limit.  I don't think the spec validator would catch this...

For example, a request could set the required bytes to 0GB, but set a limit of 8GiB, and that would be an invalid request for us with that default value. what do we do then? Use 8GiB? Or return an error? I'd wager the former.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
+	volSizeMB := int64(volumeutil.RoundUpSize(volSizeBytes, GbInBytes)) * 1024
+
+	// Volume Type
+	datastoreType := vclib.TypeDatastoreCluster
+	volType := params[AttributeFirstClassDiskParentType]
```
codenrhoden: Any additional checking needed here for if this parameter is missing? looks like it is required.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
+	//TODO: zones are currently unimplemented. supports single VC/DC only.
+	// Please see function for more details
+	zone := "TODO"
+	discoveryInfo, err := c.connMgr.WhichVCandDCByZone(ctx, zone)
+	if err != nil {
+		log.Errorf("Failed to retrieve VC/DC based on zone %s. Err: %v", zone, err)
+		return nil, err
+	}
+
+	err = discoveryInfo.DataCenter.CreateFirstClassDisk(ctx, datastoreName, datastoreType, volName, volSizeMB)
```
codenrhoden: Does `CreateFirstClassDisk` do the idempotency check for us? What happens if the volume is already created successfully beforehand? If the disk is already created, we just want to return it's details. I'll double check this before finishing this review...

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
+		attributes[AttributeFirstClassDiskOwningDatastore] = firstClassDisk.DatastoreInfo.Info.Name
+	} else {
+		attributes[AttributeFirstClassDiskParentName] = firstClassDisk.DatastoreInfo.Info.Name
+	}
+
+	resp := &csi.CreateVolumeResponse{
+		Volume: &csi.Volume{
+			Id:            firstClassDisk.Config.Id.Id,
+			CapacityBytes: int64(units.FileSize(firstClassDisk.Config.CapacityInMB * MbInBytes)),
+			Attributes:    attributes,
```
codenrhoden: It took me a while to figure out why this was bothering me. It's because the `Attributes` field does not exist in the 1.0 spec...  Which uncovers this new can of worms that this code is still using spec 0.3.  :( I doubt we want to take that on along with this PR, but the sooner the better. We need to update GoCSI for 1.0 and pull that in.  I can't believe I missed that.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
 	ctx context.Context,
 	req *csi.DeleteVolumeRequest) (
 	*csi.DeleteVolumeResponse, error) {
 
-	return nil, nil
+	params := req.GetControllerDeleteSecrets()
+
+	// Volume Type
+	datastoreType := vclib.TypeDatastoreCluster
+	volType := params[AttributeFirstClassDiskParentType]
```
codenrhoden: same as before, I wonder what happens when this parameter is not present.

Furthermore, in 1.0 spec, there are only two fields. volume ID, and secrets. Secrets are meant to be keys, not the volume context/parameters from before. So that brings up an interesting question of how we are going to get that information, because it does not look available to me.

pkg/common/connectionmanager/connectionmanager.go
```diff
@@ -188,7 +194,7 @@ func (cm *ConnectionManager) WhichVCandDCByZone(zone string) (*DiscoveryInfo, er
 		return nil, ErrUnsupportedConfiguration
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
dvonthenen: the ctx is used internally and not a ctx being provided by a user, but I can certainly change it to be explicit

pkg/csi/service/fcd/controller.go
dvonthenen: all of these fields must be provided, with update to check if each one is set and error out if it isnt

pkg/csi/service/fcd/controller.go
dvonthenen: will perform a check to make sure a disk of a similar name has not been create... if it is, I can do one of the following:
1) error out saying disk already exists
2) just return success with the disk info

Let me know which one you prefer

pkg/csi/service/fcd/controller.go
dvonthenen: Will cross that road when we update to 1.0

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
...
+
+	datastoreName := params[AttributeFirstClassDiskParentName]
+
+	//TODO: zones are currently unimplemented. supports single VC/DC only.
+	// Please see function for more details
+	zone := "TODO"
+	discoveryInfo, err := c.connMgr.WhichVCandDCByZone(ctx, zone)
+	if err != nil {
+		log.Errorf("Failed to retrieve VC/DC based on zone %s. Err: %v", zone, err)
+		return nil, err
```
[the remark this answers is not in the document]
dvonthenen: Ok. Will update.

pkg/common/connectionmanager/connectionmanager.go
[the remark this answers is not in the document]
dvonthenen: actually, do not need the cancel... so this issue goes away.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +74,112 @@ func (c *controller) CreateVolume(
 	req *csi.CreateVolumeRequest) (
 	*csi.CreateVolumeResponse, error) {
 
-	return nil, nil
+	params := req.GetParameters()
+
+	// Volume Name
+	volName := req.GetName()
+	if len(volName) == 0 {
+		myUUID, err := uuid.NewUUID()
```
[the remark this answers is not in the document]
dvonthenen: Will make the changes based on our offline discussion and check for required fields and throw an error regardless if the validator is on/off. if on, will obviously pass.

pkg/csi/service/fcd/controller.go
dvonthenen: Will make the change.

pkg/csi/service/fcd/controller.go
dvonthenen: Will make the changes based on our discussion. will return success if the disks attribs are the same.

pkg/csi/service/fcd/controller.go
dvonthenen: Will be making changes based on the new 1.0 spec

pkg/csi/service/fcd/controller.go
dvonthenen: Will be making changes based on the new 1.0 spec

dvonthenen:
Rebased with CSI 1.0 and included everyone's feedback. This is ready for review again. Let me know if there is anything else to include.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
...
+
+	datastoreName := secrets[AttributeFirstClassDiskParentName]
+
+	//check for required parameters
+	if secrets == nil {
+		return nil, status.Errorf(codes.Internal, "Create parameters is a required parameter.")
+	} else if len(volName) == 0 {
+		return nil, status.Errorf(codes.Internal, "Volume name is a required parameter.")
+	} else if len(secrets[AttributeFirstClassDiskParentType]) == 0 {
+		return nil, status.Errorf(codes.Internal, "Volume parmater %s is a required parameter.", AttributeFirstClassDiskParentType)
```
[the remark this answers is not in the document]
dvonthenen: Will fix.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -93,48 +96,202 @@ func (nm *NodeManager) removeNode(node *v1.Node) {
...
-	nm.addNodeInfo(nodeInfo)
+	getVMFound := func() bool {
+		mutex.Lock()
+		found := vmFound
+		mutex.Unlock()
+		return found
+	}
 
-	return nil
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
dvonthenen: Yep it is. will remove the cancel.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -93,48 +96,202 @@ func (nm *NodeManager) removeNode(node *v1.Node) {
...
+						setGlobalErr(err)
+					} else {
+						glog.V(2).Infof("Did not find node %s in vc=%s and datacenter=%s",
+							myNodeID, res.vc, res.datacenter.Name())
+					}
+					continue
+				}
+
+				var oVM mo.VirtualMachine
+				err = vm.Properties(ctx, vm.Reference(), []string{"config", "summary", "summary.config", "guest.net", "guest"}, &oVM)
```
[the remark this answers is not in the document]
dvonthenen: will reduce the list... i didnt know that it would get child fields.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -93,48 +96,202 @@ func (nm *NodeManager) removeNode(node *v1.Node) {
...
+				}
+
+				glog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
+					nodeID, vm, res.vc, res.datacenter.Name())
+				glog.V(2).Info("Hostname: ", oVM.Guest.HostName, " UUID: ", oVM.Summary.Config.Uuid)
+
+				nodeInfo := &NodeInfo{dataCenter: res.datacenter, vm: vm, vcServer: res.vc,
+					UUID: oVM.Summary.Config.Uuid, NodeName: oVM.Guest.HostName, NodeAddresses: addrs}
+				nm.addNodeInfo(nodeInfo)
+				for range queueChannel {
```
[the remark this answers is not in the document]
dvonthenen: Yea, I noticed it a long long time ago and it's one of those things that scares me to remove... like in case it is a timing thing. hahah! will remove and cross my fingers! LOL

dvonthenen:
Updates based on doug's comments

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
 	req *csi.CreateVolumeRequest) (
 	*csi.CreateVolumeResponse, error) {
 
-	return nil, nil
+	secrets := req.GetSecrets()
```
codenrhoden: I don't think we are actually going to be using the secrets field at this time.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
...
+	}
+
+	datastoreName := secrets[AttributeFirstClassDiskParentName]
+
+	//check for required parameters
+	if secrets == nil {
+		return nil, status.Errorf(codes.Internal, "Create parameters is a required parameter.")
+	} else if len(volName) == 0 {
+		return nil, status.Errorf(codes.Internal, "Volume name is a required parameter.")
+	} else if len(secrets[AttributeFirstClassDiskParentType]) == 0 {
```
codenrhoden: Ah, yeah, this should be coming from the request `Parameters` field, not secrets.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
...
 func (c *controller) DeleteVolume(
 	ctx context.Context,
 	req *csi.DeleteVolumeRequest) (
 	*csi.DeleteVolumeResponse, error) {
 
-	return nil, nil
+	secrets := req.GetSecrets()
+
+	//check for required parameters
+	if secrets == nil {
```
codenrhoden: secrets should still not be a required field for us. We're not going to use it, as all the credentials we need will be part of configuring the plugin.

I definitely see a problem, though, so want to mention here again. The CSI spec seems to assume that knowing the volume ID is enough to delete it. For us, I don't think that's the case (at least, not currently). I"m not sure how to handle this if it's the case that we need to know the parent type and parent name -- nothing in the CO is going to pass that to us, and it's not going to be in `secrets` either.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
...
+	}
+
+	// Volume Type
+	datastoreType := vclib.TypeDatastoreCluster
+	volType := secrets[AttributeFirstClassDiskParentType]
+	if volType == string(vclib.TypeDatastore) {
+		datastoreType = vclib.TypeDatastore
+	}
+
+	datastoreName := secrets[AttributeFirstClassDiskParentName]
```
codenrhoden: it may be the case that since we are only supporting a single datastore at this time, that this is a "fixed" parameter for the plugin (via config) for now. Definitely will have to think about this, but seems to be a limitation of the spec right now.

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,133 @@ func (c *controller) CreateVolume(
...
+		return nil, status.Errorf(codes.Internal, "Failed to retrieve VC/DC based on zone %s. Err: %v", zone, err)
+	}
+
+	firstClassDisk, err := discoveryInfo.DataCenter.GetFirstClassDisk(
+		ctx, datastoreName, datastoreType, volName, vclib.FindFCDByName)
+	if err == nil {
+		log.Warningf("Volume with name %s already exists. Checking for similar parameters.", volName)
+
+		if firstClassDisk.Config.CapacityInMB != volSizeMB {
+			return nil, status.Errorf(codes.Internal,
```
codenrhoden: this should be `codes.AlreadyExists`

dvonthenen: Will update

dvonthenen:
Will be providing additional updates based on conversations... for now this PR is still a WIP

pkg/csi/service/fcd/controller.go
dvonthenen: Switched to params for create

pkg/csi/service/fcd/controller.go
dvonthenen: using new method WhichVCandDCByFCDId

dvonthenen:
Implements WhichVCandDCByFCDId and removes use of secrets... ready for review!

pkg/csi/service/fcd/controller.go
```diff
@@ -72,15 +75,123 @@ func (c *controller) CreateVolume(
...
+
+	// Volume Type
+	var datastoreName string
+	datastoreType := discoveryInfo.FCDInfo.ParentType
+	if datastoreType == vclib.TypeDatastore {
+		datastoreName = discoveryInfo.FCDInfo.DatastoreInfo.Info.Name
+	} else {
+		datastoreName = discoveryInfo.FCDInfo.StoragePodInfo.Summary.Name
+	}
+
```
dougm: Should `DeleteVolume` be idempotent too and just return if the disk doesn't exist?

pkg/common/vclib/datacenter.go
```diff
@@ -439,3 +581,65 @@ func (dc *Datacenter) GetAllFirstClassDisks(ctx context.Context) ([]*FirstClassD
...
+		glog.Errorf("GetAllDatastores failed. Err: %v", err)
+		return nil, err
+	}
+
+	for _, datastore := range datastores {
+		fcd, err := datastore.GetFirstClassDiskInfo(ctx, fcdID, FindFCDByID)
+		if err == nil {
+			glog.Infof("DoesFirstClassDiskExist(%s): FOUND", fcdID)
+			return fcd, nil
+		}
```
dougm: May want to return `err` here if `err != ErrNoDiskIDFound` or at least log the `err`.

pkg/common/vclib/datacenter_test.go
```diff
@@ -133,7 +133,7 @@ func TestDatacenter(t *testing.T) {
 		t.Error(err)
 	}
 
-	diskPath := ds.Path(avm.Name + "/disk1.vmdk")
+	diskPath := (ds.Datastore).Path(avm.Name + "/disk1.vmdk")
```
dougm: cosmetic: no need for parens there.

pkg/common/vclib/datacenter.go
dvonthenen: I didn't want to flood the logs with errors. If you have 100 datastores and the FCD doesnt exist, you will have 100 lines in the log. Was just better to log it once.

pkg/common/vclib/datacenter_test.go
dvonthenen: Yup... was left over from my debugging a while back

pkg/csi/service/fcd/controller.go
dvonthenen: Good call. If it doesn't exist, will just return success.

Issue #105, opened 2018-11-28
csi: implement Controller publish/unpublish


codenrhoden:
As part of implementing CSI spec, add capability for Controller{Publish,Unpublish}. This functionality reaches out the vCenter and attaches/detaches the volume to the requested node.

PR #114, opened 2018-12-07
Initial implementation for CSI Publish/Unpublish

davidvonthenen:
Initial implementation for Publish/Unpublish for vSphere CSI.
- Handles DatastoreClusters, VSAN datastores, and "traditional" Datastores

Other:
- changed existing DetachDisk to be idempotent

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/105

- tested CSI Publish/Unpublish FCD on DatastoreClusters, VSAN datastores, and "traditional" Datastores
- tested on k8s 1.11.3
- tested on vSphere 6.7U1

**Release note**:

### DIFF
--- a/pkg/csi/service/fcd/controller.go
+++ b/pkg/csi/service/fcd/controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	log "github.com/sirupsen/logrus"
 	"github.com/vmware/govmomi/units"
+	"github.com/vmware/govmomi/vim25/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	clientset "k8s.io/client-go/kubernetes"
@@ -84,12 +85,12 @@ func (c *controller) CreateVolume(
 
 	//check for required parameters
 	if params == nil {
-		msg := fmt.Sprintf("Create parameters is a required parameter.")
-		log.Errorf(msg)
+		msg := "Create parameters is a required parameter."
+		log.Error(msg)
 		return nil, status.Errorf(codes.Internal, msg)
 	} else if len(volName) == 0 {
-		msg := fmt.Sprintf("Volume name is a required parameter.")
-		log.Errorf(msg)
+		msg := "Volume name is a required parameter."
+		log.Error(msg)
 		return nil, status.Errorf(codes.Internal, msg)
 	} else if len(params[AttributeFirstClassDiskParentType]) == 0 {
 		msg := fmt.Sprintf("Volume parameter %s is a required parameter.", AttributeFirstClassDiskParentType)
@@ -157,6 +158,8 @@ func (c *controller) CreateVolume(
 
 	attributes := make(map[string]string)
 	attributes[AttributeFirstClassDiskType] = FirstClassDiskTypeString
+	attributes[AttributeFirstClassDiskVcenter] = discoveryInfo.VcServer
+	attributes[AttributeFirstClassDiskDatacenter] = discoveryInfo.DataCenter.Name()
 	attributes[AttributeFirstClassDiskName] = firstClassDisk.Config.Name
 	attributes[AttributeFirstClassDiskParentType] = string(firstClassDisk.ParentType)
 	if firstClassDisk.ParentType == vclib.TypeDatastoreCluster {
@@ -185,8 +188,8 @@ func (c *controller) DeleteVolume(
 
 	//check for required parameters
 	if len(req.VolumeId) == 0 {
-		msg := fmt.Sprintf("Volume ID is a required parameter.")
-		log.Errorf(msg)
+		msg := "Volume ID is a required parameter."
+		log.Error(msg)
 		return nil, status.Errorf(codes.Internal, msg)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -79,13 +76,6 @@ func configFromSimWithTLS(tlsConfig *tls.Config, insecureAllowed bool) (vcfg.Con
 	model.Service.TLS = tlsConfig
 	s := model.Service.NewServer()
 
-	// STS simulator
-	path, handler := sts.New(s.URL, vpx.Setting)
```
dougm: Need these for the LoginByToken test

pkg/csi/service/fcd/controller.go
```diff
@@ -224,15 +225,101 @@ func (c *controller) ControllerPublishVolume(
 	req *csi.ControllerPublishVolumeRequest) (
 	*csi.ControllerPublishVolumeResponse, error) {
 
-	return nil, nil
+	//check for required parameters
+	if len(req.VolumeId) == 0 {
+		msg := fmt.Sprintf("Volume ID is a required parameter.")
```
dougm: no need for `Sprintf` here or below.  can also s/Errorf/Error/

pkg/csi/service/fcd/controller.go
```diff
@@ -224,15 +225,101 @@ func (c *controller) ControllerPublishVolume(
...
+
+	vm, err := discoveryInfo.DataCenter.GetVMByDNSName(ctx, req.NodeId)
+	if err != nil {
+		log.Errorf("GetVMByDNSName(%s) failed. Err: %v", req.NodeId, err)
+		return nil, err
+	}
+
+	filePath := fcd.Config.Backing.(*types.BaseConfigInfoDiskFileBackingInfo).FilePath
+	options := &vclib.VolumeOptions{SCSIControllerType: vclib.PVSCSIControllerType}
+	diskUUID, err := vm.AttachDisk(ctx, filePath, options)
```
dougm: Should this be idempotent?  Can check if the disk is attached like so:
```go
                devices, err := vm.Device(ctx)
		disk := devices.SelectByBackingInfo(&types.VirtualDiskFlatVer2BackingInfo{
			VirtualDeviceFileBackingInfo: types.VirtualDeviceFileBackingInfo{
				FileName: filePath,
			},
		})
                if len(disk) != 0 { // disk is already attached }
```

pkg/csi/service/fcd/controller.go
```diff
@@ -224,15 +225,101 @@ func (c *controller) ControllerPublishVolume(
...
 
 func (c *controller) ControllerUnpublishVolume(
 	ctx context.Context,
 	req *csi.ControllerUnpublishVolumeRequest) (
 	*csi.ControllerUnpublishVolumeResponse, error) {
 
-	return nil, nil
+	//check for required parameters
+	if len(req.VolumeId) == 0 {
+		msg := fmt.Sprintf("Volume ID is a required parameter.")
```
dougm: same here, no need for the f's

pkg/csi/service/fcd/controller.go
```diff
@@ -224,15 +225,101 @@ func (c *controller) ControllerPublishVolume(
...
+	fcd := discoveryInfo.FCDInfo
+
+	vm, err := discoveryInfo.DataCenter.GetVMByDNSName(ctx, req.NodeId)
+	if err != nil {
+		log.Errorf("GetVMByDNSName(%s) failed. Err: %v", req.NodeId, err)
+		return nil, err
+	}
+
+	filePath := fcd.Config.Backing.(*types.BaseConfigInfoDiskFileBackingInfo).FilePath
+	err = vm.DetachDisk(ctx, filePath)
```
dougm: Can also use `SelectByBackingInfo` here to check if the disk is already detached.

Gopkg.lock
```diff
@@ -1193,18 +1191,15 @@
     "github.com/spf13/pflag",
     "github.com/vmware/govmomi",
     "github.com/vmware/govmomi/find",
-    "github.com/vmware/govmomi/lookup/simulator",
```
dougm: Removing these imports is causing CI to fail.

pkg/csi/service/fcd/controller.go
codenrhoden: Yes, this should be idempotent.

pkg/cloudprovider/vsphere/vsphere_test.go
dvonthenen: What's weird is `make test` works on my local setup... hmmmm. Will revert.

pkg/csi/service/fcd/controller.go
dvonthenen: yea... it is just from copying an pasting to make it easier.

pkg/csi/service/fcd/controller.go
dvonthenen: The idempotency takes place inside AttachDisk:
`	// If disk is already attached, return the disk UUID
	if attached {
		diskUUID, _ := vm.Datacenter.GetVirtualDiskPage83Data(ctx, vmDiskPath)
		return diskUUID, nil
	}`

pkg/csi/service/fcd/controller.go
dvonthenen: Yup same copy and paste

pkg/csi/service/fcd/controller.go
dvonthenen: The idempotency takes place into DetachDisk:
```
	if device == nil {
		glog.Warningf("No virtual device found with diskPath: %q on VM: %q", vmDiskPath, vm.InventoryPath)
		return nil
	}
```
Returning nil is the disk is not attached.

Gopkg.lock
dvonthenen: Yup, was trying to removing unnecessary code but went a little too far it seems.

dvonthenen:
Please see comments above about where the idempotency takes place. It occurs within the AttachDisk/DetachDisk functions and not outside in controller.go.

pkg/csi/service/fcd/controller.go
dougm: Ah yes, that's the first change in this PR, I had looked at DetachDisk in the current tree.

pkg/csi/service/fcd/controller.go
dougm: Could swear I checked first, but yeah can clearly see AttachDisk is idempotent.  And in looking a second time, I noticed there are VirtualMachine AttachDisk/DetachDisk API methods that were added in 6.5.  This may simplify things a bit, but probably functionally the same - I'll open an issue we can look closer later.

pkg/csi/service/fcd/controller.go
```diff
@@ -84,11 +85,11 @@ func (c *controller) CreateVolume(
 
 	//check for required parameters
 	if params == nil {
-		msg := fmt.Sprintf("Create parameters is a required parameter.")
+		msg := "Create parameters is a required parameter."
 		log.Errorf(msg)
```
dougm: can also change all the `Errorf` to `Error`

dvonthenen: Will fix

dvonthenen:
Changed the Errorf -> Error as needed.

codenrhoden:
LGTM! 👍

PR #112, opened 2018-12-06
Implement CSI Node Stage/Unstage

codenrhoden:
This patch adds logic for CSI Node staging and unstaging. This takes a
pre-attached block volume, and mounts it to the given path in the
staging request. `gofsutil` is used to take care of the formatting,
mounting, and parsing of the mount tables.

**Which issue this PR fixes**: fixes #106

**Special notes for your reviewer**: Tested this with node-only funcitonally. An FCD was previously attached to a node, and tested that I could idempotently stage and unstage that volume to a pre-created staging directory.

**Release note**:

```release-note
NONE
```

PR #116, opened 2018-12-07
CSI Node Publish/Unpublish

### LINKED ISSUE #107: csi: Implement Node Publish/Unpublish
Implement Node{Publish,Unpublish}Volume.

This API bind-mounts the given volume from the staged location to the workload location.

codenrhoden:
This patch implements NodePublishVolume and NodeUnpublishVolume.

The code assumes the volume has already been staged, since we support
the NODE_STAGE_UNSTAGE capability, and it is up to the CO to ensure
these semantics are followed. The volume is then bind-mounted from the
staging target to the target. The SP (storage plugin) does create the
target dir, and removes it upon unpublish, per the spec.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #107 

This PR needs to be rebased after the PR for node stage/unstage is merged (#112)

Tested on a node that had an FCD attached, then use the `csc` tool to repeated publish and unpublish the volume to various targets (testing for idempotency and ability to publish the same volume to multiple targets).

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/csi/service/node.go
+++ b/pkg/csi/service/node.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/akutz/gofsutil"
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


pkg/csi/service/node.go
```diff
@@ -360,3 +452,100 @@ func contains(list []string, item string) bool {
...
+	if os.IsNotExist(err) {
+		if err := os.Mkdir(path, 0750); err != nil {
+			log.WithField("dir", path).WithError(
+				err).Error("Unable to create dir")
+			return false, err
+		}
+		log.WithField("path", path).Debug("created directory")
+		return true, nil
+	}
+	if !st.IsDir() {
```
dougm: May want to check if `if err != nil { return err }` here, else `st` could be nil.

codenrhoden:
@dougm updated from your feedback! thanks -- that code is almost 9 months old from an old project. =)

PR #118, opened 2018-12-07
Fix cloudprovider test races

dougm:
GuestFullName is the OS type, which is the same for all VMs. This means the FindByDnsName impl could randomly return the wrong VM.
Change to using the vm.Name which is unique in these test cases.

DiscoverNode had a race where wg.Done() could be called before wg.Add(), causing wg.Wait() to return earlier than expected.


dougm:
/assign @dvonthenen 

cc @codenrhoden

dvonthenen:
ahhhhhhh!!!

/lgtm

dvonthenen:
Most of the time, the VM would be nil

PR #119, opened 2018-12-08
More fixes for cloudprovider test races

dougm:
- Protect read/write access to VSphereInstance.Conn field (in-tree does the same)

- Avoid use of runtime.SetFinalizer in tests as it randomly kicks in, logging out client sessions

- Fixup error message when vm search fails

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -42,17 +42,20 @@ func init() {
 		if err != nil {
 			return nil, err
 		}
-		return newVSphere(cfg)
+		return newVSphere(cfg, true)
 	})
 }
 
 // Creates new Controller node interface and returns
-func newVSphere(cfg vcfg.Config) (*VSphere, error) {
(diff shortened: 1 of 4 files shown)


dougm:
/assign @dvonthenen 

cc @codenrhoden

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -42,17 +42,20 @@ func init() {
...
 
 // Creates new Controller node interface and returns
-func newVSphere(cfg vcfg.Config) (*VSphere, error) {
+func newVSphere(cfg vcfg.Config, finalize ...bool) (*VSphere, error) {
 	vs, err := buildVSphereFromConfig(cfg)
 	if err != nil {
 		return nil, err
 	}
-	runtime.SetFinalizer(vs, logout)
+	if len(finalize) == 1 && finalize[1] {
```
codenrhoden: does this need to be `finalize[0]`?  If len == 1, there will be only 1 element.

PR #124, opened 2018-12-17
Initial release of CSI YAML + Page83Data

dvonthenen:
This provides YAML to deploy the vsphere-csi controller and node components for the driver. It also provides documentation updates to guide a user to test a basic pod for reference.

An additional change was made:
- provide the Page83Data in the PublishContext that will be needed by the node component to mount the appropriate device to the VM.
- NodeGetInfo was changed to use os.Hostname() and not getSystemUUID() because k8s uses the Hostname to uniquely identify nodes. Please see past PRs on the CCM.

Was tested on:
k8s 1.13.0 cluster
vsphere 6.7u1


dvonthenen:
CC @codenrhoden @dougm

dvonthenen:
I should note that if you actually want to test this before we tag with `v0.1.1` use `latest` in the YAML for the vsphere-csi image.

docs/deploying_csi_vsphere_with_rbac.md
```diff
@@ -117,52 +120,114 @@ Create the secret by running the following command:
...
-
-You have two options for deploying `csi-vsphere`. It can be deployed either as a simple Pod or in a DaemonSet. There really isn't much difference between the two other than the DaemonSet will do leader election and the Pod will just assume to be the leader.
+#### 5. Deploy `csi-vsphere-controller`
 
 **IMPORTANT NOTES:**
-- Deploy either as a Pod or in a DaemonSet, but *DO NOT* deploy both.
-- The YAML to deploy as a Pod or a DaemonSet assume that your Kubernetes cluster was deployed using [kubeadm](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/). If you deployed your cluster using alternate means, you will need to modify the either of the YAML files in order to provided necessary files or paths based on your deployment.
-
-##### Deploy `csi-vsphere` as a Pod
+- The YAML to deploy CSI controller assumes that your Kubernetes cluster was deployed using [kubeadm](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/). If you deployed your cluster using alternate means, you will need to modify the either of the YAML files in order to provided necessary files or paths based on your deployment.
```
codenrhoden: > you will need to modify the either of the YAML files in order to provided necessary files or paths based on your deployment.

This may need some cleanup -- not sure what it is saying here.

docs/deploying_csi_vsphere_with_rbac.md
```diff
@@ -117,52 +120,114 @@ Create the secret by running the following command:
...
+The YAML to deploy `csi-vsphere-node` can be found in [vsphere-csi-node-ds.yaml](https://github.com/kubernetes/cloud-provider-vsphere/raw/master/manifests/csi/vsphere-csi-node-ds.yaml).
 
 Run the following command:
 
 ```bash
-[k8suser@k8master ~]$ kubectl create -f vsphere-csi-ds.yaml
+[k8suser@k8master ~]$ kubectl create -f vsphere-csi-node-ds.yaml
+```
+
+#### 7. Create CRDs for CSI driver registration
```
codenrhoden: This is where I need to read more of the CSI docs. I really thought the sidecars were supposed to take care of this. @vladimirvivien ?

docs/deploying_csi_vsphere_with_rbac.md
```diff
@@ -117,52 +120,114 @@ Create the secret by running the following command:
...
+
+To apply them to your Kubernetes cluster, run the following command:
+
+```bash
+[k8suser@k8master ~]$ kubectl create -f vsphere-csi-crd.yaml
+```
+
+#### 8. Create your StorageClass and PersistentVolumeClaim by Example
+
+Each StorageClass (SC) is going to be unique to each user as it depends on the vSphere configuration you have. The PersistentVolumeClaim (PVC) is also therefore unique since the PVC depends on the SC. You can find examples of each in the [manifests/csi](https://github.com/kubernetes/cloud-provider-vsphere/tree/master/manifests/csi) directory for reference. The important thing to note in the [StorageClass](https://github.com/kubernetes/cloud-provider-vsphere/tree/master/manifests/csi/example-vsphere-sc.yaml) as seen below is that you need to provide as paramters the type of datastore you will be using (`DatastoreCluster` or `Datastore`) and it's corresponding name.
```
codenrhoden: I'm not sure I agree the SC is unique per user. Each SC should define a specific datastore or datastore cluster that is accessible via the credentials given to the CSI plugin. But that doesn't mean that SC can't be used by multiple users, does it?

manifests/csi/vsphere-csi-node-ds.yaml
```diff
@@ -0,0 +1,79 @@
...
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /csi/
+            - name: registration-dir
+              mountPath: /registration/
+        - name: vsphere-csi-node
+          image: gcr.io/cloud-provider-vsphere/vsphere-csi:v0.1.1
+          env:
+          - name: VSPHERE_KUBE_CONFIG
```
codenrhoden: I don't think this ENV var will be necessary for the Node Service.

docs/deploying_csi_vsphere_with_rbac.md
dvonthenen: It basically means the default locations for various things like the kubernetes clusters certs and etc are where kubeadm placed them. If you rolled your cluster by hand or used some other alternate method, you need to update the YAML to reflect those locations.

docs/deploying_csi_vsphere_with_rbac.md
dvonthenen: They are required because they provide the look up mechanism for the driver. I went as far not installing these CRDs and the driver absolutely doesn't not function without them.

docs/deploying_csi_vsphere_with_rbac.md
dvonthenen: When I mean user... I mean the vsphere/k8s admin. They are unique because every "user" will have a unique names for their own datastores.

manifests/csi/vsphere-csi-node-ds.yaml
dvonthenen: Yup, you are correct. Not needed. Just a copy and paste issue. WIll remove it.

docs/deploying_csi_vsphere_with_rbac.md
codenrhoden: Well that's weird -- they are supposed to be optional. Oh well -- we want to make use of them.

docs/deploying_csi_vsphere_with_rbac.md
codenrhoden: I was more picking on the grammar here, `the either of` and `to provided necessary` both are off...  So I would suggest:

> If you deployed your cluster using alternate means, you will need to modify the YAML files in order to provide necessary files or paths based on your deployment.

docs/deploying_csi_vsphere_with_rbac.md
dvonthenen: I only have a portion of the error that I was using to search for the reason... it starts out with:
```
the server could not find the requested resource (get csinodeinfos.csi.storage.k8s.io <the node FQDN>)
```
then spews more junk after it. This error occurs in the driver_registrar sidecar container.

PR #126, opened 2018-12-18
Update CSI container image with FS utilities

codenrhoden:
Install utilities for ext2/3/4, XFS, and BtrFS, along with lsblk.

`toybox` has to be uninstalled from Photon 2.0, otherwise it conflicts with `e2fsprogs` and others.

I considered also installing `ca-certificates`, but chose not to since our YAML examples mount `/etc/ssl/certificates` from the host.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cluster/images/csi/Dockerfile
+++ b/cluster/images/csi/Dockerfile
@@ -12,6 +12,16 @@
 
 FROM photon:2.0
 
+RUN tdnf -y remove toybox
+
+RUN tdnf -y install \
+  util-linux \
+  e2fsprogs \
+  xfsprogs \
+  btrfs-progs
+
+RUN tdnf clean all
+
 ADD vsphere-csi /bin/
 
 CMD ["/bin/vsphere-csi"]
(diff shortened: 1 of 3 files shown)


codenrhoden:
This takes the image from 80.1MB to 131MB.

dvonthenen:
Working on a fix for...
```
Warning  FailedMount             <invalid> (x3 over <invalid>)  kubelet, k8sworker2.local  MountVolume.SetUp failed for volume "pvc-8b26a1ac-0309-11e9-bf31-005056abd481" : rpc error: code = InvalidArgument desc = exceeds size limit: TargetPath: max=128, size=131
```

codenrhoden:
Added a second commit to this PR that disables request validation against the CSI spec.

The error that David saw is valid -- K8s is violating the spec by sending a target path that is > 128 bytes. See https://github.com/container-storage-interface/spec/blob/v1.0.0/spec.md#size-limits.

🤷‍♂️

codenrhoden:
@vladimirvivien you may be interested in the error seen here. I'll have to reach out on Slack, or maybe just file an issue, that K8s is violating the spec in this case.

PR #127, opened 2019-01-10
Refactor Config package to allow for specialization

dvonthenen:
Refactor pkg/common/config to allow for specialization or customization of a config file based on the components need. This allows a component to have a configuration that only pertains to it without requiring other components config to be aware of it.

An example of how this could be done is by creating a Config object that wraps the common one:
```
import (
 vcfg "k8s.io/cloud-provider-vsphere/pkg/common/config
)

struct MyComponent {
  vcfg.Config
  OtherField string
}
```

Was tested on:
k8s 1.13.1 cluster with 10 worker nodes
vSphere 6.7u1

**Release note**:
This is only a refactor of existing code and the changes to CCM and CSI functionality does not impact the end user, documentation, or etc.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -34,7 +34,8 @@ import (
 )
 
 func ParseConfig(configFile string) (config.Config, error) {
-	var cfg config.Config
+	cfg := config.Config{}
+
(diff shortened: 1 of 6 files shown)


dvonthenen:

Issue #79, opened 2018-10-09
csi: support for zone-local, non-shared datastores


codenrhoden:

dougm:
See also: https://github.com/kubernetes/kubernetes/issues/67703

codenrhoden:
/close
This was implemented by #136

PR #136, opened 2019-01-30
Initial Implementation for CSI multi-VC/DC Support

dvonthenen:
This implements mutli-vCenter and multi-Datacenter support in the CSI driver components.
This also indirectly implements the second half of Zones support for CSI by way of supporting multi-VC/DC and adding code to support CreateVolume in a particular region/zone through the use of the native kubernetes topology support in CSI/k8s.

Some notable changes:
- Enables VOLUME_ACCESSIBILITY_CONSTRAINTS to allow the topology to get passed to the CSI driver. Note that the native topology support works for 1.12.X and higher. Capability exists in code to support 1.11.X and below.
- Moved implementation of nodemanager.DiscoverNode() into the pkg/common/connectionmanager to be reused by other components which is why you see a lot of code deletion in `nodemanager.go`.
- As a part of the move DiscoverNode() in the CM, add a `list.go` (example: grab all DCs in a config), `search.go` (where DiscoverNode() now lives along with DiscoverFcd()), and `zones.go` (handles object look up, like Datacenters or clusters, based on tags/zones) to be reused by CCM and CSI which is why you see a lot of additional code/files in `connectionmanager.go`
- refactored the Zones Interface in the CCM to make use of the common code above which is why you see a lot of the deletion of code in `zones.go`
- implements GLOG logging in the CSI components in `controller.go` based on CSI Debug being enabled. YAML is updated as well.
- updated CSI YAML to remove unnecessary properties (SSL config, kubeconfig, etc) that were actually interfering with permissions related to the CSI service account 
- updated CSI YAML to put place holders for zones

**TODOs**:
I have not created the images I plan on posting in the  docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md documentation, but I will go back and fill that out later. I thought it was better to get this PR out for review and update with the diagrams/pictures later.

https://github.com/vmware/cna-upstream-planning/issues/323

Tested CCM on:
k8s 1.13.1 cluster with 10 worker nodes in 3 zones
vSphere 6.7u1
Deployed a test pod which targeted a particular zone using a persistent volume
`make test` passes

**Release note**:
The single VC/DC configuration should be backward compatible and all older single VC/DC based vsphere.conf should still work with this updated code.


docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -0,0 +1,187 @@
...
+
+There are many options for creating vSphere tags. One such method would be to use [govc](https://github.com/vmware/govmomi/tree/master/govc). All the examples below will make use of this method. You could also create tags by accessing the vSphere REST APIs directly or by using the vSphere UI.
+
+Using the example above, if it is decided that `k8s-region` and `k8s-zone` are to be used for your Category labels, then you can create those vSphere Categories using `govc` by running the following command:
+```bash
+[k8suser@k8master ~]$ GOVC_URL=https://REPLACE_VSPHERE_USERNAME:REPLACE_VSPHERE_PASSWORD@REPLACE_VSPHERE_IP/sdk GOVC_INSECURE=TRUE ./govc tags.category.create -d "Kubernetes region" k8s-region
```
frapposelli: Instead of using inline env vars, why not exporting them at the beginning of the example? maybe removing `GOVC_INSECURE=TRUE` which is not best practice.

docs/deploying_csi_vsphere_with_rbac.md
```diff
@@ -2,6 +2,18 @@
 
 This document is designed to quickly get you up and running with the `csi-vsphere`.
 
+## Requirements
+
+The vSphere Container Storage Interface (CSI) driver depends on First Class Disks (FCDs) and as such, has the following requirements:
```
frapposelli: I would explicitly state that versions prior to vSphere 6.5 are not supported

manifests/csi/vsphere-csi-controller-ss.yaml
```diff
@@ -43,45 +43,26 @@ spec:
...
-            - name: VSPHERE_KUBE_CONFIG
-              value: "*"
             - name: CSI_ENDPOINT
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
             - name: X_CSI_MODE
               value: "controller"
             - name: X_CSI_SPEC_REQ_VALIDATION
               value: "false"
+            - name: GLOG_LEVEL
+              value: "2"
```
frapposelli: is this the default log value? these manifests are the "easy button" way of doing deployment, they should be minimal and not explicit default values

manifests/csi/vsphere-csi-controller-ss.yaml
```diff
@@ -43,45 +43,26 @@ spec:
...
+#            - name: X_CSI_DEBUG
+#              value: "true"
```
frapposelli: leftover comment?

docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md
dvonthenen: Will do.

docs/deploying_csi_vsphere_with_rbac.md
dvonthenen: Will do.

pkg/csi/service/fcd/controller.go
```diff
@@ -56,6 +58,16 @@ func New() vTypes.Controller {
 }
 
 func (c *controller) Init(config *vcfg.Config) error {
+	glogLevel := os.Getenv(GLOG_LEVEL)
```
codenrhoden: Hmm. I'm not sure I like having to check for this as well. I see why we are doing it, because the common components are using glog, but I wonder if we could instead trigger the below logic based of the X_CSI_LOG_LEVEL field, and or the X_CSI_DEBUG. Like, `if X_CSI_DEBUG == true; set glog = 5`, stuff like that. Set the glog level automatically.

pkg/csi/service/fcd/controller.go
```diff
@@ -118,11 +130,11 @@ func (c *controller) CreateVolume(
 	}
 
 	datastoreName := params[AttributeFirstClassDiskParentName]
+	zone := params[AttributeFirstClassDiskZone]
+	region := params[AttributeFirstClassDiskRegion]
```
codenrhoden: This is an area where I need to do more reading, and see how things are implemented in the sidecars, but I would not expect zone/region information to come in as attributes... I would have expected them to be in the `TopologyRequirement` within the [spec](https://github.com/container-storage-interface/spec/blob/v1.0.0/spec.md#createvolume)

docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -0,0 +1,186 @@
+# Deploying the vSphere CCM and CSI in a Multi-vCenter OR Multi-Datacenter Environment using Zones
+
+This document is designed to quickly get you up and running in a vSphere configuration that consists of multiple vCenter or a multiple Datacenter environment via using zones.
+
+## Dependent Requirements
```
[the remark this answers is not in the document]
dvonthenen: Sure. Will change.

docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Why Do We Need to Use Zones in a Multi-vCenter or Multi-Datacenter Environment
+
+There exist 2 significant issues when deploying Kubernetes workloads or pods in a mutli-vCenter or single vCenter with multiple Datacenters. They are:
+1. Datastore objects, specifically names and even morefs (Managed Object References), are not unique across vCenters instances
+2. Datastore objects, specifically names, are not unique within a single vCenter since objects of the same name can exist in different Datacenters
+
+TODO: IMAGE
+
+There needs to be a mechanism in place to allow end-users continue to use the human readable "friendly" names for objects like datastores and datastore clusters and still be able to target workloads to use resources from them. This is where the concept of zones or zoning comes in. Zones allow you to partition datacenters and compute clusters so that the end-user can target workloads to specific locations in your vSphere environment.
```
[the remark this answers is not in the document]
dvonthenen: Fixing.

pkg/csi/service/fcd/controller.go
dvonthenen: Without setting GLOG to 2, there is absolutely zero logging and no way to debug or even see what is going on.

pkg/csi/service/fcd/controller.go
dvonthenen: I will double check to see if this is already being sent over. If it is, I will delete that and pull from the appropriate place and update the documentation.

pkg/csi/service/fcd/controller.go
codenrhoden: Oh, I agree with you there.  I'm saying rather than support `GLOG_LEVEL` as an env var for the CSI plugin, we do the following:

```go
flag.Set("logtostderr", "true")
flag.Set("stderrthreshold", "INFO")

lvl := log.GetLevel()
if lvl == log.Debug
  flag.Set("v", "5")
if lvl == log.Info
  flag.Set("v", "4")

flag.Parse()
```

Something like that (with whatever is the right mapping from the logrus log levels to the glog levels)

The above psuedocode works (just reading the exsting log level) because GoCSI does [this](https://github.com/rexray/gocsi/blob/9407f5f4ed38592fe11b57c1c9d3621df05b753b/gocsi.go#L39-L56) to set it at the beginning.

Also, I think this code should live at service/service.go:BeforeRun() [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/csi/service/service.go#L94) instead of in fcd/controller.go. It should be a common thing that always happens.

docs/deploying_ccm_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -0,0 +1,186 @@
...
+  nodeSelector:
+    failure-domain.beta.kubernetes.io/region: IF_USING_ZONES_REPLACE_WITH_TARGETED_REGION_VALUE
+    failure-domain.beta.kubernetes.io/zone: IF_USING_ZONES_REPLACE_WITH_TARGETED_ZONE_VALUE
+  volumes:
+    - name: my-fcd-volume
+      persistentVolumeClaim:
+        claimName: vsphere-csi-pvc
+```
+
+*IMPORTANT*: Just to re-emphasize topics discussed in this document, the datastore or datastore cluster that the persistent volume is to be provisioned from must be available to the `region` and `zone` and by all hosts within that `region` and `zone` pairing since kubernetes is what is performing the scheduling of pods.
```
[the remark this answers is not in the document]
dvonthenen: Fixed

pkg/csi/service/fcd/controller.go
dvonthenen: Yup, gotcha now... will update

pkg/csi/service/service.go
```diff
@@ -95,9 +96,22 @@ func (s *service) BeforeServe(
 	// Get the SP's operating mode.
 	s.mode = csictx.Getenv(ctx, gocsi.EnvVarMode)
 
+	// Set glog level based on CSI debug being enabled
+	var glogLevel string
+	csiDebug := csictx.Getenv(ctx, gocsi.EnvVarMode)
```
codenrhoden: This doesn't get you the DEBUG flag, it gets you the MODE. But like I put in the example of my other comment, you don't need to go read the ENV var here, because GoCSI has already read and parsed the logger's level. Just do:

```go
glogLevel := "2"
lvl := log.GetLevel()
if lvl = log.DebugLevel {
  glogLevel = "4"
}
```

That would be my approach.

pkg/csi/service/fcd/controller.go
dvonthenen: i dumped the topology fields in the Create request…. and nothing is set. seems like a bug on the kubernetes side. until that is fixed, we need those labels.

codenrhoden:
I believe this PR would also take care of #79, right?

manifests/csi/vsphere-csi-controller-ss.yaml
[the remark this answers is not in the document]
dvonthenen: Already did

pkg/csi/service/service.go
dvonthenen: ok will update

manifests/csi/vsphere-csi-controller-ss.yaml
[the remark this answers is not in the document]
dvonthenen: Yup already did

dvonthenen:
> I believe this PR would also take care of #79, right?

I believe so based on a verbal convo for what the issue means

dvonthenen:
Looking into one other method for transfering the zone info to CSI Create...

codenrhoden:
Fixes #79

dvonthenen:
After checking into the CSI topology info/labels (comment above), soliciting for more feedback.

codenrhoden:
> After checking into the CSI topology info/labels (comment above), soliciting for more feedback.

Yeah, looks like the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190124-csi-topology.md) for CSI topology is implementable, but unstarted. I should have just checked there first. :)

dvonthenen:
> This is looking good!
> 
> My main question is about how zone/region is passed on VolumeCreate from the provisioner sidecar. I would not expect these to come in as attributes, I thought they would end up in the TopologyRequirement field, since that is what that is for.



> > After checking into the CSI topology info/labels (comment above), soliciting for more feedback.
> 
> Yeah, looks like the [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190124-csi-topology.md) for CSI topology is implementable, but unstarted. I should have just checked there first. :)

No worries. Trying it out is a "for sure" validation.

dvonthenen:
Revisiting that topology info on Volume Create... seems like there may be additional stuff to enable.

pkg/csi/service/fcd/constants.go
```diff
@@ -38,4 +38,9 @@ const (
 	AttributeFirstClassDiskVcenter         = "vcenter"
 	AttributeFirstClassDiskDatacenter      = "datacenter"
 	AttributeFirstClassDiskPage83Data      = "page83data"
+	AttributeFirstClassDiskZone            = "zone"
+	AttributeFirstClassDiskRegion          = "region"
+
+	//Additional ENVVARs
+	GLOG_LEVEL = "GLOG_LEVEL"
```
codenrhoden: Now that this is set automatically, we don't need this anymore.

pkg/csi/service/fcd/controller.go
```diff
@@ -25,13 +25,15 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	log "github.com/sirupsen/logrus"
-	"github.com/vmware/govmomi/units"
-	"github.com/vmware/govmomi/vim25/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	clientset "k8s.io/client-go/kubernetes"
+	apis "k8s.io/kubernetes/pkg/kubelet/apis"
```
codenrhoden: This is one is just a comment. I know I'm fighting a losing battle, but I hate seeing imports from `k8s.io/kubernetes`. `k8s.io/client-go` makes sense, but packages from kubelet? That's a lot of stuff to pull in just to get label names/string constants.

pkg/csi/service/fcd/constants.go
dvonthenen: Yep will remove it

pkg/csi/service/fcd/controller.go
dvonthenen: Yea, I agree. Only way to make sure that the label doesn't change from underneath us would be to import... but the odds of them changing the label are practically zero. Too much other stuff would break. Will change.

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -55,13 +55,18 @@ func (i *instances) NodeAddresses(ctx context.Context, nodeName types.NodeName)
 		return node.NodeAddresses, nil
 	}
 
-	if err := i.nodeManager.DiscoverNode(string(nodeName), FindVMByName); err != nil {
-		glog.V(2).Infof("instances.NodeAddresses() could NOT FIND node %s err: %v", string(nodeName), err)
-		return []v1.NodeAddress{}, ErrNodeNotFound
+	if err := i.nodeManager.DiscoverNode(string(nodeName), FindVMByName); err == nil {
+		glog.V(2).Info("instances.NodeAddresses() FOUND with ", string(nodeName))
+		if i.nodeManager.nodeNameMap[string(nodeName)] == nil {
+			glog.Errorln("DiscoverNode succeeded, but CACHE missed.")
```
andrewsykim: I would group these two log lines together, it may overlap with other log lines get confusing. Also would add the node name in there somehow so it's clear which nodes need to be checked for case sensitivity

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -109,6 +114,11 @@ func (i *instances) InstanceID(ctx context.Context, nodeName types.NodeName) (st
 
 	if err := i.nodeManager.DiscoverNode(string(nodeName), FindVMByName); err == nil {
 		glog.V(2).Info("instances.InstanceID() FOUND with ", string(nodeName))
+		if i.nodeManager.nodeNameMap[string(nodeName)] == nil {
+			glog.Errorln("DiscoverNode succeeded, but CACHE missed.")
```
andrewsykim: same here, group together

pkg/csi/service/fcd/vsphere_helper.go
```diff
@@ -0,0 +1,87 @@
...
+*/
+
+package fcd
+
+import (
+	"sort"
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
```
andrewsykim: I like this package, but why break consistency and use logrus instead of glog here?

pkg/cloudprovider/vsphere/instances.go
dvonthenen: Yup makes sense

pkg/cloudprovider/vsphere/instances.go
dvonthenen: Yup makes sense

pkg/csi/service/fcd/vsphere_helper.go
dvonthenen: the gocsi package that does the code generation for the skeleton puts that there. i think multiple logging libraries is weird, but if we address that, we should do it in a following PR since the plumbing as existed for some time now.

We have a TODO which changes from glog to klog. We should think about this consolidation then.

pkg/csi/service/fcd/vsphere_helper.go
```diff
@@ -0,0 +1,87 @@
...
+
+	for vc, vsi := range cm.VsphereInstanceMap {
+
+		var err error
+		for i := 0; i < NUM_OF_CONNECTION_ATTEMPTS; i++ {
+			err = cm.ConnectByInstance(ctx, vsi)
+			if err == nil {
+				break
+			}
+			time.Sleep(time.Duration(RETRY_ATTEMPT_DELAY_IN_SECONDS) * time.Second)
```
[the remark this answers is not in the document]
dvonthenen: It's not. I was moving around code that was using it, but it isn't really needed anymore.

pkg/csi/service/fcd/vsphere_helper.go
andrewsykim: Ahh I see, thanks. I just opened https://github.com/kubernetes/cloud-provider-vsphere/issues/137 cause I didn't see any open issues for it :)

pkg/csi/service/fcd/vsphere_helper.go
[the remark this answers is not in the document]
dvonthenen: I take that back... I tried removing, but it is complaining about it. Maybe because it's in a const? Oh well... will just leave it.

```
[go] invalid operation: RETRY_ATTEMPT_DELAY_IN_SECONDS * time.Second (mismatched types int and time.Duration)
```

pkg/common/connectionmanager/custom_errors.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package connectionmanager
+
+import (
+	"errors"
+)
+
+// Error Messages
+const (
```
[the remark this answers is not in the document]
dvonthenen: It doesn't matter to me either way.... only reason why I did it was to remain consistent where it happens everywhere else in the code especially `pkg/common/vclib` which was lifted from k/k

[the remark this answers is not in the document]
dvonthenen: If anyone has a strong preference either way, we can change it.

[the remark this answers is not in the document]
andrewsykim: makes sense, dont' feel strongly here so feel free to ignore my comment :)

dvonthenen:
@andrewsykim Thanks for the feedback!

pkg/csi/service/fcd/vsphere_helper.go
codenrhoden: Yeah, this stems more from the idea that CSI plugins are separate from K8s, and the core GoCSI code uses `logrus` (and has since inception). It may be a moot point, but I vastly prefer `logrus` (and structured logging in general` to glog/klog. 🤷‍♂️ 

So it wasn't really about having two (I didn't realize until this PR that *any* of the code pulled into the CSI plugin had glog/klog code), it was that the CSI plugin uses `logrus` because `GoCSI` does. I'd really hate to change that.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -118,195 +103,50 @@ func (nm *NodeManager) removeNode(node *v1.Node) {
...
-					break
-				}
-
-				glog.V(4).Infof("Finding node %s in vc=%s and datacenter=%s", myNodeID, vc, datacenterObj.Name())
-				queueChannel <- &vmSearch{
-					vc:         vc,
-					datacenter: datacenterObj,
-				}
+	addrs := []v1.NodeAddress{}
+	for _, v := range oVM.Guest.Net {
```
dougm: Consider is using this method, which only returns IP address that have a corresponding vSphere virtual NIC device: https://github.com/vmware/govmomi/blob/e560c7bb1e709826a897c9da4baad0944c2cdda0/object/virtual_machine.go#L254-L259

Also see discussion in: https://github.com/kubernetes/kubernetes/pull/73721

dvonthenen:
> Haven't done a full review yet.. Other than the 1 comment, have you looked at testing against vcsim? Multi DCs are supported. Multi VCs we can probably simulate too.

I wasn't aware that vcsim could do multi-DC and VC. I will need to talk to you about this. I know one of the next major things we need to look at are the unit and e2e tests.

dougm:
> I wasn't aware that vcsim could do multi-DC and VC. I will need to talk to you about this. I know one of the next major things we need to look at are the unit and e2e tests.

Yeah, you can creating using the API or pre-populate with any number using the `Model`: https://godoc.org/github.com/vmware/govmomi/simulator#example-Model

You can also start any number of vcsim instances.  There is an issue if that's done in the same process (they'll have the same inventory), but that can be fixed.

I can help with implementing the tests for these cases too, hit me up.

/lgtm

Issue #137, opened 2019-02-05
Investigate using k8s.io/klog instead of glog


andrewsykim:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
We should investigate using k8s.io/klog instead of glog. `klog` is a fork of `glog` so we should validate that there are no breaking changes in the way we use glog if we were to switch.

dvonthenen:
There was also small additional discussion about consolidating to use a single logger. Currently using glog(->klog) and logrus. I know logrus is used in the gocsi framework/skeleton code that was created... so not entirely sure how feasible this is.

dvonthenen:
Implemented and currently testing...

dvonthenen:
PR was merged
/close

PR #140, opened 2019-02-12
Switch from using glog to klog

dvonthenen:
The kubernetes community is moving away from glog to klog. This updates all components (CCM, CSI, CLI) used to klog.

https://github.com/kubernetes/cloud-provider-vsphere/issues/137

Tested CCM and CSI on:
- vSphere 6.7u1

**Release note**:
Internal package change

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1086,6 +1086,14 @@
   version = "kubernetes-1.11.0"
 
 [[projects]]
+  digest = "1:72fd56341405f53c745377e0ebc4abeff87f1a048e0eea6568a20212650f5a82"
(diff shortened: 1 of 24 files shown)


dvonthenen:
CC: @fabio @dougm @codenrhoden

codenrhoden:
Looks good to me. My only question is why `glog` still shows up as a dependency in `Gopkg.lock`

dvonthenen:
klog is version 0.2.0 and glog is still present because there are subcomponents that depend on it. I read the klog docs… glog and klog can co-exist in the same project (and be vendored in together). I verify that we see log lines originating from both glog and klog.

Issue #139, opened 2019-02-06
csi: Check for min-VC and min-Cluster support


dvonthenen:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
The CSI driver currently only supports 6.5 and higher.

**What you expected to happen**:
Give a warning to the user if the VC or the cluster attached to a zone is pre-6.5 and produce an error. The number of configurations that can be created are infinite... so attempt to detect easy/simple unsupported configurations.

**How to reproduce it (as minimally and precisely as possible)**:
Use an older VC configuration (a negative test).

**Anything else we need to know?**:
Nope

**Environment**:
NA

dvonthenen:
PR merged
/close

PR #141, opened 2019-02-12
Check for min vCenter version

dvonthenen:
This checks the minimum vCenter API version (which is 6.5) on the CSI FCD controller. If unsupported, the controller will print an error message and will fail to start the controller service.

I had hopes that we could check clusters within DCs, but quickly realized that there could be clusters that are being used to manage non-Kubernetes earmarked clusters that can be of an older ESX version. Only check that can be done without going through a bunch of hoops is checking the vCenter API version. Docs already outline the support matrix for the CSI driver.

https://github.com/kubernetes/cloud-provider-vsphere/issues/139

Added go tests to check the boundary cases.

Tested CCM and CSI on:
- vSphere 6.7u1

**Release note**:
NA

### DIFF
--- a/pkg/common/connectionmanager/connectionmanager.go
+++ b/pkg/common/connectionmanager/connectionmanager.go
@@ -20,8 +20,8 @@ import (
 	"context"
 	"sync"
 
-	"k8s.io/klog"
 	"k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog"
 
 	vcfg "k8s.io/cloud-provider-vsphere/pkg/common/config"
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/credentialmanager"
@@ -187,4 +187,18 @@ func (cm *ConnectionManager) VerifyWithContext(ctx context.Context) error {
 		}
 	}
 	return nil
+}
+
+// APIVersion returns the version of the vCenter API
+func (cm *ConnectionManager) APIVersion(vcenter string) (string, error) {
+	if err := cm.Connect(context.Background(), vcenter); err != nil {
+		return "", err
+	}
+
+	instance := cm.VsphereInstanceMap[vcenter]
+	if instance == nil || instance.Conn.Client == nil {
+		return "", ErrConnectionNotFound
+	}
+
+	return instance.Conn.Client.ServiceContent.About.ApiVersion, nil
 }
--- a/pkg/csi/service/fcd/controller.go
+++ b/pkg/csi/service/fcd/controller.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 	volumeutil "k8s.io/kubernetes/pkg/volume/util"
 
 	"github.com/vmware/govmomi/units"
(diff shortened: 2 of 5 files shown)


dvonthenen:

pkg/csi/service/fcd/vsphere_helper.go
```diff
@@ -31,8 +33,35 @@ import (
 const (
 	NUM_OF_CONNECTION_ATTEMPTS     int = 3
 	RETRY_ATTEMPT_DELAY_IN_SECONDS int = 1
+
+	MINIMUM_SUPPORTED_VCENTER_MAJOR int = 6
+	MINIMUM_SUPPORTED_VCENTER_MINOR int = 5
 )
 
+func APICheck(version string) error {
```
frapposelli: I see this is exported, is it going to be used in other packages?

pkg/common/connectionmanager/connectionmanager.go
```diff
@@ -188,3 +188,17 @@ func (cm *ConnectionManager) VerifyWithContext(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (cm *ConnectionManager) APIVersion(vcenter string) (string, error) {
```
frapposelli: can you add a comment here? to keep `go vet` happy 😄

pkg/common/connectionmanager/connectionmanager.go
```diff
@@ -188,3 +188,17 @@ func (cm *ConnectionManager) VerifyWithContext(ctx context.Context) error {
...
+	err := cm.Connect(context.Background(), vcenter)
```
frapposelli: nit: can you make it terse with
```go
if err := cm.Connect(context.Background(), vcenter); err != nil {
```

pkg/csi/service/fcd/controller.go
```diff
@@ -71,6 +72,21 @@ func (c *controller) Init(config *vcfg.Config) error {
...
 
+	//VC check... FCD is only supported in 6.5+
+	for vc := range connMgr.VsphereInstanceMap {
+		api, err := connMgr.APIVersion(vc)
+		if err != nil {
+			klog.Errorf("APIVersion failed err=%v", err)
+			return err
+		}
+
+		err = APICheck(api)
```
frapposelli: nit: terse
```go
if err := APICheck(api); err != nil {
```

pkg/csi/service/fcd/vsphere_helper.go
dvonthenen: Nope... will rename the function. Was complaining that "API" was in lowercase.

frapposelli: maybe `checkAPI`?

pkg/common/connectionmanager/connectionmanager.go
dvonthenen: Will do

pkg/common/connectionmanager/connectionmanager.go
dvonthenen: Yup

pkg/csi/service/fcd/controller.go
dvonthenen: same

dvonthenen:
Updated based on feedback. Thanks @frapposelli

frapposelli:
cool stuff, thanks @dvonthenen !

/approve
/lgtm

Issue #142, opened 2019-02-13
CCM >v0.1.0 fails on Photon2 OS


akutz:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
Running CCM:latest on Photon2 OS results in a panic.

**What you expected to happen**:
No panic.

**How to reproduce it (as minimally and precisely as possible)**:
Use CCM:latest on Photon2 OS.

**Anything else we need to know?**:
When discovering node information, the CCM on Photon2 OS panics:

```shell
2019-02-14T00:28:28.77211952Z stderr F I0214 00:28:28.772000       1 flags.go:27] FLAG: --address="127.0.0.1"
2019-02-14T00:28:28.772228035Z stderr F I0214 00:28:28.772189       1 flags.go:27] FLAG: --allocate-node-cidrs="false"
2019-02-14T00:28:28.772298274Z stderr F I0214 00:28:28.772266       1 flags.go:27] FLAG: --allow-untagged-cloud="false"
2019-02-14T00:28:28.77236737Z stderr F I0214 00:28:28.772325       1 flags.go:27] FLAG: --alsologtostderr="false"
2019-02-14T00:28:28.772413658Z stderr F I0214 00:28:28.772389       1 flags.go:27] FLAG: --bind-address="0.0.0.0"
2019-02-14T00:28:28.772476189Z stderr F I0214 00:28:28.772444       1 flags.go:27] FLAG: --cert-dir="/var/run/kubernetes"
2019-02-14T00:28:28.772532703Z stderr F I0214 00:28:28.772497       1 flags.go:27] FLAG: --cidr-allocator-type="RangeAllocator"
2019-02-14T00:28:28.772613304Z stderr F I0214 00:28:28.772579       1 flags.go:27] FLAG: --cloud-config="/etc/cloud/cloud-provider.conf"
2019-02-14T00:28:28.772668646Z stderr F I0214 00:28:28.772645       1 flags.go:27] FLAG: --cloud-provider="vsphere"
2019-02-14T00:28:28.772731503Z stderr F I0214 00:28:28.772700       1 flags.go:27] FLAG: --cluster-cidr=""
2019-02-14T00:28:28.772775354Z stderr F I0214 00:28:28.772752       1 flags.go:27] FLAG: --cluster-name="kubernetes"
[...]

PR #143, opened 2019-02-14
Bug fix for NodeManager and Photon 2 UUID

dvonthenen:
This fixes:
- a crash in the NodeManager when the VM is not found via UUID introduced accidentally with this PR https://github.com/kubernetes/cloud-provider-vsphere/pull/136 Missing a return statement: https://github.com/kubernetes/cloud-provider-vsphere/pull/143/files#diff-b7f52468613fc2d2fc4b257bed84a46cR138
- a bug with the way Photon 2 reports the UUID which is then passed from the kubelet to the CCM. It does not conform to the way the RHEL, Ubuntu, and etc reports it. The reporting format of the UUID has been fixed in Photon 3 though.

https://github.com/kubernetes/cloud-provider-vsphere/issues/142

Added go tests to check UUID revert.

Tested CCM and CSI on:
- vSphere 6.7u1
- OS: RHEL

**Release note**:
NA

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -21,10 +21,10 @@ import (
 	"errors"
 	"net"
 
-	"k8s.io/klog"
 	"k8s.io/api/core/v1"
 	clientv1 "k8s.io/client-go/listers/core/v1"
 	pb "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/proto"
+	"k8s.io/klog"
 	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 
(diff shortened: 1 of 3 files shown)


dvonthenen:
Pending testing from @akutz on Photon 2.

akutz:
Hi @dvonthenen,

Thank you for such a quick fix. If I read the code correctly it looks good over here:

```shell
I0214 19:55:31.917002       1 nodemanager.go:114] Discovered VM using normal UUID format
```

The one thing I'd suggest is that the lookup order should be reversed. Given that other Linux distributions require the conversion of the UUID, why not default to that first and then use the original UUID if the first pass fails?

dvonthenen:
CC: @frapposelli @dougm

PR #147, opened 2019-02-22
Part 1: Add CSI go tests

dvonthenen:
We need more CSI tests! Tests added:
- Complete controller go test (Create, List, Publish, Unpublish, Delete)
- Test ListVolumes boundary conditions mainly around pagination
- Test ListVoumes order
- Test for Zones support using 2 Datacenters

Additional changes found while added go tests:
- updated govmomi to latest to fix a bug in device Key/UnitNumber in the older version
- Removed some klog.V(X) to always display since they are very useful debug statements to have all the time
- Changed some klog.V(X) values as some weren't needed as often and others were
- Found a bug in GetAllDatastoreClusters in which returns an error when no DatastoreClusters exist causing the entire ListVolumes to fail
- Changed ListVolumes to be on a 0-based index instead of 1-based. This is used for the NextToken when doing pagination of volumes. 1-based was too confusing.

NA

Tested CCM and CSI on:
k8s 1.13.2 cluster with 10 worker nodes in 3 zones
vSphere 6.7u1
Deployed a test pod which targeted a particular zone using a persistent volume
`make test` passes

**Release note**:
NA

### DIFF
--- a/pkg/csi/service/fcd/controller_test.go
+++ b/pkg/csi/service/fcd/controller_test.go
@@ -1,0 +1,621 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fcd
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"log"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	lookup "github.com/vmware/govmomi/lookup/simulator"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/simulator/vpx"
+	sts "github.com/vmware/govmomi/sts/simulator"
+	"github.com/vmware/govmomi/vapi/rest"
+	vapi "github.com/vmware/govmomi/vapi/simulator"
+	"github.com/vmware/govmomi/vapi/tags"
+
+	vcfg "k8s.io/cloud-provider-vsphere/pkg/common/config"
+	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
+)
+
+// configFromSim starts a vcsim instance and returns config for use against the vcsim instance.
+// The vcsim instance is configured with an empty tls.Config.
+func configFromSim(multiDc bool) (vcfg.Config, func()) {
+	return configFromSimWithTLS(new(tls.Config), true, multiDc)
+}
+
+// configFromSimWithTLS starts a vcsim instance and returns config for use against the vcsim instance.
+// The vcsim instance is configured with a tls.Config. The returned client
+// config can be configured to allow/decline insecure connections.
+func configFromSimWithTLS(tlsConfig *tls.Config, insecureAllowed bool, multiDc bool) (vcfg.Config, func()) {
+	var cfg vcfg.Config
+	model := simulator.VPX()
+
+	if multiDc {
+		model.Datacenter = 2
+		model.Datastore = 1
+		model.Cluster = 1
+		model.Host = 0
+	}
+
+	err := model.Create()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	model.Service.TLS = tlsConfig
+	s := model.Service.NewServer()
+
+	// STS simulator
+	path, handler := sts.New(s.URL, vpx.Setting)
+	model.Service.ServeMux.Handle(path, handler)
+
+	// vAPI simulator
+	path, handler = vapi.New(s.URL, nil)
+	model.Service.ServeMux.Handle(path, handler)
+
+	// Lookup Service simulator
+	model.Service.RegisterSDK(lookup.New())
+
+	cfg.Global.InsecureFlag = insecureAllowed
+
+	cfg.Global.VCenterIP = s.URL.Hostname()
+	cfg.Global.VCenterPort = s.URL.Port()
+	cfg.Global.User = s.URL.User.Username()
+	cfg.Global.Password, _ = s.URL.User.Password()
+	// Configure region and zone categories
+	if multiDc {
+		cfg.Global.Datacenters = "DC0,DC1"
+	} else {
+		cfg.Global.Datacenters = vclib.TestDefaultDatacenter
+	}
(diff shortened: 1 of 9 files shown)


pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	model.Service.TLS = tlsConfig
+	s := model.Service.NewServer()
+
+	// STS simulator
+	path, handler := sts.New(s.URL, vpx.Setting)
+	model.Service.ServeMux.Handle(path, handler)
+
+	// vAPI simulator
+	path, handler = vapi.New(s.URL, nil)
+	model.Service.ServeMux.Handle(path, handler)
```
dougm: Doesn't matter for this test, but you can use `model.Service.Handle` here.  Calls the same `ServeMux.Handle` underneath, but also connects the rest simulator to a soap simulator interface to support the [FCD tag APIs](https://pubs.vmware.com/vsphere-6-5/topic/com.vmware.wssdk.apiref.doc/vim.vslm.vcenter.VStorageObjectManager.html)

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	}
+	return configFromSim(multiDc)
+}
+
+type SearchIndex struct {
+	*simulator.SearchIndex
+	vm *simulator.VirtualMachine
+}
+
+func (s *SearchIndex) FindByDnsName(req *types.FindByDnsName) soap.HasFault {
```
dougm: With govmomi updated, that includes `FindByDnsName` support.  Though you still need to set `vm.Guest.HostName = vm.Name`, don't recall why we didn't do that by default.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	// Get a simulator DS
+	myds := simulator.Map.Any("Datastore").(*simulator.Datastore)
+
+	err := connMgr.Connect(ctx, config.Global.VCenterIP)
+	if err != nil {
+		t.Errorf("Failed to Connect to vSphere: %s", err)
+	}
+
+	search := object.NewSearchIndex(connMgr.VsphereInstanceMap[config.Global.VCenterIP].Conn.Client)
+	si := simulator.Map.Get(search.Reference()).(*simulator.SearchIndex)
```
dougm: You can use `simulator.Map.SearchIndex()` here instead.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	}
+
+	respCreate, err := c.CreateVolume(ctx, reqCreate)
+	if err != nil {
+		t.Fatalf("CreateVolume failed: %v", err)
+	}
+
+	volID := respCreate.Volume.VolumeId
+	volName := respCreate.Volume.VolumeContext[AttributeFirstClassDiskName]
+	log.Println("ID:", volID)
```
dougm: `t.Logf` instead so these only show up with `go test -v`.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+		}
+
+		volNameRet := respCreate.Volume.VolumeContext[AttributeFirstClassDiskName]
+		if !strings.EqualFold(volName, volNameRet) {
+			t.Errorf("[CREATE] Name of FCD does not match %s != %s", volName, volNameRet)
+		}
+	}
+
+	// get first 10
+	resp, err := c.ListVolumes(context.Background(), &csi.ListVolumesRequest{MaxEntries: 10})
```
dougm: You have a `ctx` var in scope that could be used in place of  `context.Background()` here and the ones below.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	}
+
+	// Setup a multi-DC environment with zones!
+	// Setup DC0
+	dc0, err := vclib.GetDatacenter(ctx, vsi.Conn, "DC0")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Attach a random tag to DC0
```
dougm: I think the "random" comment comes from the in-tree test, but isn't a random tag in these tests :)
https://github.com/kubernetes/kubernetes/blob/2f2945732b5821f9d51c6acdfb0e9fabe81bf3a4/pkg/cloudprovider/providers/vsphere/vsphere_test.go#L421-L422

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	if err = m.AttachTag(ctx, regionID, dc0.Reference()); err != nil {
+		t.Fatal(err)
+	}
+	if err = m.AttachTag(ctx, zoneIDwest, dc0.Reference()); err != nil {
```
dougm: No need to call `.Reference()` in any of the AttachTag calls, the param is a `mo.Reference` interface, which all the inventory types satisfy.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
...
+	reqCreate.AccessibilityRequirements.Requisite = append(reqCreate.AccessibilityRequirements.Requisite, topology)
+
+	respCreate, err := c.CreateVolume(ctx, reqCreate)
+	if err != nil {
+		t.Fatalf("CreateVolume failed: %v", err)
+	}
+
+	volID := respCreate.Volume.VolumeId
+	volName := respCreate.Volume.VolumeContext[AttributeFirstClassDiskName]
+	log.Println("ID:", volID)
```
dougm: `t.Logf` here too.

pkg/csi/service/fcd/controller_test.go
dvonthenen: Yup, will change

pkg/csi/service/fcd/controller_test.go
dvonthenen: Copy and past error. Will update.

pkg/csi/service/fcd/controller_test.go
dvonthenen: yup very true

pkg/csi/service/fcd/controller_test.go
dvonthenen: Yup

pkg/csi/service/fcd/controller_test.go
dvonthenen: Got it.

pkg/csi/service/fcd/controller_test.go
```diff
@@ -0,0 +1,644 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
codenrhoden: year should be 2019 since this is a brand new file.

pkg/csi/service/fcd/controller_test.go
dvonthenen: Will update

pkg/csi/service/fcd/controller_test.go
dvonthenen: Yup

dvonthenen:
Updates based on feedback

PR #149, opened 2019-02-27
Part 2: More go tests

dvonthenen:
Final PR for go testing. Tests added:
- Now testing the instances interface
- Now testing ConnectionManager... testing all the ways we look up or find things. Meaning we test list, search, and zone-based searches. New `_test.go` file for each category. These are used everywhere in both CCM and CSI.
- configFromEnvOrSim now supported multi-DC environment
- NodeManager now testing using multi-DC
- Improved the ListVolumes test for ordering
- Add ProviderID tests

Refactored:
- NodeManager and CCM Zones go tests don't need to initialize using newVSphere(). Was overkill and did a bunch of mostly unnecessary stuff.

NA

Since all changes were to `go tests` code only. Just made sure that `make test` passed as excepted.

**Release note**:
NA

### DIFF
--- a/pkg/common/connectionmanager/zones_test.go
+++ b/pkg/common/connectionmanager/zones_test.go
@@ -1,0 +1,310 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package connectionmanager
+
+import (
+	"context"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vapi/rest"
+	"github.com/vmware/govmomi/vapi/tags"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
+)
+
+func TestRemovePortFromHost(t *testing.T) {
+	str1 := removePortFromHost("127.0.0.1")
+	if !strings.EqualFold("127.0.0.1", str1) {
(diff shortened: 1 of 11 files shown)


dvonthenen:

pkg/common/connectionmanager/zones_test.go
```diff
@@ -0,0 +1,310 @@
...
+	"strings"
+	"testing"
+
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vapi/rest"
+	"github.com/vmware/govmomi/vapi/tags"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
+)
+
+func TestRemovePortFromHost(t *testing.T) {
```
codenrhoden: I'm not sure whether to ask for a change here, and I might be in the minority, but I don't like the way a lot of these tests are structured. Specifically, when a single function is being tested multiple times with different inputs, it's better to structure the tests where there is a slice of test parameters (capturing the permutations of inputs and expected outputs), then range over those tests. It makes adding new test parameters easier. These are called test tables

In this specific example, there should only be one place where `removePortFromHost()` is called. Setting up tests this way also allows one to start running the tests in parallel (`t.parallel()`) if the test supports it, but that's a separate issue.

Here's a real world example: https://github.com/kubernetes/kubernetes/blob/b2e9b2a842b72d3a19ad7e540d0589dd84722d2a/pkg/util/mount/mount_test.go#L24

And a couple of good blogs:
https://github.com/golang/go/wiki/TableDrivenTests
https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go

I'm not sure I want to derail this PR by bikeshedding on this, but wanted to bring it up.

dvonthenen: I actually prefer the individual tests because it makes it super obvious what failed. If I am in the minority, I can change it. Doesn't really matter to me as long as everything is getting tested.

dvonthenen: This would also apply to `util_test.go` since it has similar structure.

I would mention that we would not likely add any more since those are all of the boundary conditions.

codenrhoden: yeah, definitely not something I would insist on. =) Thanks for taking this on!

dougm: I'm a fan of table driven tests, but agree they can be trickier to implement/debug.  The new-ish `Run` method is supposed to help with that, providing more context, but I have yet to make use of it personally: https://golang.org/pkg/testing/#T.Run

PR #153, opened 2019-03-07
CSI ginkgo

codenrhoden:
This patch adds Ginkgo tests for the CSI Plugin that can be exercised
when the plugin has not been configured (meaning, no vCenter
interaction).

These tests start an actual CSI plugin and communicate with it over
gRPC. The connection is in-memory thanks to the memconn package,
which makes testing simpler and faster.

The reference to "-tags unit" has been removed from the Makefile's
"unit" target, as no Go source files had a unit build constraint in
them. This call was ineffective.

In a separate commit, the reference to glog is removed from Gopkg.toml. This is no longer a direct dependency, and does not belong in Gopkg.toml anymore.

`make test` passes

This is actually a first step on the way to E2E tests. This covers what we can test on a "real" CSI plugin without configuring it and deploying it on vSphere infrastructure. Next step will be a set of tests that can be run on any (Linux) node that has access to vCenter and perform the appropriate storage functionality on a single node. After that will come consuming the K8s testing framework to test functionality within K8s for a full E2E test.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -51,6 +51,14 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:e6f895360ed15ead971e11a68cbcff0d6bb4b7bb23b064404934acc6010516a5"
+  name = "github.com/akutz/memconn"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "e0a19f53d865bbc181e3ed1204978510b315f7e2"
+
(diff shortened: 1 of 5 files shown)


codenrhoden:
/test pull-cloud-provider-vsphere-test

test failures seem unrelated to my changes.... re-testing

PR #159, opened 2019-03-10
Make Env Vars take precedence over config file

codenrhoden:
The way that the config file was parsed and loaded previously was done
in a way that when reading a vSphere.conf file, env vars were checked
first, all errors from that were ignored, then values were loaded from
file and the config was validated.

Env vars should take precedence. This patch changes it so that when
reading a config file, that file is loaded, any valid env vars overwrite
the values from the file, and then the whole composite config is
validated.

It was also true (in the case of CSI) that a vsphere.conf file was
always required. This makes future testing difficult, as for basic
cases, it should be possible to configure the CSI plugin via Env vars
only. This patch makes that possible.

It was previously possible that if you did use Env Var only for config,
that an invalid config was returned without error. This patch fixes that
as well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -33,23 +33,22 @@ import (
 	"k8s.io/cloud-provider-vsphere/pkg/common/config"
 )
 
-func ParseConfig(configFile string) (config.Config, error) {
-	var cfg config.Config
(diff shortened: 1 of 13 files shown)


dvonthenen:
/lgtm

CC: @frapposelli

PR #164, opened 2019-03-12
Add env var to disable CSI K8s client

codenrhoden:
This PR will make it possible (may not be all that's needed) to run the [CSI-Sanity test suite](https://github.com/kubernetes-csi/csi-test) against our CSI plugin outside of a K8s pod.

CSI plugins are intended to be CO agnostic. The current CSI plugin was
always initializing a K8s client and informer, which would cause plugin
initialization to fail if it was not running within a K8s pod.

This patch adds a new env var, X_CSI_DISABLE_K8S_CLIENT which can be set to
"true" (or any other truthy value) to disable said initialization.

This patch also changes some methods around the K8s client and informer
to no longer use a pointer to a Go interface, as Go interfaces are
already pointers.

`pkg/csi/envvars.go` was moved to `pkg/csi/types/envvars.go` to avoid circular imports.

The struct `pkg/csi/service/fcd/controller.controller` no longs holds a reference to the Kubernetes clientset or the informer, as these were never used again outside of the `Init` function.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/vsphere-csi/main.go
+++ b/cmd/vsphere-csi/main.go
@@ -39,4 +39,15 @@ const usage = `    X_CSI_VSPHERE_APINAME
         Specifies the name of the API to use when talking to vCenter
 
 				The default value is "FCD" (First Class Disk)
+
+    X_CSI_VSPHERE_CLOUD_CONFIG
+        Specifies the path to the vsphere.conf file
+
+        The default falue is "/etc/cloud/vsphere.conf"
+
+    X_CSI_DISABLE_K8S_CLIENT
+        Boolean flag that disables the Kubernetes API client to retrieve
+        secrets.
+
+        The default value is "false"
 `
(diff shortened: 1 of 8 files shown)


frapposelli:
I encourage this change, but I **strongly** recommend to make the default `true` 😄

dvonthenen:
I believe the CSI driver already supports other COs. I had even tested this against DC/OS a while back.

codenrhoden:
Okay, I changed the default behavior and updated the PR description and release note to reflect that.

dvonthenen:
Yep, talked about it off line... i guess i broke that functionality when trying to refactor the CCM and CSI to use the same k8s connection creation mechanism.

codenrhoden:
@dvonthenen has pointed out that I could trigger the K8s init logic off of `VSPHERE_SECRETS_DIRECTORY` instead of add a new env var. I'm not sure if I like that better, as I would have not known to do that.

The current logic is still broken once the connection manager code was unified.

Thoughts?

should I make the env var more clear as to what it's doing -- like `X_CSI_DISABLE_K8S_CLIENT` or something? Because you definitely could set that, disable the client, but still have it work just fine on K8s.

dvonthenen:
Yeah, if you want to make it more clear what that field is for, go for it. Basically by setting that `VSPHERE_SECRETS_DIRECTORY` you are instructing the "secrets" to be mounted along with the container instead of getting them via the API.

codenrhoden:
@dvonthenen PTAL

PR #166, opened 2019-03-12
Enable linting on Prow

akutz:
This patch introduces support for linting in a pre-submit, blocking Prow job. The Makefile's `check` target now invokes `hack/check.sh`, which in turn invokes `make fmt`, `make vet`, and `make lint`, translating their results into the JUnit report that is consumable by the K8s test-grid.

This [gist](https://gist.github.com/akutz/42d1dc17d64edab19dbc1c11fc9fbaef) has an example of the `make check` command as well as the JUnit report the command produces.

The report is written to `${ARTIFACTS}/junit_check.xml`.

**Special notes for your reviewer**: NA

**Release note**: NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,6 +215,7 @@ endif # ifndef X_BUILD_DISABLED
 ################################################################################
 ##                                 TESTING                                    ##
 ################################################################################
+export ARTIFACTS ?= .
 PKGS_WITH_TESTS := $(sort $(shell find . -name "*_test.go" -type f -exec dirname \{\} \;))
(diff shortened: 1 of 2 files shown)


codenrhoden:
/lgtm
was going to suggest that this get put in place. nice work!

akutz:
Related to https://github.com/kubernetes/test-infra/pull/11736

dvonthenen:
@akutz was on hold... ok to merge as well?

akutz:
Hi @dvonthenen,

Please! Thank you for checking!

Issue #146, opened 2019-02-19
Provide documentation on how to run CCM and CSI with Kubeadm


frapposelli:
As `kubeadm` is the standard way to deploy Kubernetes on operating systems, we should provide instructions on how to deploy CCM and CSI when using it.

/kind feature

andrewsykim:
ref: https://github.com/kubernetes/cloud-provider/issues/9

mylesagray:
I signed it

frapposelli:
this is important for v1

frapposelli:
docs can be found here: https://cloud-provider-vsphere.sigs.k8s.io/tutorials/kubernetes-on-vsphere-with-kubeadm.html

Issue #41, opened 2018-08-17
Merge and improve the documentation of the cloud provider


embano1:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug
> /kind feature

/kind improvement

**What happened**:
Currently, the ownerhship and status of the in-tree cloud provider documentation is unclear. Also, it ties the cloud provider to the storage related Project "Hatchway" (https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/), where a separation of concerns is perhaps better.

Going forward with both, in-tree and out-of-tree cloud provider, we should consolidate, re-host and clean up the documentation (supported versions, etc.). Also related tools and scripts, e.g. VCP UX deployment script, should be validated against the latest versions or marked as outdated as the current version (https://github.com/vmware/kubernetes/tree/enable-vcp-uxi) seems to be broken with K8s 1.11.

There's several open issues regarding the documentation, e.g. https://github.com/vmware/kubernetes/issues/479 and https://github.com/vmware/kubernetes/issues/491. Also, the current documentation seems to be incorrect with regards to the vsphere.conf on the workers: https://github.com/vmware/kubernetes/issues/501

andrewsykim:
ref https://github.com/kubernetes/cloud-provider/issues/9

frapposelli:
@cantbewong can you help with some links to the docs for in-tree that you authored?

andrewsykim:
Some background information on this from SIG Cloud Provider:

* Ideally we want all providers following a set template so we can easily pull all the docs into a centralized page (something like https://kubernetes.io/docs/setup/pick-right-solution/) 
* There's a KEP to standardize on provider docs which is still a WIP but I think it would still be beneficial to start implementing it based on the proposed changes in https://github.com/kubernetes/enhancements/pull/827

andrewsykim:
/remove-lifecycle stale 

WIP in https://github.com/kubernetes/cloud-provider-vsphere/pull/181

frapposelli:
/close

fixed in #181

PR #181, opened 2019-04-10
vSphere Cloud Provider Documentation

andrewsykim:
Iterating on the vSphere cloud provider docs as per https://github.com/kubernetes/cloud-provider/issues/9, https://github.com/kubernetes/cloud-provider-vsphere/issues/41 and https://github.com/kubernetes/cloud-provider-vsphere/issues/146. The PR is currently just a template/outline and has no content. Once we agree on a general outline we can start working on the content.

https://github.com/kubernetes/cloud-provider/issues/9
https://github.com/kubernetes/cloud-provider-vsphere/issues/41

This doc covers both the in-tree and out-of-tree vSphere cloud provider and assumes the reader is relatively new to using Kubernetes. 

**Release note**:

```release-note
NONE
```


andrewsykim:
@cantbewong @embano1 @frapposelli @dougm @dvonthenen @mylesagray

docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+  - [kube-controller-manager](#kube-controller-manager)
+  - [cloud-controller-manager](#cloud-controller-manager)
+  - [kubeadm](#kubeadm)
+- [Architecture](#architecture)
+  - [Kubernetes using the In-Tree vSphere Provider](#kubernetes-using-the-in-tree-vsphere-provider)
+  - [Kubernetes using the Out-of-Tree vSphere Provider](#kubernetes-using-the-out-of-tree-vsphere-provider)
+- [vSphere Integrations](#vsphere-integrations)
+  - [Kubernetes Nodes](#kubernetes-nodes)
+  - [Kubernetes Zones/Regions Topology](#kubernetes-zones-regions-topology)
+  - [Kubernetes LoadBalancers](#kubernetes-loadbalancers)
```
andrewsykim: We don't support this yet but I think worth calling that out + future intentions

docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+  - [Kubnernetes Routes](#kubernetes-routes)
```
andrewsykim: We don't support this yet but I think worth calling that out + future intentions

docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+- [Example Configurations and Manifests](#example-manifests)
+- [Tutorials](#tutorials)
+- [Addons](#addons)
+  - [Storage](#storage)
```
codenrhoden: I like the callouts for the addons. Let's just make sure that when we do fill this with content, we make it clear that the out-of-tree provider *requires* the CSI plugin.

andrewsykim: Good point, how about renaming this to `Required Plugins` and linking out to CSI docs?

codenrhoden: Sounds good -- we may want both categories: optional and required?

andrewsykim: How about `Required` and `Recommended`? I feel too many things fall into optional plugins 😆

andrewsykim:
@mylesg filled out the first few sections, left VMware specific component definitions for you to fill out

andrewsykim:
Oops, meant to ping @mylesagray sorry

PR #184, opened 2019-05-07
use string data for example vSphere secret manifest

andrewsykim:

Use stringData for secret manifest, easier to install since you don't have to take the extra step of base64 encoding your secret values.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/manifests/controller-manager/vccm-secret.yaml
+++ b/manifests/controller-manager/vccm-secret.yaml
@@ -3,6 +3,6 @@ kind: Secret
 metadata:
   name: vccm
   namespace: kube-system
-data:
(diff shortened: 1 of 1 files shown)


andrewsykim:
oops typo

/assign @dvonthenen

frapposelli:
Holding this PR until we define if it's going to make the 0.2.0 cut
/hold

dvonthenen:
@frapposelli yea, plus we didn't make the change on the CSI side for the 0.2.0 release. if we do make this change on the CCM side, the instructions will be different between the two and might be confusing. Currently, they are identical.

PR #189, opened 2019-06-04
Fix logging for vendored packages

dvonthenen:
This does effectively the same thing that the vSphere CSI plugin already does for logging when klog and glog are intermixed. This is described on the klog github page.

None

**Release note**:
No behavioral changes


dvonthenen:
CC: @frapposelli @dougm

PR #195, opened 2019-06-11
add order-only pre-req for deploy, as it's beeing skipped when recursing

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

**What this PR does / why we need it**: this explicits the order-only pre-requirement as it is being skipped while recursing using `$(MAKE)`

**Which issue this PR fixes** : fixes #

/assign @akutz 
**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ dist: dist-ccm
 ################################################################################
 # The deploy target is for use by Prow.
 .PHONY: deploy
-deploy:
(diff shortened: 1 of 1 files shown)


codenrhoden:
/lgtm

Once we see the deploys from Prow working again, we'll want to make the same update(s) in the CSI repo. So glad to see this tracked down!

Issue #202, opened 2019-06-17
CCM failing to find node in VC 


EleanorRigby:
**Is this a BUG REPORT or FEATURE REQUEST?**: BUG

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
On deploying the CCM as a pod, it gives CrashLoopBackoff. The container fails to start.
from the logs, I believe CCM failed to find a node because the node name is different from a vSphere DNS name.
PKS uses a `--host-override`  when starting kubelet and changes the default hostname that CCM uses to `find_node_by_dns_name` 

**Possible Solution**:
To use some sort of best-effort to

 - find vm by IP
rescue NOTFOUND
 - find vm by name
rescue NOTFOUND
 - find vm by uuid
rescue NOTFOUND
 - find vm by label/xyz etc...
and fail if not found with any means.

**What you expected to happen**:
CCM to start running.

**How to reproduce it (as minimally and precisely as possible)**:
Deploy a PKS (k8s) cluster where hostname for worker nodes is different from the DNS name on vSphere for those worker nodes(vm)
Repeat the ccm deploy procedure in tutotrial

**LOGS**:
Logs for failed ccm
```
worker/fc691ecf-5050-450d-b347-ad880f237ccd:/var/log/containers# cat vsphere-cloud-controller-manager_kube-system_vsphere-cloud-controller-manager-9a5afd96fc9a2696a5654e10db72bfa16b67fa81cc24f7cb5e1799ead31f0ed7.log
{"log":"I0617 22:13:04.855020       1 flags.go:27] FLAG: --address=\"0.0.0.0\"\n","stream":"stderr","time":"2019-06-17T22:13:04.857014104Z"}
{"log":"I0617 22:13:04.855120       1 flags.go:27] FLAG: --allocate-node-cidrs=\"false\"\n","stream":"stderr","time":"2019-06-17T22:13:04.857100641Z"}
{"log":"I0617 22:13:04.855128       1 flags.go:27] FLAG: --allow-untagged-cloud=\"false\"\n","stream":"stderr","time":"2019-06-17T22:13:04.857105678Z"}
{"log":"I0617 22:13:04.855132       1 flags.go:27] FLAG: --alsologtostderr=\"true\"\n","stream":"stderr","time":"2019-06-17T22:13:04.85710867Z"}
[...]

PR #203, opened 2019-06-18
Implement Lookup VM by IP

dvonthenen:
This addresses the issue reported below. It allows for the capability for the VMName to be overridden from the FQDN/hostname to the IP address of the node. Implements FindVMByIP.

This also addresses a bug a crash issue when the ProviderID cannot be found or when the ProviderID is malformed.

**Which issue this PR fixes**: https://github.com/kubernetes/cloud-provider-vsphere/issues/202

Tested on 1.14.1 with `--hostname-override=<ip address>` on the node.
Using vSphere 6.7u2 (but that version shouldn't matter).

**Release note**:
NA

### DIFF
--- a/pkg/cloudprovider/vsphere/util.go
+++ b/pkg/cloudprovider/vsphere/util.go
@@ -19,12 +19,17 @@ package vsphere
 import (
 	"fmt"
 	"strings"
+
+	"k8s.io/klog"
 )
 
 const (
 	// ProviderPrefix is the Kubernetes cloud provider prefix for this
 	// cloud provider.
 	ProviderPrefix = "vsphere://"
+
+	// MinUUIDLen is the min length for a valid UUID
+	MinUUIDLen int = 36
 )
 
 // GetUUIDFromProviderID returns a UUID from the supplied cloud provider ID.
@@ -40,6 +45,10 @@ func GetUUIDFromProviderID(providerID string) string {
 // K8s:    56492e42-22ad-3911-6d72-59cc8f26bc90
 // VMware: 422e4956-ad22-1139-6d72-59cc8f26bc90
 func ConvertK8sUUIDtoNormal(k8sUUID string) string {
+	if len(k8sUUID) < MinUUIDLen {
+		klog.Errorf("The UUID length is invalid. Returning UUID=%s as is.", k8sUUID)
+		return k8sUUID
+	}
 	uuid := fmt.Sprintf("%s%s%s%s-%s%s-%s%s-%s-%s",
 		k8sUUID[6:8], k8sUUID[4:6], k8sUUID[2:4], k8sUUID[0:2],
 		k8sUUID[11:13], k8sUUID[9:11],
--- a/pkg/common/connectionmanager/search.go
+++ b/pkg/common/connectionmanager/search.go
@@ -35,6 +35,8 @@ func (f FindVM) String() string {
 		return "byUUID"
 	case FindVMByName:
 		return "byName"
+	case FindVMByIP:
+		return "byIP"
 	default:
 		return "byUnknown"
 	}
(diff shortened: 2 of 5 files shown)


pkg/cloudprovider/vsphere/util.go
```diff
@@ -40,6 +44,10 @@ func GetUUIDFromProviderID(providerID string) string {
 // K8s:    56492e42-22ad-3911-6d72-59cc8f26bc90
 // VMware: 422e4956-ad22-1139-6d72-59cc8f26bc90
 func ConvertK8sUUIDtoNormal(k8sUUID string) string {
+	if len(k8sUUID) < 36 {
```
frapposelli: this was probably meant to reference `MinUUIDLen`

dvonthenen: It was... I even made the const for it... that's what happens when you get sidetracked. 😀

pkg/common/connectionmanager/search.go
```diff
@@ -63,6 +63,8 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
 	if searchBy == FindVMByUUID {
 		klog.V(3).Info("WhichVCandDCByNodeID by UUID")
 		myNodeID = strings.ToLower(nodeID)
+	} else if searchBy == FindVMByIP {
```
frapposelli: this is a nit more than anything, but I really dislike `else if` statements in go 😄, can you make this a `switch`?

pkg/common/connectionmanager/search.go
```diff
@@ -163,6 +165,8 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
 				var err error
 				if searchBy == FindVMByUUID {
 					vm, err = res.datacenter.GetVMByUUID(ctx, myNodeID)
+				} else if searchBy == FindVMByIP {
```
frapposelli: same as above 🙂

pkg/common/connectionmanager/search.go
dvonthenen: Yup

pkg/common/connectionmanager/search.go
dvonthenen: Will fix

pkg/common/connectionmanager/search.go
```diff
@@ -187,12 +191,18 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
 					continue
 				}
 
+				hostName := oVM.Guest.HostName
+				if searchBy == FindVMByIP {
+					klog.V(2).Info("WhichVCandDCByNodeID by IP. Overriding VMName from=%s to to=%s", oVM.Guest.HostName, myNodeID)
```
frapposelli: this needs to be `.Infof` or it will not pass lint

pkg/cloudprovider/vsphere/util.go
```diff
@@ -19,12 +19,16 @@ package vsphere
...
+
+	"k8s.io/klog"
 )
 
 const (
 	// ProviderPrefix is the Kubernetes cloud provider prefix for this
 	// cloud provider.
 	ProviderPrefix = "vsphere://"
+
+	MinUUIDLen int = 36
```
frapposelli: this also needs a comment

dvonthenen: Adding comment

pkg/common/connectionmanager/search.go
dvonthenen: Good catch!

pkg/common/vclib/datacenter.go
```diff
@@ -77,6 +77,23 @@ func GetNumberOfDatacenters(ctx context.Context, connection *VSphereConnection)
...
+	svm, err := s.FindByIp(ctx, dc.Datacenter, ipAddy, true)
+	if err != nil {
+		klog.Errorf("Failed to find VM by IP. VM IP: %s, err: %+v", ipAddy, err)
+		return nil, err
+	}
+	if svm == nil {
+		klog.Errorf("Unable to find VM by IP. VM IP: %s", ipAddy)
+		return nil, ErrNoVMFound
+	}
+	virtualMachine := VirtualMachine{object.NewVirtualMachine(dc.Client(), svm.Reference()), dc}
```
dougm: `FindByIp` returns interface type `Reference`, that'll be a `*object.VirtualMachine` so you can avoid `NewVirtualMachine`:
```go
virtualMachine := VirtualMachine{svm.(*object.VirtualMachine), dc}
```

Same goes for the other cases where CCM uses SearchIndex.FindBy{Uuid,DnsName}.. not a huge resource savings, but would make the code slightly more concise.  Up to you on making that change here or in a separate PR, or not at all.

dvonthenen: Updated!

dvonthenen:
Updated with latest comments

Issue #204, opened 2019-06-20
CCM failing with Panic


EleanorRigby:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:
CCM failed to come up with `CrashLoopBackOff`

**What you expected to happen**:
CCM should have not crashed.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy a PKS testbed.
2. Configure external cloud provider in kube-apiserver, kubelet, kube-controller-manager
3. Deploy CCM

**Anything else we need to know?**:
Logs
```
worker/db7189f1-1661-4fec-b03b-4fb330c1c38d:/var/log/containers# cat vsphere-cloud-controller-manager-6d45h_kube-system_vsphere-cloud-controller-manager-15de1908fa3435b0991f0541f988c1b83bfc8e8cac91a1cbaa9274144c32ea80.log
{"log":"I0620 03:03:23.063799       1 flags.go:27] FLAG: --address=\"0.0.0.0\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064195259Z"}
{"log":"I0620 03:03:23.064114       1 flags.go:27] FLAG: --allocate-node-cidrs=\"false\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064351562Z"}
{"log":"I0620 03:03:23.064138       1 flags.go:27] FLAG: --allow-untagged-cloud=\"false\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064357099Z"}
{"log":"I0620 03:03:23.064152       1 flags.go:27] FLAG: --alsologtostderr=\"true\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064360387Z"}
{"log":"I0620 03:03:23.064214       1 flags.go:27] FLAG: --bind-address=\"0.0.0.0\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064403632Z"}
{"log":"I0620 03:03:23.064244       1 flags.go:27] FLAG: --cert-dir=\"/var/run/kubernetes\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064408664Z"}
{"log":"I0620 03:03:23.064264       1 flags.go:27] FLAG: --cidr-allocator-type=\"RangeAllocator\"\n","stream":"stderr","time":"2019-06-20T03:03:23.064412012Z"}
[...]

PR #205, opened 2019-06-20
Fix UUID/ProviderID to only use lower case based on VC and k8s matching/lookup

dvonthenen:
This fixes an issue where a user-defined vSphere ProviderID `--provider-id=` is of mixed case (lower/upper). The internal workings of the cloud provider interface discovers the ProviderID in lower case and the internal lookup fails if with at least 1 UpperCase character when user-provider.

Additional changes:
- Added negative go tests for cloud provider Instance interface
- Purposely made provided UUID ToUpper in certain go tests to validate this fix
- Fixed Infof -> Info in certain locations

**Which issue this PR fixes**: https://github.com/kubernetes/cloud-provider-vsphere/issues/204

Tested on vSphere 6.7u2

**Release note**:
Does not affect user documentation or behavior

### DIFF
--- a/pkg/cloudprovider/vsphere/instances_test.go
+++ b/pkg/cloudprovider/vsphere/instances_test.go
@@ -22,7 +22,7 @@ import (
 	"testing"
 
 	"github.com/vmware/govmomi/simulator"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	clientv1 "k8s.io/client-go/listers/core/v1"
@@ -80,7 +80,7 @@ func TestInstance(t *testing.T) {
 	vm := simulator.Map.Any("VirtualMachine").(*simulator.VirtualMachine)
 	name := vm.Name
 	vm.Guest.HostName = name
-	UUID := vm.Config.Uuid
+	UUID := strings.ToUpper(vm.Config.Uuid)
 	k8sUUID := ConvertK8sUUIDtoNormal(UUID)
 
 	node := &v1.Node{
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/vsphere/instances_test.go
```diff
@@ -22,7 +22,7 @@ import (
 	"testing"
 
 	"github.com/vmware/govmomi/simulator"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
yastij: nit: is this change needed ?

pkg/cloudprovider/vsphere/util.go
```diff
@@ -34,7 +34,7 @@ const (
 
 // GetUUIDFromProviderID returns a UUID from the supplied cloud provider ID.
 func GetUUIDFromProviderID(providerID string) string {
-	return strings.TrimPrefix(providerID, ProviderPrefix)
+	return strings.ToLower(strings.TrimSpace(strings.TrimPrefix(providerID, ProviderPrefix)))
```
yastij: nit: this is becoming a bit long

pkg/common/connectionmanager/search.go
```diff
@@ -202,12 +202,14 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
 					hostName = myNodeID
 				}
 
+				UUID := strings.ToLower(strings.TrimSpace(oVM.Summary.Config.Uuid))
```
yastij: nit: lowercase `UUID`

pkg/common/connectionmanager/search.go
```diff
@@ -202,12 +202,14 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
...
+
 				klog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
 					nodeID, vm, res.vc, res.datacenter.Name())
-				klog.V(2).Info("Hostname: ", hostName, " UUID: ", oVM.Summary.Config.Uuid)
+				klog.V(2).Info("Hostname: ", hostName, " UUID: ", UUID)
```
yastij: same

pkg/common/connectionmanager/search.go
```diff
@@ -202,12 +202,14 @@ func (cm *ConnectionManager) WhichVCandDCByNodeID(ctx context.Context, nodeID st
...
 
 				vmInfo = &VMDiscoveryInfo{DataCenter: res.datacenter, VM: vm, VcServer: res.vc,
-					UUID: oVM.Summary.Config.Uuid, NodeName: hostName}
+					UUID: UUID, NodeName: hostName}
```
yastij: same

pkg/common/connectionmanager/search.go
dvonthenen: go lint complains if UUID isn't in caps

pkg/common/connectionmanager/search.go
dvonthenen: Same comment as above

pkg/common/connectionmanager/search.go
dvonthenen: Same as above

pkg/cloudprovider/vsphere/instances_test.go
dvonthenen: The go lint seems to have added that in to remove ambiguity. Each time I try to save the file without it, it adds it back in.

pkg/cloudprovider/vsphere/util.go
dvonthenen: Broke up the nesting of calls a little.

dvonthenen:
Updated based on feedback from @yastij

dvonthenen:
CC: @fabio @dougm for another look

pkg/common/connectionmanager/search.go
frapposelli: Just a nit, I like using printf style formatting instead of appending variables, it would make it consistent with the rest of the code.

dvonthenen: Yup. Updated for uniformity.

dvonthenen:
Updated based on @frapposelli feedback

Issue #207, opened 2019-06-25
InstanceShutdownByProviderID is not implemented 


divyenpatel:
**Is this a BUG REPORT or FEATURE REQUEST?**:
> /kind feature

**What happened**:
Cloud Provider vSphere does not implement `InstanceShutdownByProviderID `. This is very important for determining if Node VM is shutdown, and so volumes attached to node VM are safe to detach.

- https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/instances.go#L172
- https://github.com/kubernetes/cloud-provider/blob/master/cloud.go#L169

**What you expected to happen**:
Need implementation of this interface.  

**Anything else we need to know?**:
In-tree vSphere Cloud Provider implements this interface - https://github.com/kubernetes/kubernetes/blob/release-1.15/staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go#L737

**Environment**:
- vsphere-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

PR #208, opened 2019-06-25
Add InstanceShutdownByProviderID implementation

frapposelli:
This PR adds a quick and naive implementation of InstanceShutdownByProviderID, used to track when a node is shutdown and safe to detach storage from

Fixes: #207

/cc: @dvonthenen @andrewsykim

### DIFF
--- a/pkg/cloudprovider/vsphere/instances.go
+++ b/pkg/cloudprovider/vsphere/instances.go
@@ -171,5 +171,22 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 // InstanceShutdownByProviderID returns true if the instance is in safe state to detach volumes
 func (i *instances) InstanceShutdownByProviderID(ctx context.Context, providerID string) (bool, error) {
 	klog.V(4).Info("instances.InstanceShutdownByProviderID() called")
-	return false, cloudprovider.NotImplemented
+
+	// Check if node has been discovered already
+	uid := GetUUIDFromProviderID(providerID)
+	if _, ok := i.nodeManager.nodeUUIDMap[uid]; !ok {
+		// IF the uuid is not cached, we end up here
+		klog.V(2).Info("instances.InstanceShutdownByProviderID() NOT CACHED")
+		if err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID); err != nil {
+			klog.V(4).Info("instances.InstanceShutdownByProviderID() NOT FOUND with ", uid)
+			// if we can't discover, return false with an error in tow
+			return false, err
+		}
+		klog.V(2).Infof("instances.InstanceShutdownByProviderID() EXISTS with %q", uid)
+	}
+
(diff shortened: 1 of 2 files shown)


frapposelli:
Keeping it WIP as it hasn't been tested on a real environment yet

dvonthenen:
/lgtm

Will test this out tomorrow morning.

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -171,5 +171,27 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
...
+		klog.V(2).Info("instances.InstanceShutdownByProviderID() NOT CACHED")
+		if err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID); err != nil {
+			klog.V(4).Info("instances.InstanceShutdownByProviderID() NOT FOUND with ", uid)
+			// if we can't discover, return false with an error in tow
+			return false, err
+		}
+		klog.V(2).Infof("instances.InstanceShutdownByProviderID() EXISTS with %q", uid)
+	}
+
+	vm, err := i.nodeManager.shakeOutNodeIDLookup(ctx, uid, cm.FindVMByUUID)
```
dvonthenen: We can remove the call to shakeOutNodeIDLookup() because `nodeUUIDMap[uid]` contains `vm            *vclib.VirtualMachine` and that call is fairly expensive.

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -171,5 +171,27 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
...
+	if err != nil {
+		klog.V(2).Infof("instances.InstanceShutdownByProviderID() shakeOutNodeIDLookup failed with %q", err)
+		return false, err
+	}
+	active, err := vm.VM.IsActive(ctx)
```
dvonthenen: Just after the call to `IsActive()` can we add a log statement like below in order to observe the result:
```
klog.Infof("VM=%s IsActive=%t", uid, active)
```

pkg/cloudprovider/vsphere/instances.go
BaluDontu: Agree with @dvonthenen . Please use nodeUUIDMap[uid] to retrieve the vm instance.

pkg/cloudprovider/vsphere/instances.go
frapposelli: done 👍

pkg/cloudprovider/vsphere/instances.go
frapposelli: done, I figured there was a lookup table for it, but didn't occur to me it was this obvious 🙂

frapposelli:
@dvonthenen please test it again 👍

dvonthenen:
@frapposelli will be testing the updates this morning

PR #211, opened 2019-07-02
Fixed log lines not appearing when using klog.V(X) and glog.

dvonthenen:
Addresses yet another issue when using klog and sub or vendored components using glog. This was affecting when klog.V(X) log entries would not appear. As in the example below, the ORG line would not print in the logs, but the NEW line would.

```
klog.V(2).Infof("ORG VM=%s IsActive=%t", uid, active)
klog.Infof("NEW VM=%s IsActive=%t", uid, active)
```

Verified fix on 1.114.1 and 1.13.7.

NA

Used openstack CCM as a basis for comparison.

**Release note**:
Does not affect user documentation or user interaction

### DIFF
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -65,6 +65,27 @@ func main() {
 		Use: "cloud-controller-manager",
 		Long: `The Cloud controller manager is a daemon that embeds
 the cloud specific control loops shipped with Kubernetes.`,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			// Glog requires this otherwise it complains.
+			goflag.CommandLine.Parse(nil)
+
+			// TODO: We need to revisit this when we vendor newer components of k8s
+			// TODO: such as but not limited to k/k, client, apimachinery, etc
+			//
+			// This is a temporary hack to enable proper logging until upstream dependencies
+			// are migrated to fully utilize klog instead of glog.
+			klogFlags := goflag.NewFlagSet("klog", goflag.ExitOnError)
+			klog.InitFlags(klogFlags)
+
+			// Sync the glog and klog flags.
+			cmd.Flags().VisitAll(func(f1 *pflag.Flag) {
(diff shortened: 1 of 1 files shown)


dvonthenen:
CC @frapposelli

frapposelli:
sneaky little 🐛 

/lgtm
/approve

holding for @andrewsykim PTA:
/hold

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -65,6 +65,24 @@ func main() {
 		Use: "cloud-controller-manager",
 		Long: `The Cloud controller manager is a daemon that embeds
 the cloud specific control loops shipped with Kubernetes.`,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			// Glog requires this otherwise it complains.
+			goflag.CommandLine.Parse(nil)
+
+			// This is a temporary hack to enable proper logging until upstream dependencies
+			// are migrated to fully utilize klog instead of glog.
+			klogFlags := goflag.NewFlagSet("klog", goflag.ExitOnError)
```
andrewsykim: Is it just k8s.io/kubernetes using glog at the moment? If/when we update k/k vendor to use a version that is using klog can we remove this?

dvonthenen: Definitely k/k, but not really sure what else. We can definitely take a look at it when we vendor in a new version.

dvonthenen: Resolve based on the discussion in slack:

> andrewsykim: There are some optimizations and bug fixes that went in from v1.11 so we should have a discussion about what version to go to before GA though, but yeah feel free to hold cancel that one, maybe add a TODO comment

dvonthenen:
Updated based on feedback

andrewsykim:
/lgtm 

We should revisit the TODO comments after https://github.com/kubernetes/cloud-provider-vsphere/issues/212

dvonthenen:
> /lgtm
> 
> We should revisit the TODO comments after #212

Totally agree

PR #216, opened 2019-07-16
Fix repo owners file

frapposelli:
This PR fixes the repo OWNERS file, aligning with the current standard (OWNERS_ALIASES and OWNERS) used by other projects, it also removes @kerneltime and @luomiao that are no longer involved with the project

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,17 @@
+# See the OWNERS docs: https://git.k8s.io/community/docs/devel/owners.md
+
+aliases:
+  sig-cloud-provider-leads:
+    - andrewsykim
+    - hogepodge
+    - jagosan
+  cloud-provider-vsphere-maintainers:
+    - abrarshivani
+    - andrewsykim
+    - baludontu
+    - divyenpatel
+    - dougm
+    - frapposelli
+    - imkin
+    - sandeeppissay
+    - dvonthenen
--- a/OWNERS
+++ b/OWNERS
@@ -1,10 +1,8 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
 approvers:
-- abrarshivani
-- baludontu
-- divyenpatel
-- imkin
-- kerneltime
-- luomiao
-- frapposelli
-- dougm
-- dvonthenen
+  - sig-cloud-provider-leads
+  - cloud-provider-vsphere-maintainers
+
+reviewers:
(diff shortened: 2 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -0,0 +1,17 @@
+# See the OWNERS docs: https://git.k8s.io/community/docs/devel/owners.md
+
+aliases:
+  sig-cloud-provider-leads:
```
codenrhoden: So this is something that has been bothering me for a long time across several projects... Why is that alias defined here at all? Shouldn't it be coming from kubernetes/org, aka [here](https://github.com/kubernetes/org/blob/master/OWNERS_ALIASES#L38).

I believe the Prow plugin walks all these redefinitions, but I think for top-level teams, the canonical source should be the `org` repo, otherwise this is a maintenance nightmare.

codenrhoden: I personally think the only use of `OWNER_ALIASES` for our projects would be define a group of our choosing for reviewers vs approvers, but at that point, we should just list them directly in `OWNERS`. Either way you have to modify a file. I see no use for `OWNER_ALIASES`. The `cloud-provider-leads` is a great example -- the list of people there are about to change, but this file here would override the top-level one, giving someone approval rights to our repo that is no longer a co-chair of the SIG.

/ rant

frapposelli: fair point, I wasn't aware that prow could walk the definitions back to `kubernetes/org`, I'll amend the file 👍 

I still see the value in keeping the `OWNERS_ALIASES` file as it defines who is part of what "local" group while `OWNERS` specify the groups that are able to approve and/or review.

codenrhoden: Cool. To be clear, I *think* that's the case, but haven't definitely proven it. I've looked through the Prow plugin code around this before, and think that's happening, but I wanted to start the conversation. I should probably just ask the test-infra folks sometime.

nikhita: :wave: Just want to confirm that prow will _not_ walk back definitions to k/org or any other repo. There is an open issue for this: https://github.com/kubernetes/test-infra/issues/10065, but it's not possible today. Members for all aliases need to be explicitly defined in the OWNERS_ALIASES file.

frapposelli: Ok, added it back 😄, thanks @nikhita and @codenrhoden for helping in getting to the bottom of this, it was indeed confusing 😅

PR #222, opened 2019-07-30
Add script for running shellcheck on scripts

codenrhoden:
Adds shellcheck linting.

Also adds a 'shellcheck' makefile target, and corrects a shellcheck
error with the gitbooks build script

This will be added to Prow as a pre-submit separately.

**Release note**:

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ IMAGE_CCM := $(REGISTRY)/vsphere-cloud-controller-manager
 
 version:
 	@echo $(VERSION)
-	
+
 # Printing the image versions are defined early so Go modules aren't forced.
 print-ccm-image:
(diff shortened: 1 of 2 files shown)


hack/build-gitbooks.sh
```diff
@@ -18,9 +18,9 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-export KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
```
codenrhoden: This was never used anywhere, so there was no need to export it. Rather than clean it up to pass the linter (which it does not), I just used the same `cd` command we've been using across all our other projects.

dvonthenen: Sounds reasonable to me

PR #226, opened 2019-08-02
Introduce new release tooling and structure

codenrhoden:
This patch adds a hack/release.sh script that can be used both locally
and in CI for generating and pushing images and binaries to GCR and GCS.
The Makefile is modified to make use of this script, and the integration
test is modified to use the new images.

The location of the images and binaries moves with this change, mainly
splitting the latest builds from master into one channel versus a
release channel that only contains tagged releases. Doc updates and
manifest changes need to come in after this as well.

The CPI container image is also now based on distroless.

Okay, this is  a big one. This change would bring CPI in-line with both CAPV and CSI, as far as how releases are done. Major changes to note...

Tagged images would now now live at:
`gcr.io/cloud-provider-vsphere/cpi/release`. So, for example, if we release v0.3.1, the image would be at `gcr.io/cloud-provider-vsphere/cpi/release/manager:v0.3.1`.

Images that are built from master after each merge will live at `gcr.io/cloud-provider-vsphere/cpi/ci`, and those built for testing during PRs would live at `gcr.io/cloud-provider-vsphere/cpi/pr`. This allows us to clean up old images (like for PRs) w/o worrying about if someone else is using them. The PR images should never be used other than for the PR itself.

This should also fix the error that's happening right now when pushing the binaries. Since we added the step to push the CCM binary to GCS using `gsutil`, the push has [been failing](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-cloud-provider-vsphere-deploy/1154396491809296384) with errors like:
```
gsutil cp .build/bin/vsphere-cloud-controller-manager.linux_amd64 gs://artifacts.cloud-provider-vsphere.appspot.com/binaries/675f0b47/
Copying file://.build/bin/vsphere-cloud-controller-manager.linux_amd64 [Content-Type=application/octet-stream]...
/ [0 files][    0.0 B/ 52.3 MiB]                                                
ResumableUploadAbortException: 403 Insufficient Permission
```
This is because no `gcloud auth` command was done. We logged into gcr via Docker, but not GCS. I probably need to create the GCS buckets still and apply the appropriate lifecycle to them, just like we did with CAPV.

Once this is merged in, we can added a new post-submit to Prow that pushes tagged releases for us automatically. The existing post-submit that runs post merge to master should still work, as it uses the same make target: `make deploy`.

/assign @frapposelli 
/assign @yastij 
/assign @dvonthenen 

/hold

I'm putting the hold on here because even if this looks good to others, I really want the buy-in from Fabio that these changes are okay.

Also, I expect it's going to take me a couple iterations here to get the integration (and maybe conformance?) tests to pass.

**Release note**:

```release-note
```

### DIFF
--- a/cluster/images/controller-manager/Dockerfile
+++ b/cluster/images/controller-manager/Dockerfile
@@ -10,8 +10,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM photon:2.0
+################################################################################
+##                               BUILD ARGS                                   ##
+################################################################################
+# This build arg allows the specification of a custom Golang image.
+ARG GOLANG_IMAGE=golang:1.12.6
 
-ADD vsphere-cloud-controller-manager /bin/
+# The distroless image on which the CPI manager image is built.
+#
+# Please do not use "latest". Explicit tags should be used to provide
+# deterministic builds. This image doesn't have semantic version tags, but
+# the fully-qualified image can be obtained by entering
+# "gcr.io/distroless/static:latest" in a browser and then copying the
+# fully-qualified image from the web page.
+ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:9b60270ec0991bc4f14bda475e8cae75594d8197d0ae58576ace84694aa75d7a
 
-CMD ["/bin/vsphere-cloud-controller-manager"]
+################################################################################
+##                              BUILD STAGE                                   ##
+################################################################################
+# Build the manager as a statically compiled binary so it has no dependencies
+# libc, muscl, etc.
+FROM ${GOLANG_IMAGE} as builder
+
+# This build arg is the version to embed in the CPI binary
+ARG VERSION=unknown
+
+# This build arg controls the GOPROXY setting
+ARG GOPROXY
+
+WORKDIR /build
+COPY go.mod go.sum ./
+COPY pkg/    pkg/
+COPY cmd/    cmd/
+ENV CGO_ENABLED=0
+ENV GOPROXY ${GOPROXY:-}
+RUN go build -a -ldflags='-w -s -extldflags=static -X main.version=${VERSION}' -o vsphere-cloud-controller-manager ./cmd/vsphere-cloud-controller-manager
+
+################################################################################
+##                               MAIN STAGE                                   ##
+################################################################################
+# Copy the manager into the distroless image.
+FROM ${DISTROLESS_IMAGE}
+LABEL maintainer="Travis Rhoden <trhoden@vmware.com>"
+COPY --from=builder /build/vsphere-cloud-controller-manager /bin/vsphere-cloud-controller-manager
+ENTRYPOINT [ "/bin/vsphere-cloud-controller-manager" ]
--- a/manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
+++ b/manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
@@ -36,9 +36,8 @@ spec:
       serviceAccountName: cloud-controller-manager
       containers:
         - name: vsphere-cloud-controller-manager
-          image: gcr.io/cloud-provider-vsphere/vsphere-cloud-controller-manager:latest
+          image: gcr.io/cloud-provider-vsphere/cpi/release/manager:latest
           args:
-            - /bin/vsphere-cloud-controller-manager
             - --v=2
             - --cloud-provider=vsphere
             - --cloud-config=/etc/cloud/vsphere.conf
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -7,7 +7,7 @@ PROJECT_ROOT ?= $(abspath ../..)
 
 CLUSTER_NAME := ccm-integration-test
 VERSION ?= $(shell git describe --always --dirty)
-CCM_IMAGE := gcr.io/cloud-provider-vsphere/vsphere-cloud-controller-manager:$(VERSION)
+CCM_IMAGE := gcr.io/cloud-provider-vsphere/cpi/pr/manager:$(VERSION)
 
 KIND_CONFIG := kind-config.yaml
 .PHONY: $(KIND_CONFIG)
@@ -53,7 +53,7 @@ deploy-ccm:
 	kubectl -n kube-system create -f secrets.yaml && \
 	kubectl -n kube-system apply -f ../../manifests/controller-manager/cloud-controller-manager-roles.yaml && \
 	kubectl -n kube-system apply -f ../../manifests/controller-manager/cloud-controller-manager-role-bindings.yaml && \
-	sed 's~gcr.io/cloud-provider-vsphere/vsphere-cloud-controller-manager:latest~$(CCM_IMAGE)~g' <../../manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml | kubectl -n kube-system apply -f -
+	sed 's~gcr.io/cloud-provider-vsphere/cpi/release/manager:latest~$(CCM_IMAGE)~g' <../../manifests/controller-manager/vsphere-cloud-controller-manager-pod.yaml | kubectl -n kube-system apply -f -
 
 delete-ccm:
 	kubectl -n kube-system delete configmaps cloud-config && \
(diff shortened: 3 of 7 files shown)


cluster/images/controller-manager/Dockerfile
```diff
@@ -10,8 +10,42 @@
...
 
-CMD ["/bin/vsphere-cloud-controller-manager"]
+################################################################################
+##                               MAIN STAGE                                   ##
+################################################################################
+# Copy the manager into the distroless image.
+FROM ${DISTROLESS_IMAGE}
+LABEL maintainer="Travis Rhoden <trhoden@vmware.com>"
+COPY --from=builder /build/vsphere-cloud-controller-manager /vsphere-cloud-controller-manager
+ENTRYPOINT [ "/vsphere-cloud-controller-manager" ]
```
frapposelli: This change will have an impact on the manifests, they currently pass the whole path of the binary as `/bin/vsphere-cloud-controller-manager` followed by arguments.

I'm fine with moving this to `ENTRYPOINT` (it just makes sense) but we need to amend the documentation/manifests at the same time.

test/integration/Makefile
```diff
@@ -7,7 +7,7 @@ PROJECT_ROOT ?= $(abspath ../..)
 
 CLUSTER_NAME := ccm-integration-test
 VERSION ?= $(shell git describe --always --dirty)
-CCM_IMAGE := gcr.io/cloud-provider-vsphere/vsphere-cloud-controller-manager:$(VERSION)
+CCM_IMAGE := gcr.io/cloud-provider-vsphere/cpi/pr/manager:$(VERSION)
```
frapposelli: same comment here for the manifests, fine with changing the url but needs to be reflected in the manifests

cluster/images/controller-manager/Dockerfile
codenrhoden: Good catch on this one, that makes sense. In fact, I think we should talk about what we want... CSI actually uses `CMD ["/bin/vsphere-csi"]`, and then all of it's arguments are env vars, so there is no call to a `command` with arguments in the YAML at all.

CAPV uses `ENTRYPOINT`. I wasn't trying to standardize this portion of the Dockerfiles across the projects -- just use a multi-stage build that always used a known Golang image to build and use distroless if possible.

test/integration/Makefile
codenrhoden: For sure. I was holding off on that portion of the docs because I was afraid of creating a race condition where the docs point to the new place but the new place isn't working yet. My experience here has been once this PR goes in, it takes an iteration or two (usually 2-3 hours) between this repo and the Prow jobs to get everything working correctly. But that's a pretty narrow window, so it's probably better to just change it now. :)

cluster/images/controller-manager/Dockerfile
codenrhoden: yeah, I really wasn't intending to change this part... good catch, I'm going to change it back.

frapposelli: I'm not against changing it to `ENTRYPOINT` as this would remove the need to specify a full path to the binary in the yaml file, and I see that as a desirable change 😄

Since the manifests need to be changed anyway (given the new location of the containers) maybe it's the right time to standardize on `ENTRYPOINT`

manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       serviceAccountName: cloud-controller-manager
       containers:
         - name: vsphere-cloud-controller-manager
-          image: gcr.io/cloud-provider-vsphere/vsphere-cloud-controller-manager:latest
+          image: gcr.io/cloud-provider-vsphere/cpi/release/manager:latest
```
codenrhoden: @frapposelli So here is where I need your input for sure. What I did here was point to the latest *release* -- which won't exist once this merges until we tag a new release. To preserve the existing behavior, we would instead `s/release/ci`, which would always point to the latest build from the tip of `master`, which is actually what we are doing now.

So that begs a question -- what do we want our manifests to point users to by default? The latest *release*, or the latest build from master? I would argue for the latest release, but if we do that, then we have this temporary situation where once this goes in, there is no release to point to for some time.

That's why I was thinking about changing the manifests later -- it's okay to point to an older version until we are happy with this new tooling/release structure. Thoughts?

cluster/images/controller-manager/Dockerfile
```diff
@@ -10,8 +10,42 @@
...
 
-CMD ["/bin/vsphere-cloud-controller-manager"]
+################################################################################
+##                               MAIN STAGE                                   ##
+################################################################################
+# Copy the manager into the distroless image.
+FROM ${DISTROLESS_IMAGE}
+LABEL maintainer="Travis Rhoden <trhoden@vmware.com>"
+COPY --from=builder /build/vsphere-cloud-controller-manager /bin/vsphere-cloud-controller-manager
+CMD [ "/bin/vsphere-cloud-controller-manager" ]
```
codenrhoden: @frapposelli I changed this back, so no changes to manifests should be needed around this part. Unless `distroless` requires an `ENTRYPOINT` or something? That's I'm not sure of... But I think this should work.

manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
frapposelli: `latest` should be the latest tagged version, and should be the one referenced in the manifest.
`ci` or `<sha id>` should be the latest build from master, available for people to use but not reference in the examples.

Since we're almost ready to tag a new release, I'd like to keep refining this PR and then tag a new version as soon as we merge, this would put the new manifests in effect right away.

cluster/images/controller-manager/Dockerfile
frapposelli: we should be good with that

manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
codenrhoden: Agreed. In fact, what you will be able to do is use either `release/manager:latest`, which will be the latest "official release", OR use `ci/manager:latest` which is the most recent build from master. I like that they are separate. :) So, from your comments, I think I what I have in the manifests now is correct.

codenrhoden:
Doh, the unit-test is stuck (won't start)... Previously the only way I've found to fix this is a force-push.  but, just for fun...

/test pull-cloud-provider-vsphere-unit-test

codenrhoden:
Okay, two main changes based on @frapposelli's feedback...

The container images now use `ENTRYPOINT`, and the manifests are modified to no longer pass the path to the CPI binary.

The manifests are modified to point to `release/manager:latest`, which will always point to the latest **tagged, official** release, rather than a constantly changing tip of master (which you can still get by using `ci/manager:latest` if you want.

This is ready for a re-review whenever you like.

/hold cancel

codenrhoden:
I pushed a new version to fix one bug I found -- the `VERSION` build arg in the modified Dockerfile wasn't being used correctly. I needed to define it within the correct build stage instead of at the beginning. build args are only valid within the given build stages, so the previous method had no effect.

codenrhoden:
FYI, Once this goes in, I'll also introduce the GOPROXY changes from [here](https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/42). Which will speed up builds a good bit. We can probably ditch the CPI specific "ci" image for most builds as well -- except for conformance I believe. But that will need a more detailed look before changing anything.

manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
frapposelli: LGTM 👍

test/integration/Makefile
frapposelli: 👍

frapposelli:
Everything looks good!

/approve
/lgtm

I'm going to keep an hold on this PR until we are ready to tag, let's sync up on next steps at our monthly meeting tomorrow.

/hold

codenrhoden:
> I'm going to keep an hold on this PR until we are ready to tag, let's sync up on next steps at our monthly meeting tomorrow.

Okay. I'd advise putting everything in place a day or so before the tag, Once this is merged, we need to make all the Prow changes to make sure of the new targets, test that all the image/binary pushes are working, etc. It should be pretty smooth, and is a fast cycle to make changes, but still. =) We don't want to merge this and then tag immediately. Need at least a few hours in between.

codenrhoden:
@fabio since we are holding off on this, I'm going to go ahead and add the GOPROXY support, just like I did in https://github.com/kubernetes-sigs/vsphere-csi-driver/pull/42. It speeds up builds in Prow when using docker-in-docker.

frapposelli:
/lgtm
/hold cancel

@codenrhoden ready to merge this 👍

PR #237, opened 2019-08-20
Fix import syntax in cloudprovidervsphere.pb.go

codenrhoden:
Starting on 8/16, the `make fmt` target has been returning [an error](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-cloud-provider-vsphere-deploy/1162441111394848768) in the post-submit deploy step.

I'm not sure why this is only being caught in the post-submit and not the pre-submit, but sure enough the file is formatted funny with two different import sections. Since this is a generated file, I figured tried to regenerate teh `pb.go` file with `protoc` but that generated so many differences it didn't look right. So I'm just fixing this explicitly.

At some point, we need to visit generating the protobuf file again with the version we have in the `go.mod` file -- they are definitely differences.

w/o this fix, the post-submit deploy step was always failing. Except, with the new release tooling, this would actually succeed. We now assume that anything that makes it to `master` is okay to build, and we don't run the linting scripts as a pre-requisite. This should still get fixed, though.

**Release note**:

```release-note
NONE
```

/assign @frapposelli 
/hold
^ I'd like to see this get merged after https://github.com/kubernetes/test-infra/pull/13983, as it will be a good test to see that the post-submit deploy is working.

### DIFF
--- a/pkg/cloudprovider/vsphere/proto/cloudprovidervsphere.pb.go
+++ b/pkg/cloudprovider/vsphere/proto/cloudprovidervsphere.pb.go
@@ -8,9 +8,6 @@ import (
 	math "math"
 
 	proto "github.com/golang/protobuf/proto"
-)
-
-import (
 	context "golang.org/x/net/context"
 	grpc "google.golang.org/grpc"
 )


dvonthenen:
that's auto-generated code from gRPC. It's probably going to flag this each time we add new methods to it... just need to keep this in mind when updating.

/approve
/lgtm

codenrhoden:
> that's auto-generated code from gRPC. It's probably going to flag this each time we add new methods to it... just need to keep this in mind when updating.
> 

Yeah, I think it's worth re-visiting the generation of this file completely. I gave that a shot first -- I downloaded protoc 3.7.1 (the latest in Homebrew, but latest release is actually 3.9.x...), and then `go install`'d `protoc-gen-go` 1.3.2, which is the version of protobuf that we pin to in `go.mod`. Then I used `protoc` to regen that file and it was very, very different. Deleted a bunch of code and that caused things not to compile.

I didn't want to open that can of worms. :)

dvonthenen:
Yup, I can take a look at it right after I finish with the multi-service account feature.

Issue #233, opened 2019-08-16
Support different secret/secretNamespace per vCenter


akutz:
/kind feature

Currently the CCM only supports a single `secret` and `secretNamespace` for all vCenters as global properties. This issue tracks support for a `secret`/`secretNamespace` per vCenter.

dvonthenen:
Discussed with @akutz, got more requirements, and started implementation on it

dvonthenen:
Closing because PR https://github.com/kubernetes/cloud-provider-vsphere/pull/241 was merged

/close

dvonthenen:
@akutz This came up in the cloud-provider-vsphere meeting. The group just wanted to make sure that was implemented satisfies the needs you posted in this issue.

What was implemented allows for:
- a secret/secretNamespace per vCenter which maps to a given vCenter username/password
- allows for multiple connections to a vCenter with each connection holding a different username/password stored in a secret/secretNamespace. (ie using vCenter user accounts for role-based access to vCenter resources, such as but not limited to a cluster or resource pool.)

akutz:
lgtm!

PR #241, opened 2019-08-26
Initial commit for handling multiple secrets

dvonthenen:
This PR implements the following issue (https://github.com/kubernetes/cloud-provider-vsphere/issues/233) and a bonus feature! 😀

This allows each `[VirtualCenter]` to have their own dedicated Secret containing VC credentials (username/password). By having a dedicated Secret, each `[VirtualCenter]` will be isolated from accessing each other's Secret when said Secret is placed into their own namespace. Maybe an example vsphere.conf might be helpful...

```
[Global]
# properties in this section will be used for all specified vCenters unless overriden in VirtualCenter section.
port = "443" #Optional
insecure-flag = "1" #set to 1 if the vCenter uses a self-signed cert

[VirtualCenter "10.160.140.218"]
datacenters = "vicdc"
secret-name = "vccm-europe-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.

[VirtualCenter "10.161.60.78"]
datacenters = "vic0dc,vic1dc"
secret-name = "vccm-us-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.

[Labels]
region = k8s-region
zone = k8s-zone
```

The "bonus" feature is something I am calling TenantRefs. Prior to this PR, a single `[VirtualCenter]` was constrained to have either all or a subset of Datacenters contribute to the k8s cluster. Those DCs that weren't contributing to the cluster, couldn't have a different configuration because the  `[VirtualCenter]` "key" was the ip/fqdn of that vCenter Server. A TenantRefs is set by defining the `server` field under the `[VirtualCenter]` section. If that `server` field is set, it is assumed that the `value` in `[VirtualCenter "value"]` is the TenantRef. By implementing TenantRefs, this allows the potential for each VC/DC to have its own unique configuration based on the access controls of a given set of creds. So what does this feature translate into in English? Think of this as possibly enabling multi-tenancy. Maybe an example configuration might help...

```
[Global]
# properties in this section will be used for all specified vCenters unless overriden in VirtualCenter section.
port = "443" #Optional
insecure-flag = "1" #set to 1 if the vCenter uses a self-signed cert

[VirtualCenter "us-west-tenant1"]
server = "10.160.140.218"
datacenters = "tenant1"
secret-name = "vccm-tenant1-secret"
secret-namespace = "tenant1"
# port, insecure-flag will be used from Global section.

[VirtualCenter "us-east-tenant2"]
server = "10.161.60.78"
datacenters = "tenant2"
secret-name = "vccm-tenant2-secret"
secret-namespace = "tenant2"
# port, insecure-flag will be used from Global section.

[VirtualCenter "us-east-tenant3"]
server = "10.161.60.78"
datacenters = "us-east"
secret-name = "vccm-tenant3-secret"
secret-namespace = "tenant3"
# port, insecure-flag will be used from Global section.

[VirtualCenter "us-east-tenant4"]
server = "10.161.60.78"
datacenters = "us-east"
secret-name = "vccm-tenant4-secret"
secret-namespace = "tenant4"
# port, insecure-flag will be used from Global section.

[Labels]
region = k8s-region
zone = k8s-zone
```

Some things to note based on the example configuration above:
- the defined `TenantRef`s starting from the top of the config file to the bottom are: `us-west-tenant1`, `us-east-tenant2`, `us-east-tenant3`, and `us-east-tenant4`. TenantRefs must be unique!
- each tenant has their own secret which means each tenant has their own vSphere user account
- tenant3 and tenant4 are sharing a given DC. the resources each tenant is using can be assigned based on the vSphere user account (maybe each account has their own ResourcePool or ComputeCluster)
- tenant2 has its own dedicated DC but sharing the same VC as tenant3 and 4
- tenant1 has their own VC (and subsequently their own DC)

Here is an example vsphere.conf based on what I actually tested:
```
[Global]
# properties in this section will be used for all specified vCenters unless overridden in VirtualCenter section.
port = "443" #Optional
insecure-flag = "1" #set to 1 if the vCenter uses a self-signed cert

[VirtualCenter "europe"]
server = "10.160.140.218"
datacenters = "vicdc"
secret-name = "vccm-europe-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.

[VirtualCenter "us-west"]
server = "10.161.60.78"
datacenters = "vic0dc"
secret-name = "vccm-us-west-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.

[VirtualCenter "us-east"]
serever = "10.161.60.78"
datacenters = "vic1dc"
secret-name = "vccm-us-east-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.

[Labels]
region = k8s-region
zone = k8s-zone
```

Other minor changes along for the ride in this PR:
- cleaned up some functions in the Config package. it made more sense to make them member functions of the Config instead of passing the Config into the function
- the old Connect(vcServer string) in ConnectionManager is no longer sufficient to determine which VirtualCenterConfig is being referenced and since there is only one method for Connect() now, I renamed ConnectByInstance() to Connect(). Example of ambiguity in old Connect(vcServer string):
```
[VirtualCenter "10.161.60.78"]
datacenters = "vic0dc"
...

[VirtualCenter "10.161.60.78"]
datacenters = "vic1dc"
...
```
- Moved a number of consts and user-defined errors into their own go file. Adding more just started to make the go files with business logic larger in length (ie it's easier to manage this way and a number of package already do this).
- Removed the function removePortFromHost() because URL.Hostname() exists and returns the Hostname minus the port.
- Renamed SecretCredentialManager to CredentialManager because it actually does handle Creds from the Config as well. So the name should accurately reflect what it actually manages... ie all credentials.

**Which issue this PR fixes**: https://github.com/kubernetes/cloud-provider-vsphere/issues/233

Tested on vSphere 6.7u2
Test cases included:
- Using the backward compatible (aka Config format before this feature was introduced) with a single global secret
- Using the new `TenantRef`s with each their own dedicated Secret. In my test configuration, that's 3 Datacenters mapping to 3 `[VirtualCenter]` with 3 secrets.
- Using a hybrid of the previous two test cases. In other words, each `[VirtualCenter]` has their own `TenantRef`, 2 of the 3 `[VirtualCenter]` had their own dedicated Secret, and the 3rd `[VirtualCenter]` used the "global" Secret. This is a complicated use case which I am sure no one will use... the point in testing this just means I have confidence everything is working fine. 🙂

Since this feature is a pretty significant change and because the default/current/backward-compatible vsphere.conf file format still works, I am going to do the documentation in a separate PR. This is mainly due to the fact that this will impact CSI and I would like to start the integration of this feature into CSI before the new CNS based implementation PR start making their way over and I lose the ability to accurately test this feature on CSI.

**Release note**:
Default/current/backward-compatible vsphere.conf still works!


dvonthenen:

dvonthenen:
Updated based feedback from @andrewsykim and updated the description (comment #1) of the PR for consistency.

The good news is by changing how we enable ConnectionLabels from what was first implemented, it's made the code much simpler. The short version of this change is you enable the ConnectionLabel by defining the `server` field under the `[VirtualCenter]` section. When `server` which contains the ip/fqdn of the vCenter Server is set, the `value` in `[VirtualCenter "value"]` is the ConnectionLabel. Check the updated description for config examples. The logic now is 💯

Thanks @andrewsykim !!

dvonthenen:
Yup, we are good to go!

pkg/common/config/config.go
```diff
@@ -401,6 +399,10 @@ func validateConfig(cfg *Config) error {
 					return ErrPasswordMissing
 				}
 			}
+		} else if cfg.IsSecretInfoProvided() && !vcConfig.IsSecretInfoProvided() {
+			vcConfig.SecretHolder = DefaultCredentialManager
+		} else if vcConfig.IsSecretInfoProvided() {
+			vcConfig.SecretHolder = vcConfig.SecretName + "/" + vcConfig.SecretNamespace
```
andrewsykim: oops I think earlier I meant that this should be `namespace/name` which is consistent how resources are key'd in client-go

andrewsykim: Though I see now it doesn't really matter like you mentioend

pkg/common/config/types.go
```diff
@@ -81,6 +81,14 @@ type VirtualCenterConfig struct {
 	User string `gcfg:"user"`
 	// vCenter password in clear text.
 	Password string `gcfg:"password"`
+	// ConnectionLabel (intentionally not exposed via the config) is a unique label to
+	// be used in place of the vcServer as the primary connection key. If one label is set,
+	// all virtual center configs must have a unique label.
+	ConnectionLabel string
```
andrewsykim: Wondering if `ConnectionLabel` is still the best name for this field since it's just a unique reference for a vCenter? A few off the top of my head:
* ID 
* Name 
* Key
* ConnectionKey

pkg/common/config/types.go
```diff
@@ -94,11 +102,18 @@ type VirtualCenterConfig struct {
 	CAFile string `gcfg:"ca-file"`
 	// Thumbprint of the VCenter's certificate thumbprint
 	Thumbprint string `gcfg:"thumbprint"`
+	// Secret Holder (intentionally not exposed via the config) is a label identify which
+	// InformerManager holds the secret
+	SecretHolder string
+	// Name of the secret were vCenter credentials are present.
```
andrewsykim: s/were/where

pkg/common/config/types.go
```diff
@@ -94,11 +102,18 @@ type VirtualCenterConfig struct {
...
+	SecretName string `gcfg:"secret-name"`
+	// Secret Namespace where secret will be present that has vCenter credentials.
```
andrewsykim: "Namespace where secret will be.."

pkg/common/config/types.go
```diff
@@ -94,11 +102,18 @@ type VirtualCenterConfig struct {
 	CAFile string `gcfg:"ca-file"`
 	// Thumbprint of the VCenter's certificate thumbprint
 	Thumbprint string `gcfg:"thumbprint"`
+	// Secret Holder (intentionally not exposed via the config) is a label identify which
+	// InformerManager holds the secret
+	SecretHolder string
```
andrewsykim: Thoughts on `SecretRef` instead since this is just a reference to a secret resource right?

pkg/common/config/config.go
dvonthenen: Yea, not a problem.

pkg/common/config/types.go
dvonthenen: I think at a high level it is. Also, from an end-user point of view, it is as well. Referring to it as any of those things doesn't have real meaning to an end-user.

Having said that, internally in code, yes it's a key. I don't care either way what the variable is called, but in the documentation, I think it most definitely should be a label. For understandability, if the docs say label, do we want to be consistent?

pkg/common/config/types.go
dvonthenen: Doesn't matter what the name is since it's internal. Will use SecretRef.

pkg/common/config/types.go
andrewsykim: Good point, I'm okay to merge as `ConnectionLabel` but when we get to docs I think it's worth another discussion to see if there's a better name for this.

pkg/common/config/config_test.go
```diff
@@ -33,6 +33,26 @@ datacenters = us-west
...
+insecure-flag = true
+
+[VirtualCenter "us-west"]
+server = "10.0.0.1"
+datacenters = "vic0dc"
+secret-name = "us-west-secret"
+secret-namespace = "kube-system"
+# port, insecure-flag will be used from Global section.
+
+[VirtualCenter "us-east"]
```
andrewsykim: Can we have a 3rd entry here that still uses the server IP as the connection label?
```
[VirtualCenter "10.0.0.3"]
datacenters = "vic0dc"
secret-name = "us-west-secret"
secret-namespace = "kube-system"
# port, insecure-flag will be used from Global section.
```

pkg/common/config/types.go
dvonthenen: If you still want to change in code, better to do it now. Tell me what you want to use.

pkg/common/connectionmanager/consts_and_errors.go
```diff
@@ -0,0 +1,69 @@
...
+	NumConnectionAttempts int = 3
+
+	// RetryAttemptDelaySecs is the number of seconds to wait between
+	// connection attempts.
+	RetryAttemptDelaySecs int = 1
+)
+
+// Error Messages
+const (
+	ConnectionNotFoundErrMsg       = "vCenter not found"
```
andrewsykim: These constants are redundant since we only ever access the error vars?

dvonthenen: they are... these were initially pulled from the in-tree provider and we never changed the format and kept adding to it. i am guessing some pieces of code might print the message in klog and then return the Err. I can't be certain.

pkg/common/config/types.go
dvonthenen: Updating to TenantRef

PR #242, opened 2019-08-28
Add a staticcheck verify script and fix errors

codenrhoden:
I figured since k/k now has support for running `staticcheck`, we should do it on our projects.

This patch adds a hack/verify-staticcheck.sh script, adds a Makefile
target for "make staticcheck", and fixes the errors found.

A couple of the other check scripts (format, lint) or modified to no
longer use `GO111MODULE=off` when they "go get" the needed tools. That
was there before to avoid having `go.mod` change when running the
scripts, but instead this patch just allows in the needed changes to
go.mod and the correct version should be installed every time. no need
to fight the tool. The `go get` commands are still needed in the
scripts, however, even if a `go mod download` has been done, because the
download won't install the tool, but `go get` will.

The initial list of errors found were:

```sh
cmd/vcpctl/provision/provision.go:40:2: var datacenter is unused (U1000)
cmd/vcpctl/provision/provision.go:42:2: var secretName is unused (U1000)
cmd/vcpctl/provision/provision.go:44:2: var secretNamespace is unused (U1000)
pkg/cli/cli.go:89:10: this value of err is never used (SA4006)
pkg/cli/user.go:44:2: field password is unused (U1000)
pkg/cloudprovider/vsphere/cloud.go:151:2: should merge variable declaration with assignment on next line (S1021)
pkg/common/config/config.go:438:4: this value of insecure is never used (SA4006)
pkg/common/connectionmanager/search.go:103:7: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/search.go:148:8: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/search.go:279:7: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/search.go:324:8: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/zones.go:178:7: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/zones.go:223:8: should omit comparison to bool constant, can be simplified to found (S1002)
pkg/common/connectionmanager/zones.go:330:36: should use make(map[string]string) instead (S1019)
pkg/common/vclib/pbm.go:68:6: should omit nil check; len() for nil slices is defined as zero (S1009)
pkg/common/vclib/utils.go:41:2: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (S1008)
pkg/common/vclib/utils.go:130:15: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (S1007)
pkg/common/vclib/virtualmachine.go:220:2: should replace loop with dsRefList = append(dsRefList, hostSystemMo.Datastore...) (S1011)
pkg/common/vclib/virtualmachine.go:292:3: should merge variable declaration with assignment on next line (S1021)
```

**Release note**:

```release-note
NONE
```

/assign @andrewsykim  
/assign @dvonthenen


codenrhoden:
/hold cancel
@dvonthenen this is ready for your re-review. I rebased it after your new stuff went in.

Issue #243, opened 2019-08-29
Update NewInformer to reuse SharedInformerFactory but create NewLister


dvonthenen:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
This is an enhancement to decrease the number of watchers generated.

**What you expected to happen**:
Everything to still work except to use less resources.

dvonthenen:
Currently, taking a look at this.

PR #244, opened 2019-09-04
Refactor using single SharedInformer

dvonthenen:
This was some fallout from creating listers for multiple secrets for VC creds. This refactors NewInformer so that only a single SharedInformer is created. In English, we only have a single watcher for all secrets.

**Which issue this PR fixes** : fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/243

Tested on k8s 1.14.1 on vsphere 6.7u2 using multiple secrets.

**Release note**:
NA

### DIFF
--- a/pkg/common/kubernetes/informers.go
+++ b/pkg/common/kubernetes/informers.go
@@ -30,17 +30,30 @@ func noResyncPeriodFunc() time.Duration {
 	return 0
 }
 
-var signalHandler <-chan struct{}
+var (
+	signalHandler   <-chan struct{}
+	informerFactory informers.SharedInformerFactory
+)
 
 // NewInformer creates a newk8s client based on a service account
-func NewInformer(client clientset.Interface) *InformerManager {
+func NewInformer(client clientset.Interface, singleWatcher bool) *InformerManager {
 	if signalHandler == nil {
 		signalHandler = signals.SetupSignalHandler()
+	}
+
+	var myInformerFactory informers.SharedInformerFactory
+	if singleWatcher {
+		if informerFactory == nil {
+			informerFactory = informers.NewSharedInformerFactory(client, noResyncPeriodFunc())
+		}
+		myInformerFactory = informerFactory
+	} else {
+		myInformerFactory = informers.NewSharedInformerFactory(client, noResyncPeriodFunc())
 	}
 	return &InformerManager{
 		client:          client,
 		stopCh:          signalHandler,
-		informerFactory: informers.NewSharedInformerFactory(client, noResyncPeriodFunc()),
+		informerFactory: myInformerFactory,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


dvonthenen:
/hold cancel

This is ready for review...

CC: @andrewsykim @frapposelli

pkg/common/kubernetes/informers.go
```diff
@@ -30,17 +30,30 @@ func noResyncPeriodFunc() time.Duration {
 	return 0
 }
 
-var signalHandler <-chan struct{}
+var (
+	signalHandler   <-chan struct{}
+	informerFactory informers.SharedInformerFactory
```
andrewsykim: Not a fan of global var for the informer factory, is there a way to build the informer factory in Initialize() and just pass down only the needed informers down to connection manager (i.e. `informerFactory.Core().Secrets()`)?

andrewsykim: hmmm okay, I see now that would require a larger refactory on the InformerManager

dvonthenen: I think it's ok as a global. The global `informerFactory` isn't exported out of the package/process so it can't be re-used by other processes. Since this global is confined to this package, you are getting exactly what you want a single informer factory.

dvonthenen: If you really do want to pass in the variable, it would require a second function for NewInformer(WithProvidedFactory) and also a second function to InitializeSecretLister(WithProvidedFactory), but it would still accomplish effectively the same thing with a private global.

pkg/common/kubernetes/informers.go
```diff
@@ -30,17 +30,30 @@ func noResyncPeriodFunc() time.Duration {
...
+)
 
 // NewInformer creates a newk8s client based on a service account
-func NewInformer(client clientset.Interface) *InformerManager {
+func NewInformer(client clientset.Interface, singleWatcher bool) *InformerManager {
```
andrewsykim: Do we need the singleWatcher toggle? In what scenario would we _not_ want a single informer for all lister/watchers?

pkg/common/kubernetes/informers.go
andrewsykim: So I think my biggest concern with this is that you would call `informerFactory.Start()` multiple times for multiple `InformerManagers` using the same shared informer factory. Ideally you construct the informer factory in one place and pass that into various types that need resource-specific informers and you would call `informerFactory.Start()` once, ideally in the same place you constructed it. With that said, I think the real problem isn't the global var but the `InformerManager` type which makes it hard to easily pass down the informer factory but probably don't want to address that in this PR.

pkg/common/kubernetes/informers.go
dvonthenen: Yea, we don't have a need for it currently. Was just trying to preserve the option. Only reason for it would be having multiple client connections based on individual service accounts.

If you want to remove it and tackle it when we get there, that cool. Just let me know.

pkg/common/kubernetes/informers.go
dvonthenen: Yea, I looked at the code behind `Start()` and also tested to make sure the behavior is correct. That's why I made the update to the function comment. You can call `Start()` multiple times as you add in new Listers and it only inits the new ones.

The downside of getting all of the Listers upfront, is it complicates the CPI init logic a lot.

PR #245, opened 2019-09-04
refactor to initialize shared informer only once

andrewsykim:

Follow-up PR to https://github.com/kubernetes/cloud-provider-vsphere/pull/244 - which also ensures the single shared informer is initialized only once and removes the need for the `InformerManager` type. 

**Release note**:

```release-note
NONE
```

/assign @dvonthenen


andrewsykim:
/close

Other vSphere components (like CSI) use what's in `pkg/common`, so going to close this for now so we keep those libraries consistent.

Issue #259, opened 2019-10-07
Remove ServiceAccount in Cloud Config


andrewsykim:
/kind feature

In https://github.com/kubernetes/cloud-provider-vsphere/commit/3c937e44618f3e0792bbf6f3fbe71a21c1024673#diff-eff14abf49a482c341f44fe1f99cbd27 we added a ServiceAccount field in the cloud config. Turns out we were mostly using that field as the client's user agent, not the client's service account. That should probably be removed from the cloud config cause we don't have any reason to override the user agent per tenant. Should be safe to remove since we haven't cut a release with the initial commit yet. 

cc @dvonthenen

PR #260, opened 2019-10-08
remove ServiceAccount in cloud config

andrewsykim:

Removes ServiceAccount in cloud config which we were using as the client's user agent. The user agent can just be a well-known constant and shouldn't be configurable. The name "ServiceAccount" also implies that changing this field changes the Kubernetes service account the controller manager uses which can be confusing. 

Fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/259 

**Release note**:

```release-note
Removes ServiceAccount in cloud config
```

### DIFF
--- a/pkg/common/config/types.go
+++ b/pkg/common/config/types.go
@@ -43,9 +43,6 @@ type Config struct {
 		SecretName string `gcfg:"secret-name"`
 		// Secret Namespace where secret will be present that has vCenter credentials.
 		SecretNamespace string `gcfg:"secret-namespace"`
-		// The kubernetes service account used to launch the cloud controller manager.
-		// Default: cloud-controller-manager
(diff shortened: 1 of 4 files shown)


pkg/common/config/types.go
```diff
@@ -43,9 +43,6 @@ type Config struct {
 		SecretName string `gcfg:"secret-name"`
 		// Secret Namespace where secret will be present that has vCenter credentials.
 		SecretNamespace string `gcfg:"secret-namespace"`
-		// The kubernetes service account used to launch the cloud controller manager.
-		// Default: cloud-controller-manager
-		ServiceAccount string `gcfg:"service-account"`
```
andrewsykim: (not a breaking change since we haven't cut a new release after we added this)

Issue #270, opened 2019-10-22
Both InternalIP and ExternalIP gets created for first IP on each vNIC


zuzzas:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

If there are multiple vNICs, then for the first address on each vNICs `status.addresses` on the Node object would contain both InternalIP and ExternalIP address type.

**What you expected to happen**:

I'd like to have a way to specify exact network name, so that when vsphere CCM iterates over vNICs, it detects correct Node address type.

**How to reproduce it (as minimally and precisely as possible)**:

Specify multiple vNICs on VM creation, and assign IPs to each one. Then observe `.status.addresses` field on the Node object.

dvonthenen:
I am trying to test this feature against what customers might do in the wild and even in my own test environment, what we thought was sufficient turns out it might not be. I'm re-opening this issue to block the 1.1 release. Going to think about this a little more and then update the implementation.

frapposelli:
closed via #284 

/close

Issue #266, opened 2019-10-11
Cloud controller is unable to patch node with IP addresses from cloud provider


ivanfilippov:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:
Cloud controller is unable to patch node with IP addresses from cloud provider on a multi-NIC VM.

```
I1011 05:22:49.646435       1 instances.go:158] instances.InstanceExistsByProviderID() CACHED with 42348b4e-4ba1-e452-e946-6d0e9a323160
I1011 05:22:49.646457       1 instances.go:78] instances.NodeAddressesByProviderID() CACHED with 42348b4e-4ba1-e452-e946-6d0e9a323160
E1011 05:22:49.656899       1 node_controller.go:188] Error patching node with cloud ip addresses = [failed to patch status "{\"status\":{\"$setElementOrder/addresses\":[{\"type\":\"ExternalIP\"},{\"type\":\"InternalIP\"},{\"type\":\"Hostname\"},{\"type\":\"ExternalIP\"},{\"type\":\"InternalIP\"},{\"type\":\"ExternalIP\"},{\"type\":\"InternalIP\"}],\"addresses\":[{\"address\":\"172.16.202.44\",\"type\":\"ExternalIP\"},{\"address\":\"172.16.201.44\",\"type\":\"ExternalIP\"},{\"address\":\"172.16.45.44\",\"type\":\"ExternalIP\"},{\"address\":\"172.16.202.44\",\"type\":\"InternalIP\"},{\"address\":\"172.16.201.44\",\"type\":\"InternalIP\"},{\"address\":\"172.16.45.44\",\"type\":\"InternalIP\"}]}}" for node "k8s-prod5-worker-2": The order in patch list:
[map[address:172.16.202.44 type:ExternalIP] map[address:172.16.201.44 type:ExternalIP] map[address:172.16.45.44 type:ExternalIP] map[address:172.16.202.44 type:InternalIP] map[address:172.16.201.44 type:InternalIP] map[address:172.16.45.44 type:InternalIP]]
 doesn't match $setElementOrder list:
[map[type:ExternalIP] map[type:InternalIP] map[type:Hostname] map[type:ExternalIP] map[type:InternalIP] map[type:ExternalIP] map[type:InternalIP]]
]
```

**What you expected to happen**:
Nodes are patched with IP addresses from cloud provider.

**How to reproduce it (as minimally and precisely as possible)**:
[...]

PR #271, opened 2019-10-22
Locate InternalIP and ExternalIP by vSphere Network name

zuzzas:

Currently, vsphere CCM will iterate over physical vNICs and setup first IP address on each as both *InternalIP* and  . This behavior is undesirable, if we are aware of vSphere Network names.

fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/270

### DIFF
--- a/pkg/cloudprovider/vsphere/config.go
+++ b/pkg/cloudprovider/vsphere/config.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2019New The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsphere
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"gopkg.in/gcfg.v1"
+)
+
+// CPIConfig is used to read and store information (related only to the CPI) from the cloud configuration file
+type CPIConfig struct {
+	Nodes struct {
+		// IP address on VirtualMachine's network interfaces included in the fields' CIDRs
+		// that will be used in respective status.addresses fields.
+		InternalNetworkSubnetCIDR string `gcfg:"internal-network-subnet-cidr"`
+		ExternalNetworkSubnetCIDR string `gcfg:"external-network-subnet-cidr"`
+	}
+}
+
+// FromEnv initializes the provided configuratoin object with values
+// obtained from environment variables. If an environment variable is set
+// for a property that's already initialized, the environment variable's value
+// takes precedence.
+func (cfg *CPIConfig) FromEnv() {
+	if v := os.Getenv("VSPHERE_NODES_INTERNAL_NETWORK_SUBNET_CIDR"); v != "" {
+		cfg.Nodes.InternalNetworkSubnetCIDR = v
+	}
+
+	if v := os.Getenv("VSPHERE_NODES_EXTERNAL_NETWORK_SUBNET_CIDR"); v != "" {
+		cfg.Nodes.ExternalNetworkSubnetCIDR = v
+	}
+}
+
+// ReadCPIConfig parses vSphere cloud config file and stores it into CPIConfig.
+// Environment variables are also checked
+func ReadCPIConfig(config io.Reader) (*CPIConfig, error) {
+	if config == nil {
+		return nil, fmt.Errorf("no vSphere cloud provider config file given")
+	}
+
+	cfg := &CPIConfig{}
+
+	if err := gcfg.FatalOnly(gcfg.ReadInto(cfg, config)); err != nil {
+		return nil, err
+	}
+
+	// Env Vars should override config file entries if present
+	cfg.FromEnv()
+
+	return cfg, nil
+}
--- a/pkg/cloudprovider/vsphere/config_test.go
+++ b/pkg/cloudprovider/vsphere/config_test.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsphere
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+const basicConfig = `
+[Nodes]
+internal-network-subnet-cidr = 192.0.2.0/24
+external-network-subnet-cidr = 198.51.100.0/24
+`
+
+func TestReadConfigGlobal(t *testing.T) {
+	_, err := ReadCPIConfig(nil)
+	if err == nil {
+		t.Errorf("Should fail when no config is provided: %s", err)
+	}
+
+	cfg, err := ReadCPIConfig(strings.NewReader(basicConfig))
+	if err != nil {
+		t.Fatalf("Should succeed when a valid config is provided: %s", err)
+	}
+
+	if cfg.Nodes.InternalNetworkSubnetCIDR != "192.0.2.0/24" {
+		t.Errorf("incorrect vcenter ip: %s", cfg.Nodes.InternalNetworkSubnetCIDR)
+	}
+
+	if cfg.Nodes.ExternalNetworkSubnetCIDR != "198.51.100.0/24" {
+		t.Errorf("incorrect datacenter: %s", cfg.Nodes.ExternalNetworkSubnetCIDR)
+	}
+}
+
+func TestEnvOverridesFile(t *testing.T) {
+	subnet := "203.0.113.0/24"
+	os.Setenv("VSPHERE_NODES_INTERNAL_NETWORK_SUBNET_CIDR", subnet)
+	defer os.Unsetenv("VSPHERE_NODES_INTERNAL_NETWORK_SUBNET_CIDR")
+
+	cfg, err := ReadCPIConfig(strings.NewReader(basicConfig))
+	if err != nil {
+		t.Fatalf("Should succeed when a valid config is provided: %s", err)
+	}
+
+	if cfg.Nodes.InternalNetworkSubnetCIDR != subnet {
+		t.Errorf("expected subnet: \"%s\", got: \"%s\"", subnet, cfg.Nodes.InternalNetworkSubnetCIDR)
+	}
+}
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -190,6 +190,29 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		klog.Warningf("Unable to find vcInstance for %s. Defaulting to ipv4.", tenantRef)
 	}
 
+	var internalNetworkSubnet *net.IPNet
+	var externalNetworkSubnet *net.IPNet
+
+	if nm.cpiCfg != nil {
+		if nm.cpiCfg.Nodes.InternalNetworkSubnetCIDR != "" {
+			_, internalNetworkSubnet, err = net.ParseCIDR(nm.cpiCfg.Nodes.InternalNetworkSubnetCIDR)
+			if err != nil {
+				return err
+			}
+		}
+		if nm.cpiCfg.Nodes.ExternalNetworkSubnetCIDR != "" {
+			_, externalNetworkSubnet, err = net.ParseCIDR(nm.cpiCfg.Nodes.ExternalNetworkSubnetCIDR)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	var addressMatchingEnabled bool
+	if internalNetworkSubnet != nil || externalNetworkSubnet != nil {
+		addressMatchingEnabled = true
+	}
+
 	found := false
 	addrs := []v1.NodeAddress{}
 	for _, v := range oVM.Guest.Net {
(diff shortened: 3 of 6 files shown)


zuzzas:
Thanks! Going to fix tests today.

zuzzas:
Some tests have failed, but these are not related to my changes.

dvonthenen:
We need to carefully think about if those new fields in the config belong in `pkg/common` level. I need to think about this some more, but I think might initial gut reaction is that this might be the first time we need to create an outer CPI specific config since this might only apply to the CPI.

What do you think @andrewsykim ?

andrewsykim:
My initial question is if we should accept network names or their subnets 🤔 need to noodle on that one too

dvonthenen:
> My initial question is if we should accept network names or their subnets 🤔 need to noodle on that one too

If we do "VM Network" names (this should be done per VC and not in [global] also), there would be a strict requirement that each NIC must have only a single IP bound to it.

andrewsykim:
> there would be a strict requirement that each NIC must have only a single ipv4 and ipv6 bound to it.

So if NICs can be bound to more than 1 address it sounds like it should be based on subnet? @zuzzas thoughts?

dvonthenen:
Also, if there is only a single ipv4 or ipv6 address, we shouldn't force/require setting those new config options. It should just work.

andrewsykim:
Fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/266

zuzzas:
I'll refactor the config and contribute to the discussion about network names vs subnet CIDR tomorrow.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -198,29 +248,77 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+					if internalNameMatched || internalSubnetMatched {
+						klog.V(2).Infof("Adding Internal IP: %s", ip)
+						v1helper.AddToNodeAddresses(&addrs,
+							v1.NodeAddress{
+								Type:    v1.NodeInternalIP,
+								Address: ip,
+							},
+						)
+						found = true
+						break
```
zuzzas: 1. Should we or should we not `break` once we "classify" an ip address as Internal, without giving it a chance to be classified as External?
2.  Should we or should we not `break` once we actually find a matching ip address? Perhaps we should collect all available matches?

Here's how different implementations approach the *second* issue. You'll see links to both in-tree and CPI providers, since some CPI providers (notably, GCP and AWS) simply include legacy in-tree providers.

1. AWS. Uses all InternalIPs found: https://github.com/kubernetes/legacy-cloud-providers/blob/master/aws/aws.go#L1434-L1445
2. GCP. Uses the first InternalIP found and the first ExternalIP found: https://github.com/kubernetes/legacy-cloud-providers/blob/master/gce/gce_instances.go#L87-L94
3. Openstack. Uses all addresses found, classifying them based on config first: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack.go#L687-L711

andrewsykim:
>I've started to muse about pros and cons of L2 vs L3 matching (network names vs IPs), but the point was moot, since a user might find both approaches desirable, so I've just coded both variants.

My thinking here is that because Kubernetes node addresses are strictly L3, the config we expose to users should also strictly be L3. I'm interested in the use-cases for L2 though -- is it possible to specify a network name that gives me more than 2 addresses for a node? If so what should be the expected behavior?

pkg/cloudprovider/vsphere/nodemanager.go
andrewsykim: I think for now, we should stick to 1 InternalIP and  1 ExternalIP -- last time I checked kubelet always uses the first InternalIP it sees, which makes the behavior hard to predict since ordering is not always guaranteed. What do you think?

pkg/cloudprovider/vsphere/config.go
```diff
@@ -0,0 +1,86 @@
...
+// CPIConfig is used to read and store information (related only to the CPI) from the cloud configuration file
+type CPIConfig struct {
+	CPI struct {
+		// address options for each network type (internal and/or external) are mutually inclusive
+		// e.g., if you specify "internal-network-name" and "internal-network-subnet-cidr"
+		// only the addresses that match BOTH will match
+
+		// VirtualMachine network interfaces backed by the following vSphere Networks
+		// will be used in respective status.addresses fields.
+		InternalNetworkName []string `gcfg:"internal-network-name"`
```
andrewsykim: How do we feel about making these accept just 1 entry and expanding this to a list in the future?

andrewsykim: I'm skeptical about use-cases involving two networks per network type, if we have valid use-cases that work when specifying more than 1 InternalIP for example, then I think we should consider this. I personally have not seen this work on vSphere (yet).

zuzzas: I've made a separate commit to roll back changes to a single possible network. I've pondered upon the use cases, and I've found a single justification for multiple network names. This CPI implementation supports **multiple** vCenters. There could be a slim chance of a having a *single* L2 segment (e.g., VLAN) under different *network names* in vCenters. 

But if we are moving away from network names as a selector, then this isn't valid anymore. The L3 IP subnet will stay the same regardless of any vCenter configurations.

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: I agree. Multiple IPs per type will needlessly complicate things. Until a concrete use case appears, ofc.

pkg/cloudprovider/vsphere/config.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"gopkg.in/gcfg.v1"
+)
+
+// CPIConfig is used to read and store information (related only to the CPI) from the cloud configuration file
+type CPIConfig struct {
```
dvonthenen: Thanks for the update on this PR! I think we might be just about there. We really need to carefully think about the config and thinking out loud here (not sure if this is a good idea or not), maybe we might be able to do something to the effect of...

```
type CPIConfig struct {
  vcfg.Config

  CPI struct {
    // ... comments ...
    InternalNetworkSubnetCIDR string `gcfg:"internal-network-subnet-cidr"`
    ExternalNetworkSubnetCIDR string `gcfg:"external-network-subnet-cidr"`
  }
}
```

That way the shared config is still a single configuration object versus having two separate config objects (one which is shared and a separate CPI one). CSI specific fields would then go into a `CSI struct` similar to this CPI.

pkg/cloudprovider/vsphere/config.go
andrewsykim: sounds good to me :)

pkg/cloudprovider/vsphere/config.go
andrewsykim: So this is going to add a `[CPI]` section to the config file? Wondering if something more descriptpive is useful to users, maybe `[Nodes]`  or something else

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -204,23 +227,62 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
 						Address: oVM.Guest.HostName,
 					},
 				)
-
-				found = true
-				break
+				for _, ip := range ips {
+					parsedIP := net.ParseIP(ip)
+					if parsedIP == nil {
+						return err
```
andrewsykim: Which err is this?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -204,23 +227,62 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+					if externalNetworkSubnet != nil && externalNetworkSubnet.Contains(parsedIP) {
+						klog.V(2).Infof("Adding External IP: %s", ip)
+						v1helper.AddToNodeAddresses(&addrs,
+							v1.NodeAddress{
+								Type:    v1.NodeExternalIP,
+								Address: ip,
+							},
+						)
+						found = true
+						break
```
andrewsykim: So I don't think we can actually break here because that means we only store the first address we see

andrewsykim: We might have to store addresses found in a map and only add the specific address type if it doesn't already exist in the map

dvonthenen:
In regards to the go tests failing... I have a PR to fix one and @codenrhoden is taking a look at the other one. Tracking both in this PR https://github.com/kubernetes/cloud-provider-vsphere/pull/273

pkg/cloudprovider/vsphere/config.go
dvonthenen: Yes, it would add a `[CPI]` section... yea I think something more descriptive might be better. I could definitely get behind something like `[Nodes]`.

dvonthenen: I think it should be ok... But just need to float the idea out there that if we call it something like `[Nodes]` on the CSI side they could come up with a similar section, have a parameter under that section, but that parameter has a completely different meaning than the CPI version of it.

dvonthenen: The config parsing does support subsections... we could do `[CPI "Nodes"] (then CSI could do [CSI "Nodes"])`

dvonthenen:
/retest

This should at least pass the CI now we have merged the two PRs to fix the unit and integration tests. Still need to resolve the overall discussion with the PR.

pkg/cloudprovider/vsphere/config.go
dvonthenen: I discussed this with @andrewsykim and the CSI repo maintainers and the consensus is to use `[Nodes]` to contain:
```
InternalNetworkSubnetCIDR string `gcfg:"internal-network-subnet-cidr"`
ExternalNetworkSubnetCIDR string `gcfg:"external-network-subnet-cidr"`
```

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: Could you elaborate on this further?
>that means we only store the first address we see

Isn't that exactly what [we want](https://github.com/kubernetes/cloud-provider-vsphere/pull/271#discussion_r341680191)?

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: It's terribly easy to stuff Go code with `if err != nil`s, isn't it? :) 
Fixed.

pkg/cloudprovider/vsphere/config.go
zuzzas: It is done. I've also added tests for the new config.

pkg/cloudprovider/vsphere/nodemanager.go
andrewsykim: We want first address we see of each type, I think at the current state it's only the first address of first type we see, ignoring the other type

andrewsykim: (because we iterate the same for loop for both types)

pkg/cloudprovider/vsphere/config_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+)
+
+const basicConfig = `
+[Nodes]
+internal-network-subnet-cidr = 192.0.2.0/24
+external-network-subnet-cidr = 198.51.100.0/24
+`
+
+func TestReadConfigGlobal(t *testing.T) {
```
andrewsykim: Nice! Thanks for adding tests!

pkg/cloudprovider/vsphere/config_test.go
```diff
@@ -0,0 +1,48 @@
+package vsphere
```
andrewsykim: Can you add the license header here like the other files (dated 2019)?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -204,23 +227,62 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		for _, family := range ipFamily {
 			ips := returnIPsFromSpecificFamily(family, v.IpAddress)
 
-			for _, ip := range ips {
-				klog.V(2).Infof("Adding IP: %s", ip)
+			if addressMatchingEnabled {
```
andrewsykim: Can we also log the hostname value here?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -204,23 +227,62 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
 						Address: oVM.Guest.HostName,
 					},
 				)
-
-				found = true
-				break
+				for _, ip := range ips {
+					parsedIP := net.ParseIP(ip)
+					if parsedIP == nil {
+						return fmt.Errorf("can't parse IP \"%s\"", ip)
```
andrewsykim: If you use %q you get the quotes for free :)

pkg/cloudprovider/vsphere/config_test.go
dvonthenen: Sweet!

pkg/cloudprovider/vsphere/config_test.go
zuzzas: Done.

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: Done.

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: Thanks, didn't know that! But I've opted to use the same style as other logging statements.

pkg/cloudprovider/vsphere/nodemanager.go
zuzzas: Oh my god... I've been looking at this `break` statement, and all I've seen was a `continue` statement. Wow.

If this part of the code looks ugly to you, perhaps, I can move it to a function and write a test? Or is it good as it is?

andrewsykim: Looks great now, thank you!

andrewsykim:
/lgtm 

Can you squash commits and use a more descriptive message please?

zuzzas:
@andrewsykim 
It is done.

andrewsykim:
/lgtm

Leaving final approval to @dvonthenen

andrewsykim:
Thanks for @zuzzas !

dvonthenen:
Seriously! Awesome contribution @zuzzas!

/approve
/lgtm

PR #272, opened 2019-11-03
[WIP] add experimental support for Service Type=LoadBalancer w/ NSX-T

andrewsykim:

Adds experimental support for NSX-T load balancers when a Service Type=LoadBalancer is created in a cluster using the vSphere cloud provider. 

This PR is still a work in progress, some pending tasks include but are not limited to:
* a lot more testing/validation (manual, unit tests, etc)
* agreement on the config file changes required to enable NSX-T, for now I just put whatever I needed to get this working in the config file but I think we should discuss what makes the most sense there, there are some values that are also hardcoded (like LB Service size), we should discuss how we want them to be configurable
* how to mark this feature as "experimental", maybe an env var gate? Ideally something not in the config file cause changing config file fields in the future is painful.

For some reason listing loadbalancer resources was not available on the go-vmware-nsxt client so I had to write functions to do this using http stdlib. Opened https://github.com/vmware/go-vmware-nsxt/issues/18 in hopes to get it in the client.

**Release note**:

```release-note
add experimental support for Service Type=LoadBalancer w/ NSX-T
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
+	github.com/vmware/go-vmware-nsxt v0.0.0-20190201205556-16aa0443042d
 	github.com/vmware/govmomi v0.21.0
 	golang.org/x/lint v0.0.0-20190409202823-959b441ac422 // indirect
 	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80
--- a/go.sum
+++ b/go.sum
@@ -314,6 +314,8 @@ github.com/tmc/grpc-websocket-proxy v0.0.0-20170815181823-89b8d40f7ca8/go.mod h1
 github.com/urfave/negroni v1.0.0/go.mod h1:Meg73S6kFm/4PpbYdq35yYWoCZ9mS/YSx+lKnmiohz4=
 github.com/vishvananda/netlink v0.0.0-20171020171820-b2de5d10e38e/go.mod h1:+SR5DhBJrl6ZM7CoCKvpw5BKroDKQ+PJqOg65H/2ktk=
 github.com/vishvananda/netns v0.0.0-20171111001504-be1fbeda1936/go.mod h1:ZjcWmFBXmLKZu9Nxj3WKYEafiSqer2rnvPr0en9UNpI=
+github.com/vmware/go-vmware-nsxt v0.0.0-20190201205556-16aa0443042d h1:95uMoAryJvAwfDSxlgQEbT9xkNyWTB5YIxCweItbjGc=
+github.com/vmware/go-vmware-nsxt v0.0.0-20190201205556-16aa0443042d/go.mod h1:AzmozsuEImWQcg9Cfef9oMNoPQKcNaps6VkyJhzbSf0=
 github.com/vmware/govmomi v0.20.1/go.mod h1:URlwyTFZX72RmxtxuaFL2Uj3fD1JTvZdx59bHWk6aFU=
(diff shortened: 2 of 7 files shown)


mandelsoft:
Hi, we from the Gardener team had a workshop on friday with some *vmware* folks.
Padmaja Vrudhula asks us to review your pull request. We are currently developing the
*vmware* extension for the gardener (https://github.com/gardener/gardener). Therefore
we required load balancer support. We also have developed a load balancer controller
(https://github.com/MartinWeindel/nsxt-lb-provider). It is a regular cloud controller
manager only implementing the load balancer part and can run beside the
vsphere-cloud-controller-manager. It is not yet part of the gardener project,
because we wanted to finalize it first.

When looking at your pull request, we found several issues:
- it does not handle a change of the port list of a service
  it only adds virtual servers, but does not delete them for vanished ports
- it does not create health checks, this might be seen as superfluous, but for
  other providers this is done.
- there is no garbage collection. It does not cleanup orphaned elements
  (for example, if a service is forcefully removed, or if there are problems 
  with the controller or if it unexpectedly crashes or is aborted)
- everything is identified by generated names. I'm not sure that 5 characters from the
  service UUID are really unique. To find the elements in the UI it might also be useful to 
  additionally use the cluster name.
- it is only possible to use one IPPool for the load balancer IPs. But it might be
  useful to offer a cluster user the possibility to choose among multiple possibilities,
  (for example a public and a private load balancer) according to preconfigured
  sets with different visibility or routing settings. We have already added this 
  feature to the openstack cloud provider, which is also used for on-prem scenarios.
- the load balancer service is not managed
- our admin told us, that it might be helpful to add tags to the generated
  elements to be able to configure firewall rules in NSX-T.

We had implemented all this in our proposal, maybe we can combine both approaches
to get rid at all of our local project. It would be great if we could find a common solution.
If useful, we could adapt our approach to be added to the regular vSphere cloud controller
manager. (We originally decided to create an own controller manager, because it requires
NSX-T, while the vSphere cloud controller manager ist purely based on vSphere.)

andrewsykim:
Hi @mandelsoft! Sorry for the late reply, I was on paternity leave for the past few weeks! I agree it would be great if we can consolidate all implementations here so users only have to deploy 1 cloud controller manager. I think we should discuss this in the next vSphere subproject meeting. @frapposelli @dvonthenen @maplain can we add this to the next meeting agenda?

dvonthenen:
> Hi @mandelsoft! Sorry for the late reply, I was on paternity leave for the past few weeks! I agree it would be great if we can consolidate all implementations here so users only have to deploy 1 cloud controller manager. I think we should discuss this in the next vSphere subproject meeting. @frapposelli @dvonthenen @maplain can we add this to the next meeting agenda?

I added this to agenda for the Feb 5th meeting

mandelsoft:
Hi @andrewsykim , @dvonthenen, to not lose just another month, should we prepare a pull request until Feb, 5th?

andrewsykim:
@mandelsoft that sounds good, if you open a PR based on your work in https://github.com/MartinWeindel/nsxt-lb-provider we can close this one and collaborate there.

mandelsoft:
Hi @andrewsykim, I'm just working on the pull request and saw that we are still using an unmerged pull request vmware/go-vmware-nsxt#19 adding some missing rest calls for NSX-T required by our solution. May be you have the chance to trigger this internally.

andrewsykim:
@mandelsoft for now you can implement those API calls as internal functions (see `listLoadBalancerVirtualServers` for an example)

dvonthenen:
FYI

Different protocols in the same Service definition with type=LoadBalancer - https://github.com/kubernetes/enhancements/pull/1438

andrewsykim:
replaced by https://github.com/kubernetes/cloud-provider-vsphere/pull/292

PR #273, opened 2019-11-06
Fix failing go test in connection_test.go

dvonthenen:
**What this PR does / why we need it**:
Just like it sounds... it's a port of this fix https://github.com/kubernetes/kubernetes/pull/81390/files#diff-a0b58bcaae491877d401a6a7011882f3L159

**Which issue this PR fixes**: NA

**Special notes for your reviewer**:
NA

**Release note**:
NA

PR #274, opened 2019-11-07
Always ensure KIND is present for integration test

codenrhoden:
This (in conjunction with a test-infra PR) will fix the failing integration tests.

This patch adds a step in the integration test Makefile to explicitly
install KIND if it is not present. The default KIND version is now
v0.5.1, which requires changes to the kind config to use the v1alpha3
API.

Part of the movitivation behind this is to be able to run the
integration test in the kubekins-e2e image rather than a custom "CI"
image created from this project.

**Release note**:

```release-note
```
/hold
^ until https://github.com/kubernetes/test-infra/pull/15171 is merged

/assign @dvonthenen

### DIFF
--- a/test/integration/Makefile
+++ b/test/integration/Makefile
@@ -22,7 +22,10 @@ else
 export KUBECONFIG := $(KUBECONFIG_ORIG)
 endif
 
-cluster-up: $(KIND_CONFIG) | $(DOCKER_SOCK)
+kind:
+	./install_kind.sh
+
+cluster-up: $(KIND_CONFIG) kind | $(DOCKER_SOCK)
 	kind create cluster --config "$(KIND_CONFIG)" --name "$(CLUSTER_NAME)"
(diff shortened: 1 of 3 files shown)


dvonthenen:
/lgtm

Just waiting on https://github.com/kubernetes/test-infra/pull/15171 to get merged

codenrhoden:
Cool -- with the test-infra stuff merged, I can run tests manually now...
/test pull-cloud-provider-vsphere-build
/test pull-cloud-provider-vsphere-unit-test
/test pull-cloud-provider-vsphere-integration-test

codenrhoden:
/hold cancel

Woohoo! Integration test passes. Only test failure here is the unit test, and I *think* that's the issue @dvonthenen is trying to fix in #273. Since it's not a required test here, we should be good to go.

codenrhoden:
appears to be no way I can skip the failed test, even though it is *not* required. Once I add it, seems to be no way to remove it. Going to close and reopen just to see what happens.

/close

PR #278, opened 2019-11-25
Remove unused node maps and informers

andrewsykim:
Removes node informer which updates a node UUID map that is no longer used. 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -20,10 +20,8 @@ import (
 	"io"
 	"runtime"
 
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/klog"
-
 	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/klog"
 
 	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/server"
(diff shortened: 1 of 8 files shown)


andrewsykim:
/assign @dvonthenen 

Credit to @akutz for finding out we don't need `nodeRegUUIDMap`

dvonthenen:
Since @andrewsykim is going on paternity leave... will copy and paste the convo we had over slack.

`nodeRegUUIDMap` should be in-use and looks like you found a bug. the function here (https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/nodemanager.go#L331) should be checking against nodeRegUUIDMap to see if the node is still in the cluster. currently, the function is reporting all nodes throughout time and not the active ones.`nodeRegUUIDMap` is probably a horrible name for the map because it’s supposed to represent nodes that are currently registered to the cluster.

we use the listener for on add informer to wipe out stale info about existing VMs (ie VMs that were previously registered but that have come back online/re-regsitered). delete is used to remove VMs vrom the `nodeRegUUIDMap` map to see which ones are currently active VMs.

dvonthenen:
Opened PR https://github.com/kubernetes/cloud-provider-vsphere/pull/282 to address the bug. Placing a hold on this PR so this doesn't get accidentally merged without following up with @andrewsykim

/hold

PR #289, opened 2020-01-10
Added fstype to default storageclass

shraderdm:
In testing, found that no default fstype is used unless specified (CentOS 7, kernel 3.10.0-1062.9.1.el7.x86_64, k8s 1.17.0). Added fstype in storageclass to avoid this issue. Can also specify other fstype such as xfs, etc.

**What this PR does / why we need it**: Adds file system type to the default storageclass to allow containers to consume the persistent volumes. Needed for certain scenarios where a filesystem type will be left blank when unspecified. 

This is my first PR, so I apologize if I've missed anything. Please let me know if I need to add more info or update my comments.

```release-note None
```

### DIFF
--- a/docs/book/tutorials/kubernetes-on-vsphere-with-kubeadm.md
+++ b/docs/book/tutorials/kubernetes-on-vsphere-with-kubeadm.md
@@ -1001,6 +1001,7 @@ metadata:
 provisioner: csi.vsphere.vmware.com
 parameters:
   storagepolicyname: "Space-Efficient"
+  fstype: ext4
 ```
 
 ```bash


shraderdm:
After additional testing, ext4 _is_ used when no fstype is specified. However, the fstype in both the storageclass and on each PV created from it are listed as blank. May still be nice to have this line in the doc to allow users to choose a different fstype for their default storageclass without having to dig to figure out how, but containers can consume the storage without this change.

dvonthenen:
Hi @shraderdm. The default is ext4 when the fstype is not specified. It could be helpful for others to know that the fstype can be changed to some user-defined value.

/lgtm
/approve

PR #292, opened 2020-01-27
Support for NSX-T based Load Balancers

mandelsoft:
This PR adds support for the NSX-T loadbalancers to the vsphere cloud-controller-manager.
It must explicitly be enabled by additional configuration options in the configuration file.
- If enabled, it manages the load balancer service and virtual servers for TCP/UDP kubernetes services, including health checks.
- It also includes a cleanup feature for deleting/releasing obsolete NSX-T elements of formerly available services in the case of a crash recovery or services deleted without a running controller manager.
- By configuring dedicated load balancer classes, it is possible to refer by annotations to various IP Pools for the allocated IP of the virtual server. 

**Release note**:

```release-note
```


cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -26,6 +26,8 @@ import (
 	"os"
 	"time"
 
+	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/loadbalancer"
```
andrewsykim: Group this with other imports starting with k8s.io/cloud-provider-vspher

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -20,6 +20,8 @@ import (
 	"io"
 	"runtime"
 
+	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/loadbalancer"
```
andrewsykim: Group this import with other imports starting with `k8s.io/cloud-provider-vsphere`

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
```diff
@@ -0,0 +1,210 @@
...
+	nsxt "github.com/vmware/go-vmware-nsxt"
+
+	corev1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	cloudprovider "k8s.io/cloud-provider"
+)
+
+const (
+	// AnnotLoadBalancerClass is the optional class annotation at the service
+	AnnotLoadBalancerClass = "loadbalancer.vmware.io/class"
```
andrewsykim: nit: LoadBalancerClassAnnotation

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
```diff
@@ -0,0 +1,210 @@
...
+	return clusterName + ":" + service.Namespace + ":" + service.Name
+}
+
+// EnsureLoadBalancer creates a new load balancer 'name', or updates the existing one. Returns the status of the balancer
+// Implementations must treat the *corev1.Service and *corev1.Node
+// parameters as read-only and not modify them.
+// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+func (p *lbProvider) EnsureLoadBalancer(ctx context.Context, clusterName string, service *corev1.Service, nodes []*corev1.Node) (*corev1.LoadBalancerStatus, error) {
+	key := objectNameFromService(service).String()
+	p.keyLock.Lock(key)
```
andrewsykim: Curious why the key-based lock. It seems like keyLock.Lock already holds a `sync.Mutex`, can we simplify and just lock `sync.Mutex` here?

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
```diff
@@ -0,0 +1,210 @@
...
+	defer p.keyLock.Unlock(key)
+
+	class, err := p.classFromService(service)
+	if err != nil {
+		return nil, err
+	}
+
+	state := newState(p.lbService, clusterName, service, nodes)
+	err = state.Process(class)
+	status, err2 := state.Finish()
```
maplain: nit: this reads a bit counter-intuitive. from what I can tell, this is because `state.Finish` is trying to do two things depending on whether `len(service.spec.Ports) == 0`: cleanup or return a valid status(because `EnsureLoadBalancerDeleted` is leveraging `EnsureLoadBalancer`). Based on the comment and function name, it seems more natural to move the latter part to `Process`, so we can do:
```	
if err != nil {
		return status, err
}
```
right after `status, err = state.Process(class)`

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
+	github.com/vmware/go-vmware-nsxt v0.0.0-20190201205556-16aa0443042d // master
```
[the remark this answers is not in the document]
andrewsykim: @maplain do you happen to have reference usage of that client for NSX-T?

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
```diff
@@ -0,0 +1,297 @@
...
+			msg := "no default load balancer class defined"
+			klog.Errorf(msg)
+			return fmt.Errorf(msg)
+		} else if class.IPPoolName == "" && class.IPPoolID == "" {
+			msg := "default load balancer class: ipPoolName and ipPoolID is empty"
+			klog.Errorf(msg)
+			return fmt.Errorf(msg)
+		}
+	} else {
+		if cfg.LoadBalancer.IPPoolName == "" && cfg.LoadBalancer.IPPoolID == "" {
```
maplain: would this `if` statement ever be hit? considering it's within the `else` after the check on `if cfg.LoadBalancer.IPPoolID == "" && cfg.LoadBalancer.IPPoolName == ""` above.

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
```diff
@@ -0,0 +1,210 @@
...
+
+// ClusterName contains the cluster name injected from main
+// TODO cluster name needed for reorg and is currently injected from main
+var ClusterName string
+
+var _ cloudprovider.LoadBalancer = &lbProvider{}
+
+// NewLBProvider creates a load balancer implementation for the given config
+func NewLBProvider(cfg *config.LBConfig) (LoadBalancer, error) {
+	if !cfg.IsEnabled() {
```
maplain: `cfg.IsEnabled` is checking `len(cfg.LoadBalancerClasses) > 0 || !cfg.LoadBalancer.IsEmpty()`, which seems like a configuration validation. Should this be captured in `cfg.LBConfig.CompleteAndValidate()`?

pkg/cloudprovider/vsphere/loadbalancer/interface.go
```diff
@@ -0,0 +1,110 @@
...
+// LoadBalancer is the interface used call the load balancer functionality
+// It extends the cloud controller manager LoadBalancer interface by an
+// initialization function
+type LoadBalancer interface {
+	cloudprovider.LoadBalancer
+	Initialize(client clientset.Interface, stop <-chan struct{})
+}
+
+// Access provides methods for dealing with NSX-T objects
+type Access interface {
```
maplain: this is a lot like NsxtBroker interface. is there any significant benefit from this extra layer of abstraction?

pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
```diff
@@ -0,0 +1,219 @@
...
+	client *nsxt.APIClient
+}
+
+// NewNsxtBroker creates a new NsxtBroker to the real API
+func NewNsxtBroker(client *nsxt.APIClient) NsxtBroker {
+	return &nsxtBroker{client: client}
+}
+
+func (b *nsxtBroker) ListIPPools() (manager.IpPoolListResult, error) {
+	result, _, err := b.client.PoolManagementApi.ListIpPools(b.client.Context, nil)
```
maplain: seems like this is still on Manager API, which is being deprecated. might be better if we switch to the Policy API.

go.mod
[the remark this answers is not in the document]
maplain: for `https://github.com/vmware/vsphere-automation-sdk-go/tree/master/services/nsxt`, not that I'm aware of. but Ent PKS was using an auto-generated Policy API client: https://gitlab.eng.vmware.com/PKS/pks-networking/tree/master/pkg/client/nsx/policy in https://gitlab.eng.vmware.com/PKS/pks-nsx-t-release/tree/policy_api/src/pkg/networkmanager/policy(though the migration from Manager API to Policy API never happened).

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
mandelsoft: The Mutex in `keyLock` is used to lock the lock operation for a dedicated key. It is released again in the `keyLock.Lock` method. This locking is folded with the locking of a dedicated Mutex created for the given key to secure the map operation to hold the Mutex for the key.

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
mandelsoft: Hmm, I think the problems is, that the `Finish` stuff should be done for all returns in `Process`. This could be done with a `defer` here, but this cannot return a new error if it fails. Therefore we separated this into a dedicated method, that can be called with this mixed error code handling.

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
```diff
@@ -0,0 +1,297 @@
...
+	// DefaultRetryMaxDelay is teh default value for maximum retry delay
+	DefaultRetryMaxDelay = 5000
+
+	// DefaultLoadBalancerClass is the default load balancer class
+	DefaultLoadBalancerClass = "default"
+)
+
+// SizeToMaxVirtualServers contains the mapping from size to maximum number of virtual servers
+var SizeToMaxVirtualServers = map[string]int{
+	SizeSmall:  20,
```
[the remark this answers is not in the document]
mandelsoft: Good question, we didn't know that these constants are release specific. Any proposal? Is there a way to query the actual version, or even better the effective values for the sizes?

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+	} else {
+		if cfg.LoadBalancer.IPPoolName == "" && cfg.LoadBalancer.IPPoolID == "" {
+			msg := "load balancer ipPoolName and ipPoolID is empty"
+			klog.Errorf(msg)
+			return fmt.Errorf(msg)
+		}
+	}
+	if cfg.NSXTSimulation == nil {
+		return cfg.NSXT.validateConfig()
```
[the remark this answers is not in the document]
mandelsoft: Hmm, the idea was to be able to call the validation function without a prior call of the `FromEnv` function. Because all this is done only once during startup it shouldn't be a critical issue.

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
mandelsoft: yepp, the check should better test whether both is given. I've fixed it.

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
mandelsoft: If only one of these cases is given or both, the config is valid. The feature is enabled if at least one such entries are given. This is checked and hidden by this method.

pkg/cloudprovider/vsphere/loadbalancer/interface.go
mandelsoft: This was intended to have a hook for testing without real mocking. We had no time so far to continue with this.

pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
mandelsoft: Do you have a code snipped with a better coding?

go.mod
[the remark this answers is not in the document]
mandelsoft: Unfortunately I cannot resolve the dns name `gitlab.eng.vmware.com`. We've had a look into this new repo, but to be honest, we couldn't figure out how to use it. Do you have some understandable docu?

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
```diff
@@ -0,0 +1,297 @@
...
+	err := cfg.CompleteAndValidate()
+	if err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+// CompleteAndValidate validates the configuration after filling in defaults
+func (cfg *LBConfig) CompleteAndValidate() error {
+	if cfg.IsEnabled() {
```
andrewsykim:
```g
if !cfg.IsEnabled() {
    return nil
}
...
...
...
```

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func TestReadConfig(t *testing.T) {
+	s1 := `
+[LoadBalancer]
+ipPoolName = pool1
+size = MEDIUM
+lbServiceId = 4711
+logicalRouterId = 1234
+
+[LoadBalancerClass "public"]
```
andrewsykim: The key name has to be "public" or "private" here? What's the reason we can't make it arbitrary?

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+ipPoolName = poolPublic
+
+[LoadBalancerClass "private"]
+ipPoolName = poolPrivate
+
+[Tags]
```
andrewsykim: This name is too generic if it's for LB tagging only.

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		if cfg.NSXT.RetryMaxDelay == 0 {
+			cfg.NSXT.RetryMaxDelay = DefaultRetryMaxDelay
+		}
+		if cfg.LoadBalancerClasses == nil {
+			cfg.LoadBalancerClasses = map[string]*LoadBalancerClassConfig{}
+		}
+		for _, class := range cfg.LoadBalancerClasses {
+			if class.IPPoolName == "" {
+				if class.IPPoolID == "" {
```
andrewsykim:
```go
if class.IPPoolName == "" && class.IPPoolID == "" {
   ...
}
```

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
```diff
@@ -0,0 +1,69 @@
...
+package loadbalancer
+
+import (
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ObjectName is a namespace/name pair
+type ObjectName struct {
```
andrewsykim: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
andrewsykim: What benefit do we get from this instead of using sync.Mutex and locking against all keys? Is it for performance reasons so we can operate against multiple keys concurrently or something else?

pkg/cloudprovider/vsphere/loadbalancer/interface.go
andrewsykim: Can we give this a more specific name at least? `NSXTClient`?

pkg/cloudprovider/vsphere/loadbalancer/reorg.go
```diff
@@ -0,0 +1,127 @@
...
+	"k8s.io/klog"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const maxPeriod = 30 * time.Minute
+
+func (p *lbProvider) reorg(client clientcorev1.ServiceInterface, stop <-chan struct{}) {
```
andrewsykim: Can you add some comments for what this is doing?

mandelsoft: done

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
mandelsoft: done

pkg/cloudprovider/vsphere/loadbalancer/reorg.go
mandelsoft: done

mandelsoft:
@andrewsykim the dedicated lock per service might be useful to avoid the global reorg go routine to block all service reconcilations

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
mandelsoft: That's just a test to check whether a given config in a config file  correctly reaches the internal data structures. So we have to use a dedicated (artificial) value.

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
mandelsoft: May be those tags can be used for all vSphere/NSXT-T elements that are generated by the controller manager to be able to identify those elements later o (or by a human being using the GUI). So far, only the LB controller creates such elements. So  it is the only controller consuming this tags section. But nevertheless it should have a more common name, because of it's intended general usage.

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
mandelsoft: done

pkg/cloudprovider/vsphere/loadbalancer/interface.go
mandelsoft: renamed

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
mandelsoft: The dedicated lock per service might be useful to avoid the global reorg go routine to block all service reconcilations

go.mod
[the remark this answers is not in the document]
mandelsoft: We found some example and are playing with it, but we still found a severe problem. The elements already created with the *old* API are not visible anymore with the new one.
Therefore is it not possible to setup infrastructure with the old one and switch the controller version to a new controller without deleting the complete cluster and recreating it.
This behavior is basically not acceptable, because we urgently need a valid migration path if VMware evolves its APIs, Assume AWS changes its REST API and suddenly no existing VM is visible anymore. That's definitely a NO-GO.

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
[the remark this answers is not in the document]
mandelsoft: Because there is only one LB Service per router possible, we don't need the sizes anymore. I've remove this part of the coding.

go.mod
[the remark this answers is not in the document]
maplain: sorry, I lost track of this thread.
May I ask it's the new Policy API or the new Manager API that's lacking some objects? It's ok to go with either one, I brought `https://github.com/vmware/vsphere-automation-sdk-go/tree/master/services/nsxt` up because I'm worried `github.com/vmware/go-vmware-nsxt` will not be maintained anymore.
@andrewsykim do we know if PM has any requirement on the NSX-T version we need to support in CAPV?

go.mod
```diff
@@ -32,11 +36,14 @@ require (
 	k8s.io/component-base v0.0.0
 	k8s.io/klog v0.3.2
 	k8s.io/kube-openapi v0.0.0-20190401085232-94e1e7b7574c // indirect
+	k8s.io/kubelet v1.15.0
 	k8s.io/kubernetes v1.15.0
 	k8s.io/sample-controller v0.0.0-20190731144349-6f8905ae4ee5
 )
 
 replace (
+	github.com/vmware/go-vmware-nsxt => github.com/MartinWeindel/go-vmware-nsxt v0.0.0-20191129144634-29aa700ba036
```
andrewsykim: what's the reason for this? was the SDK missing something?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -20,11 +20,13 @@ import (
 	"io"
 	"runtime"
 
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/log"
```
andrewsykim: nit: group this import separately

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,111 @@
+# NSX-T Load Balancer Controller
+
+*Kubernetes load balancer support using NSX-T for the `vSphere` cloud controller
+manager.*
+
+This package enriches the cloud provider interface by implementing the load
+balancing API of the cloud controller for an NSX-T environment.
+
+There are two different modes:
```
andrewsykim: Can we imply the different modes based on whether a loadbalancer service ID is provided or not in the config?

andrewsykim: Ah I see from the reading below this is the case right?

andrewsykim: For the first iteration I would actually prefer to always default to the "unmanaged mode" -- in other words already require a service ID from a user. As a future feature we can create the LB service automatically.

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,111 @@
...
+configuration file:
+
+```ini
+[LoadBalancer]
+ipPoolName = pool1
+#ipPoolId = 0815
+lbServiceId = 4711
+size = MEDIUM
+
+[LoadBalancerClass "public"]
```
andrewsykim: Can we just make this a field in the LoadBalancer config above? So `type=public` or `class=public`?

pkg/cloudprovider/vsphere/loadbalancer/access.go
```diff
@@ -0,0 +1,493 @@
...
+*/
+
+package loadbalancer
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
```
andrewsykim: nit: group imports starting with `github.com` and `k8s.io` separately

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
```diff
@@ -0,0 +1,118 @@
...
+	assertEquals("LoadBalancerClass.private.udpAppProfileName", config.LoadBalancerClasses["private"].UDPAppProfileName, "udp2")
+	if len(config.AdditionalTags) != 2 || config.AdditionalTags["tag1"] != "value1" || config.AdditionalTags["tag2"] != "value2" {
+		t.Errorf("unexpected additionalTags %v", config.AdditionalTags)
+	}
+	assertEquals("NSX-T.user", config.NSXT.User, "admin")
+	assertEquals("NSX-T.password", config.NSXT.Password, "secret")
+	assertEquals("NSX-T.host", config.NSXT.Host, "nsxt-server")
+}
+
+func TestReadConfig2(t *testing.T) {
```
andrewsykim: `TestReadConfigOnVMC`

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+func namespacedNameFromService(service *corev1.Service) types.NamespacedName {
+	return types.NamespacedName{Namespace: service.Namespace, Name: service.Name}
+}
+
+func parseNamespacedName(name string) types.NamespacedName {
+	parts := strings.Split(name, "/")
+	return types.NamespacedName{Namespace: parts[0], Name: parts[1]}
+}
+
+func stringsEquals(a []string, b []string) bool {
```
andrewsykim: use reflect.DeepEqual instead?

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	return set
+}
+
+func strptr(s string) *string {
+	return &s
+}
+
+func isNotFoundError(err error) bool {
+	if _, ok := err.(vapi_errors.NotFound); ok {
```
andrewsykim:
```go
_, ok := err.(vapi_errors.NotFound)
return ok
```

pkg/cloudprovider/vsphere/vapilogger.go
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+package vsphere
+
+import (
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/log"
+	"k8s.io/klog"
+)
+
+type klogBridge struct {
```
andrewsykim: Can you comment on why this is needed?

go.mod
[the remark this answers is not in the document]
mandelsoft: We've migrated now to the new intent based (automation API). See additional commit from @MartinWeindel

pkg/cloudprovider/vsphere/vapilogger.go
mandelsoft: The new GO SDK uses an own logger and logs by default a lot of stuff to the standard file descriptors. Therefore Martin has added a bridge to the logger used in the kubernetes cloud provider manager.

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: These are arbitrary class names, you can have as many as you want with different settings. The name `public` here is just an example.

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: The managed mode is important for us, because we want to do as less as possible for the automated infrastructure setup on our own.

Basically the mode must not explicitly be selected. So, if the service is specified, it is used. Therefore the standard usecase could be to specify the service and it works as expected.

We could mention the two modes less prominent in the docu and all examples just specify the service.

Would this be ok for you?

andrewsykim: > We could mention the two modes less prominent in the docu and all examples just specify the service.

Sounds good to me.

pkg/cloudprovider/vsphere/loadbalancer/README.md
andrewsykim: I see, so where in the config do we point to a specific class. Below I see that `LoadBalancerClassConfig` is embedded in `LoadBalancerConfig` already.

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,111 @@
...
+There are two different modes:
+
+- the *managed* mode manages the load balancer service. The NSX-T load balancer
+  service is only created if it is required.
+  This saves resources if no kubernetes service of type `LoadBalancer` is
+  actually used.
+- the *unmanaged* mode is used if the configuration specifies a load balancer
+  service id. Here only the virtual servers are managed.
+
+The basic assumption is that all nodes are bound to a logical tier1 router.
```
maplain: does this mean in `managed` mode, the Logical Router ID will be inferred by looking at the router one of the nodes is connected to?
do we have validation on this assumption `all nodes are bound to a logical tier1 router`?

andrewsykim: I think logical router ID is required in the config

andrewsykim: not sure about nodes being bound to tier 1 router though

go.mod
MartinWeindel: good catch, this is not used anymore

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: Please correct me, if I'm wrong, but as far as we know there can be only one LBService per logical router/tier1 gateway. The same way there can be only one router configured for an LBService. 
This would mean to manage multiple VIPs and LBServices and Virtual servers dependening on the router used for a node, if the nodes are connected to different routers.
I ' don't know whether this is a useful scenario. It could be useful for HA purposes, but as far as I understood from some workshop with VMware, the edge clusters provide an HA setup for the NSX-T elements. So, there should be no need to add nodes to multiple segments handled by different routers.

This would significantly complicate the complete LB handling and enforce to use some DNS based load balancing. I think we don't need this scenario.

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: The class is selected via annotation on the kubernetes service object of type load balancer.
This is similar to the OpenStack solution. 
The config contains a default, this is part of the flat section. It is used if no class is specified by the service. Then there might be any number of dedicated class specifications in the config file (each in a separate group). The values here are again defaulted by values from the global section. 
So if you just want to use different IPPools and share the same TCP/UDP profiles. The profiles are set in the global section and the IPPools in the dedicated group sections.

andrewsykim: Okay I think I'm understanding this better now, thanks for explaining :)

go.mod
```diff
@@ -32,11 +36,13 @@ require (
 	k8s.io/component-base v0.0.0
 	k8s.io/klog v0.3.2
 	k8s.io/kube-openapi v0.0.0-20190401085232-94e1e7b7574c // indirect
+	k8s.io/kubelet v1.15.0
 	k8s.io/kubernetes v1.15.0
 	k8s.io/sample-controller v0.0.0-20190731144349-6f8905ae4ee5
 )
 
 replace (
+	gopkg.in/gcfg.v1 => github.com/mandelsoft/gcfg v1.2.4-0.20191118133708-71f203a139d3
```
andrewsykim: Is this replace needed still?

go.sum
```diff
@@ -12,6 +12,8 @@ github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym
 github.com/GoogleCloudPlatform/k8s-cloud-provider v0.0.0-20181220005116-f8e995905100/go.mod h1:iroGtC8B3tQiqtds1l+mgk/BBOrxbqjH+eUfFQYRc14=
 github.com/JeffAshton/win_pdh v0.0.0-20161109143554-76bb4ee9f0ab/go.mod h1:3VYc5hodBMJ5+l/7J4xAyMeuM2PNuepvHlGs8yilUCA=
 github.com/MakeNowJust/heredoc v0.0.0-20170808103936-bb23615498cd/go.mod h1:64YHyfSL2R96J44Nlwm39UHepQbyR5q10x7iYa1ks2E=
+github.com/MartinWeindel/go-vmware-nsxt v0.0.0-20191129144634-29aa700ba036 h1:BpnIkS8rhYkqE0pzzh8xp0CeUkateo9mJIKbxpCPGGA=
```
andrewsykim: Still seeing references to your fork here

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,120 @@
...
+This load balancer controller supports the usage of multiple load balancer
+classes. Classes are preconfigured in the configuration file of the cloud
+controller manager. Every class may use another `IPPool` configured in NSX-T.
+This supports the creation of load balancers in different visibility realms,
+for example an `*internet facing* or a *private* load balancer. The
+IPPools must be preconfigured in NSX-T. To select a dedicated loadbalancer
+class the kunernetes service object must be annotated with the annotation:
+
+```yaml
+loadbalancer.vsphere.class: <class name>
```
andrewsykim: This is not consistent with the code, this should be `loadbalancer.vmware.io/class: <class name>`

pkg/cloudprovider/vsphere/loadbalancer/interface.go
```diff
@@ -0,0 +1,119 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package loadbalancer
+
+import (
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
```
andrewsykim: nit: in general can we keep imports grouped by the org name, i.e. `k8s.io` and `github.com` should be grouped separpately

go.mod
mandelsoft: There were several problems with the gcfg API. For example it does not support a generic map as section used to read the arbitrary tag list. I've provided a pull request go-gcfg/gcfg#18 for this project but nobody is reaction. 

If the controller manager switches as planned to a json config file format, we would get rid of this problem at all.

May be it is acceptable to use this fork version in between.

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/helpers.go
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/access.go
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: done

pkg/cloudprovider/vsphere/cloud.go
mandelsoft: fixed

pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
mandelsoft: We've switched to the new SDK

pkg/cloudprovider/vsphere/loadbalancer/config/config.go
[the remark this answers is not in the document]
mandelsoft: rearranged

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -71,8 +75,12 @@ func main() {
 
 	//
 	var versionFlag *pflag.Value
+	var clusterNameFlag *pflag.Value
 	pflag.CommandLine.VisitAll(func(flag *pflag.Flag) {
-		if flag.Name == "version" {
+		switch flag.Name {
+		case "cluster-name":
```
andrewsykim: nit: I actually prefer this to be `cluster-id` to match with the cluster ID flag in the CSI driver

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,145 @@
...
+To select a dedicated load balancer class different from the default one, the
+Kubernetes service object must be annotated with the annotation:
+
+```yaml
+loadbalancer.vmware.io/class: <class name>
+```
+
+If no such annotation is given the default class will be used. This gives
+the adminstrator of the cluster a chance to restrict the usage of the
+NSXT-T resources for cluster users. He can determine which elements should
```
andrewsykim: s/He/They/

go.mod
andrewsykim: I see, and this is needed for the tags feature?

pkg/cloudprovider/vsphere/loadbalancer/README.md
```diff
@@ -0,0 +1,145 @@
...
+## Configuration File
+
+The controller manager requires dedicated entries in the cloud controller's
+configuration file:
+
+```ini
+[LoadBalancer]
+ipPoolName = pool1
+#ipPoolId = 0815
+#tier1GatewayPath = /infra/tier-1s/12345
```
andrewsykim: Can you get rid of the commented fields and create a separate example actually using those fields?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -89,11 +91,17 @@ func (vs *VSphere) Initialize(clientBuilder cloudprovider.ControllerClientBuilde
...
 	}
+	if vs.loadbalancer != nil {
+		vs.loadbalancer.Initialize(client, stop)
+	}
 }
 
 // LoadBalancer returns a balancer interface. Also returns true if the
 // interface is supported, false otherwise.
 func (vs *VSphere) LoadBalancer() (cloudprovider.LoadBalancer, bool) {
+	if vs.loadbalancer != nil {
```
andrewsykim: Can we also gate this feature with an env var like `ENABLE_ALPHA_NSXT_LB`. I want to make it really clear to users that this is an alpha feature.

pkg/cloudprovider/vsphere/loadbalancer/access.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	result, err := a.broker.CreateLoadBalancerService(lbService)
+	if err != nil {
+		return nil, errors.Wrapf(err, "creating load balancer service failed for cluster %s", clusterName)
+	}
+	return &result, nil
+}
+
+func (a *access) FindLoadBalancerService(clusterName string, id string) (*model.LBService, error) {
+	if id != "" {
```
andrewsykim: nit: this can be a bit easier to follow with an early return
```go
if id == "" {
   return a.findLoadBalancerService(a.ownerTag, clusterTag(clusterName)) 
}
...
```

pkg/cloudprovider/vsphere/loadbalancer/access.go
```diff
@@ -0,0 +1,494 @@
...
+	return a.findAppProfilePathByName(profileReference.Name, resourceType)
+}
+
+func (a *access) CreateVirtualServer(clusterName string, objectName types.NamespacedName, class LBClass, ipAddress string,
+	mapping Mapping, lbServicePath, applicationProfilePath string, poolPath *string) (*model.LBVirtualServer, error) {
+	allTags := append(class.Tags(), clusterTag(clusterName), serviceTag(objectName), portTag(mapping))
+	virtualServer := model.LBVirtualServer{
+		Description: strptr(fmt.Sprintf("virtual server for cluster %s, service %s created by %s",
+			clusterName, objectName, AppName)),
+		DisplayName:            strptr(fmt.Sprintf("cluster:%s:%s", clusterName, objectName)),
```
andrewsykim: The display name format should be a constant or generated from one common function.

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
```diff
@@ -0,0 +1,182 @@
...
+// exists, returning nil if the load balancer specified either didn't exist or
+// was successfully deleted.
+// This construction is useful because many cloud providers' load balancers
+// have multiple underlying components, meaning a Get could say that the LB
+// doesn't exist even if some part of it is still laying around.
+// Implementations must treat the *corev1.Service parameter as read-only and not modify it.
+// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
+func (p *lbProvider) EnsureLoadBalancerDeleted(ctx context.Context, clusterName string, service *corev1.Service) error {
+	emptyService := *service
+	emptyService.Spec.Ports = nil
```
andrewsykim: You shouldn't touch fields that come in from the cache. If you need to nil ports, do a `.DeepCopy()` on the service object.

Also wondering why we need to nil the ports field here?

pkg/cloudprovider/vsphere/loadbalancer/reorg.go
```diff
@@ -0,0 +1,168 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog"
+)
+
+const maxPeriod = 30 * time.Minute
+
+// reorg is used to cleanup obsolete and potentially forgotten objects
```
andrewsykim: Wondering if `cleanup` is a better name here. `reorg` to me implies the resource is changing ownership

pkg/cloudprovider/vsphere/loadbalancer/state.go
```diff
@@ -0,0 +1,457 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog"
+
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
+
+	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/loadbalancer/config"
+)
+
+var verboseLevel = klog.Level(2)
```
andrewsykim: We shouldn't need to do this. The cloud-controller-manager is already setup to accept the verbose level via a flag.

pkg/cloudprovider/vsphere/loadbalancer/access.go
andrewsykim: Seems like we can use `GetLoadBalancerName()`

pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go
MartinWeindel: Good catch, thanks! This was an ugly bug.
The called EnsureLoadBalancer can create and delete virtual servers. It acts on the delta to the existing state. Therefore removing all ports translates into removing all virtual servers (there is a virtual server for each service port). Additionally, in the managed mode the load balancer service is deleted after the last virtual server has been deleted.

pkg/cloudprovider/vsphere/loadbalancer/state.go
MartinWeindel: This only sets the verbose level used for the logging in the state methods. It does not set the global verbose level.

andrewsykim: Ah I see, thanks for clarifying

cmd/vsphere-cloud-controller-manager/main.go
MartinWeindel: this is a predefined flag from Kubernetes and cannot be changed.
see https://github.com/kubernetes/kubernetes/blob/2652e176e0a24ed9a5d90cc9b3d670c0d4ed8720/cmd/controller-manager/app/options/kubecloudshared.go#L60

andrewsykim: I prefer we don't use  this flag, it has some implications that aren't so obvious (the flag description is already telling). 

I think what would be best is adding a `cluster-id` field to the global config https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/types.go#L21 and then using that value for "cluster name" for LBs.

go.mod
```diff
@@ -37,6 +41,7 @@ require (
...
```
andrewsykim: Still not a fan of using a personal forked repo here. If it's absolutely necessary can we copy it over to `third_party/forked/gcfg`? See https://github.com/kubernetes/kubernetes/tree/master/third_party/forked for an example.

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -89,11 +92,26 @@ func (vs *VSphere) Initialize(clientBuilder cloudprovider.ControllerClientBuilde
...
+		klog.Info("initializing load balancer support")
+		if loadbalancer.ClusterID == "" {
+			klog.Warning("Missing cluster id, no periodical cleanup possible")
+		}
+		vs.loadbalancer.Initialize(loadbalancer.ClusterID, client, stop)
+	}
+}
+
+func (vs *VSphere) isLoadBalancerSupportEnabled() bool {
+	_, ok := os.LookupEnv("ENABLE_ALPHA_NSXT_LB")
```
andrewsykim: I think it would be cleaner if the gate check is where we initialize the LB instance in buildVSphereFromConfig and keep `vs.loadbalancer` as nil if the gate doesn't allow it.

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -145,13 +163,24 @@ func (vs *VSphere) HasClusterID() bool {
 // Initializes vSphere from vSphere CloudProvider Configuration
 func buildVSphereFromConfig(cfg *CPIConfig) (*VSphere, error) {
 	nm := newNodeManager(cfg, nil)
-
+	lb, err := loadbalancer.NewLBProvider(&cfg.LBConfig)
```
andrewsykim: Put the gate check here and set lb to nil if it doesn't allow

go.mod
dvonthenen: I agree with this. Really not a fan of this as well.

pkg/cloudprovider/vsphere/loadbalancer/README.md
maplain: sorry for my late response.
@mandelsoft
No, even one node is connected with only one router, how could we ensure all nodes connect to the same router? also could you please clarify if the logical router ID is required in config or not?

maplain: nvm.  I'm ok with this as the starting point.
```
- the *managed* mode manages the load balancer service, also. Here the tier1
  gateway must be specified used, which is used for the segments the cluster
  nodes are connected to. The NSX-T load balancer service is only created if it
  is required. This saves resources if no kubernetes service of type
  `LoadBalancer` is actually used.
```

cmd/vsphere-cloud-controller-manager/main.go
MartinWeindel: introduced new flag cluster-id

pkg/cloudprovider/vsphere/cloud.go
mandelsoft: done

go.mod
mandelsoft: we replaced the tags section by a tag attribute containing a JSON map.

pkg/cloudprovider/vsphere/loadbalancer/reorg.go
mandelsoft: renamed

pkg/cloudprovider/vsphere/loadbalancer/access.go
mandelsoft: Introduced some display name functions in a central place for various combinations.
the `access` structure should not know anything about the `lbProvider` struct. Therefore we cannot use a method from this unknown structure. Instead we just use a common generic function.

pkg/cloudprovider/vsphere/loadbalancer/access.go
mandelsoft: done

pkg/cloudprovider/vsphere/loadbalancer/README.md
mandelsoft: Instead I've added a reference section listing all possible attributes with their meaning and combinations.

cmd/vsphere-cloud-controller-manager/main.go
mandelsoft: The reference docu for the cloud-controller maneger (https://kubernetes.io/docs/reference/command-line-tools-reference/cloud-controller-manager/) explains the option `cluster-name`with 
> --cluster-name string     Default: "kubernetes"
>   The instance prefix for the cluster.
which is exactly the use case we want to use. Therefore it really seems to to be the right choice. It also already passed by the load balancer interface of the cloud controller manager. That's another clear indicator that we are using it as intended.

go.mod
mandelsoft: we replaced it by a `tags` attribute using a JSON map as string value

pkg/cloudprovider/vsphere/loadbalancer/config/config_test.go
mandelsoft: We've moved it to the `LoadBalancer` section now

andrewsykim:
also if you can squash commits that'd be great!

andrewsykim:
Can we update this PR to be a normal PR and not a draft?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -145,13 +163,34 @@ func (vs *VSphere) HasClusterID() bool {
...
+		}
+	} else {
+		if lb == nil {
+			return nil, fmt.Errorf("To enable NSX-T load balancer support you need to configure section LoadBalancer")
+		}
+	}
+	if lb == nil {
+		klog.Infof("NSX-T load balancer support disabled")
+	} else {
+		klog.Infof("NSX-T load balancer support enabled")
```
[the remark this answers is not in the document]
dvonthenen: I second this as well

dvonthenen:
I left a comment as well, but looks good to me as an initial commit! Nice job!

pkg/cloudprovider/vsphere/cloud.go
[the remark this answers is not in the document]
MartinWeindel: Changed logging to `klog.Infof("NSX-T load balancer support enabled. This feature is alpha, use in production at your own risk.")`. Also added the sentence "since this is an alpha feature... can change." to the README.md.

andrewsykim:
This will mege automatically once it's no longer a draft

Issue #295, opened 2020-02-05
Updates to CPI Helm Chart


dvonthenen:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
Issue reported in chart repo: https://github.com/helm/charts/issues/19807
Creating this tracking issue so we don't lose sight of the work

**What you expected to happen**:
Make updates where necessary

**How to reproduce it (as minimally and precisely as possible)**:
NA

**Anything else we need to know?**:
NA

dvonthenen:
PR submitted:
https://github.com/helm/charts/pull/20612

dvonthenen:
PR merged closing this tracking issue

/close

PR #297, opened 2020-02-07
Update CPI image to use non-root account

dvonthenen:
Forces the CPI to run on a non-root account. In the distroless image, there is an account called `nobody` which will now be utilized. This will make the container image more secure!

Helps address https://github.com/kubernetes/cloud-provider-vsphere/issues/295

Tested on 6.7u3 and made sure all nodes initialized properly.

**Release note**:
NA

### DIFF
--- a/cluster/images/controller-manager/Dockerfile
+++ b/cluster/images/controller-manager/Dockerfile
@@ -52,5 +52,8 @@ RUN go build -a -ldflags='-w -s -extldflags=static -X main.version=${VERSION}' -
 # Copy the manager into the distroless image.
 FROM ${DISTROLESS_IMAGE}
 LABEL maintainer="Travis Rhoden <trhoden@vmware.com>"
+
+USER nobody
+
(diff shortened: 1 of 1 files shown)


cluster/images/controller-manager/Dockerfile
```diff
@@ -52,5 +52,8 @@ RUN go build -a -ldflags='-w -s -extldflags=static -X main.version=${VERSION}' -
 # Copy the manager into the distroless image.
 FROM ${DISTROLESS_IMAGE}
 LABEL maintainer="Travis Rhoden <trhoden@vmware.com>"
+
+USER nobody
```
andrewsykim: So this is the recommended non-root user for the distroless image?

dvonthenen: Yep! Other images based off distroless use it as well:
https://github.com/kubernetes/kube-state-metrics/blob/111c95ac200404aee140af9ff97c7f833e5dd087/Dockerfile

Issue #275, opened 2019-11-07
Support yaml for cloud config


andrewsykim:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
Support the cloud config being in yaml instead of gcfg. We would have to support both but I assume most users are already comfortable with yaml. 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- vsphere-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

frapposelli:
We should make a priority for v1.2.0 to transition to yaml

PR #305, opened 2020-03-18
Initial implementation for YAML based config

dvonthenen:
Implements a new YAML-based cloud-config which will enable a bunch of new features going forward:
- more expressive configuration which YAML provides
- is inline with the YAML-based configurations used all over in k8s
- allows the possibility for versioning in the cloud-config when YAML-based
- fixes a bunch of YAML and INI naming consistency issues in the config file as they relate to the new LoadBalancer feature
- the possibilities are endless 😀

**Which issue this PR fixes**: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/275

- Tested on a vSphere 6.7u3 configuration with 2 VCs and 3 DCs using zone support. 
- Verified my existing INI-based cloud-config worked with the code changes for backward compatibility
- Created and verified the new YAML-based cloud-config works
- Adds a bunch of YAML based `go test` to verify functionality

**Release note**:
Since this PR preserves backward compatibility of the INI-based cloud-config (and has been tested), this should have little impact (as of right now) to the end-user.


dvonthenen:
CC: @andrewsykim @frapposelli 

The lint test/check is failing because of the struct names ConfigINI and ConfigYAML, but I would like to keep these names as-is for now because this is only temporary since this also has a future deprecation notice for the INI-based configuration.

The integration test is failing for some unknown reason. I think it might be some issue with the timing when cluster coming up in the test or perhaps the verification that it's deployed correctly. See below:
```
make[2]: Entering directory '/home/prow/go/src/k8s.io/cloud-provider-vsphere/test/integration'
while ! kubectl -n kube-system logs vsphere-cloud-controller-manager >/dev/null 2>&1; do sleep 1; done && \
sleep 3 && \
kubectl -n kube-system logs vsphere-cloud-controller-manager
Error from server: Get https://ccm-integration-test-worker:10250/containerLogs/kube-system/vsphere-cloud-controller-manager/vsphere-cloud-controller-manager: dial tcp: lookup ccm-integration-test-worker on 10.63.240.10:53: no such host
```

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -67,35 +67,42 @@ Steps that will be covered in order to setup zones for the vSphere CPI, vSphere
...
+  port: "443"
+  # set insecure-flag to true if the vCenter uses a self-signed cert
+  insecure-flag: true
+  # settings for using k8s secret
+  secret-name: cpi-secret
+  secret-namespace: kube-system
+
+# VirtualCenter section
+virtualcenter:
+  tenant1:
```
andrewsykim: In the future we should support a list for this, something like:
```yaml
virutalcenter:
  - name: tenant1
    user: ...
    password: ...
  - name: tenant2
    user: ..
```

pkg/cloudprovider/vsphere/config/config_ini_legacy.go
```diff
@@ -0,0 +1,72 @@
+/*
+Copyright 2019New The Kubernetes Authors.
```
andrewsykim: nit: need a space here

pkg/cloudprovider/vsphere/config/config_yaml.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// ReadCPIConfigYAML parses vSphere cloud config file and stores it into CPIConfigYAML.
+func ReadCPIConfigYAML(byConfig []byte) (*CPIConfig, error) {
+	if len(byConfig) == 0 {
+		return nil, fmt.Errorf("Invalid YAML file")
+	}
+
+	/*
+		vCFG, err := vcfg.ReadRawConfigYAML(byConfig)
```
andrewsykim: Can this be removed now?

pkg/cloudprovider/vsphere/config/config_yaml.go
```diff
@@ -0,0 +1,84 @@
+/*
+Copyright 2019New The Kubernetes Authors.
```
andrewsykim: nit: need a space here

pkg/cloudprovider/vsphere/config/config_yaml_test.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
andrewsykim: nit: fix spacing on license header here

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
```diff
@@ -17,35 +17,39 @@
...
+/*
+	TODO:
+	When the INI based cloud-config is deprecated. This file should be deleted.
+*/
+
+func TestReadINIConfig(t *testing.T) {
 	contents := `
 [LoadBalancer]
-ipPoolName = pool1
+ip-pool-name = pool1
```
andrewsykim: Why the field name changes here?

pkg/cloudprovider/vsphere/loadbalancer/config/types_yaml.go
```diff
@@ -0,0 +1,80 @@
...
+
+	LBConfigYAML -> LBConfig
+	LoadBalancerClassConfigYAML -> LoadBalancerClassConfig
+	LoadBalancerClassConfigYAML -> LoadBalancerClassConfig
+	NsxtConfigYAML -> NsxtConfig
+*/
+
+// LBConfigYAML  is used to read and store information from the cloud configuration file
+type LBConfigYAML struct {
+	LoadBalancer      LoadBalancerConfigYAML                  `yaml:"loadbalancer"`
```
andrewsykim: I wonder if we should use json tags here and use a json -> yaml serializer. That would give us json support for free while also supporting yaml :thinking:

frapposelli: +1 on this

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
frapposelli: we should really change `virtualcenter` to `vcenter` as well, the Virtual Center name has been deprecated 10 years ago 😄

pkg/cloudprovider/vsphere/config/config_ini_legacy.go
dvonthenen: Fixed

pkg/cloudprovider/vsphere/config/config_yaml.go
dvonthenen: Removed

pkg/cloudprovider/vsphere/config/config_yaml_test.go
dvonthenen: Fixed spacing in entire block

pkg/cloudprovider/vsphere/config/config_yaml.go
dvonthenen: Fixed

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
dvonthenen: changed `virtualcenter:` to `vcenter:` in code and in docs

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
dvonthenen: We haven't cut a release since this functionality was introduced and I wanted to make the INI parsing values to be match other keys in all other configs. Currently, all keys use `-` between words. If we can't accept this change, I can revert but it would be nice to have uniform INI keys.

pkg/cloudprovider/vsphere/loadbalancer/config/types_yaml.go
dvonthenen: Spoke with @andrewsykim about this in slack. I had a concern (not sure if it is a valid concern) with supporting multiple different configuration format types. We currently support INI (legacy) and then want to add YAML for some advanced configurations using YAML unlocks. The original idea was to have a uniform experience for k8s configuration and then the CPI. The concern is having multiple different formats INI, YAML, JSON and etc and supporting all the different flavors. It seems like it's more to go wrong especially when it comes to a user configuration/interface to the CPI.

If this isn't a valid concern or there is additional calls from the community, Andrew and I can always include the additional JSON support.

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
dvonthenen: @andrewsykim I think this is the only item left for this PR

dvonthenen:
@andrewsykim we just have the one item/flag left... I can change it either way. If we do change it back, it would be preferred if the newer YAML at least matches.

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
dvonthenen: Talked with @andrewsykim and since we haven't had a release with the new LoadBalancer changes, there aren't any backward compatibility issues to be had and it's safe to update the INI keys to match all other key formats.

dvonthenen:
I took another pass at reviewing this PR and I think it looks good... and all current feedback has been addressed.

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -67,35 +67,42 @@ Steps that will be covered in order to setup zones for the vSphere CPI, vSphere
...
-        # user, password, datacenters will be used from Global section.
-
-[Labels]
-region = k8s-region
-zone = k8s-zone
+# Global properties in this section will be used for all specified vCenters unless overriden in VirtualCenter section.
+global:
+  user: YourVCenterUser
+  password: YourVCenterPass
+  port: "443"
```
andrewsykim, line 79: Guessing this has to be a string for compatiblity?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -17,18 +17,20 @@ limitations under the License.
...
 	"runtime"
 
 	v1 "k8s.io/api/core/v1"
-	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/klog"
 
-	"github.com/vmware/vsphere-automation-sdk-go/runtime/log"
+	cloudprovider "k8s.io/cloud-provider"
 
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/log"
```
andrewsykim, line 30: non-blocking nit: we should group github.com imports separately.

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -161,31 +174,30 @@ func (vs *VSphere) HasClusterID() bool {
...
+			log.SetLogger(NewKlogBridge())
 		}
-	}
-	if lb == nil {
-		klog.Infof("NSX-T load balancer support disabled")
 	} else {
-		klog.Infof("NSX-T load balancer support enabled. This feature is alpha, use in production at your own risk.")
-		// redirect vapi logging from the NSX-T GO SDK to klog
-		log.SetLogger(NewKlogBridge())
+		// explicitly nil the LB interface if ENABLE_ALPHA_NSXT_LB is not set even if the LBConfig is valid
```
andrewsykim, line 193: Previous this was 
```	
if lb != nil {
    klog.Infof("To enable NSX-T load balancer support you need to set the env variable ENABLE_ALPHA_NSXT_LB")
    lb = nil
}
```

Do we want to keep that log message?

pkg/cloudprovider/vsphere/config/config_yaml.go
```diff
@@ -0,0 +1,71 @@
+/*
+Copyright 2019 New The Kubernetes Authors.
```
andrewsykim, line 2: 2020 cause new file?

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
```diff
@@ -62,36 +66,36 @@ host = nsxt-server
...
+	assertEquals("NSXT.password", config.NSXT.Password, "secret")
+	assertEquals("NSXT.host", config.NSXT.Host, "nsxt-server")
 }
 
-func TestReadConfigOnVMC(t *testing.T) {
+func TestReadINIConfigOnVMC(t *testing.T) {
 	contents := `
 [LoadBalancer]
-ipPoolID = 123-456
+ip-pool-id = 123-456
```
andrewsykim, line 86: For posterity: this is not a breaking change since we haven't release a version supporting this yet. 

cc @mandelsoft @MartinWeindel

pkg/cloudprovider/vsphere/loadbalancer/config/types_yaml.go
```diff
@@ -0,0 +1,80 @@
...
+// LoadBalancerConfigYAML contains the configuration for the load balancer itself
+type LoadBalancerConfigYAML struct {
+	Size             string            `yaml:"size"`
+	LBServiceID      string            `yaml:"lb-service-id"`
+	Tier1GatewayPath string            `yaml:"tier1-gateway-path"`
+	AdditionalTags   map[string]string `yaml:"tags"`
+
+	// this struct use to inherit from LoadBalancerClassConfigYAML, but the YAML parser
+	// wasnt able to indirectly parse inherited fields
+	IPPoolName        string `yaml:"ip-pool-name"`
```
andrewsykim, line 46: `ip-pool-name` makes sense for INI but I think best practice for yaml is to use camel case?

pkg/common/config/config_ini_legacy.go
```diff
@@ -0,0 +1,267 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim, line 2: 2020

pkg/common/config/config_yaml.go
```diff
@@ -0,0 +1,231 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
andrewsykim, line 2: 2020

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
dvonthenen: No, definitely not. Just changed it to uint.

pkg/cloudprovider/vsphere/cloud.go
dvonthenen: Fixed

pkg/cloudprovider/vsphere/cloud.go
dvonthenen: To formatting is a little weird... it should look like this:

`	if _, ok := os.LookupEnv("ENABLE_ALPHA_NSXT_LB"); ok {
		if lb == nil {
			klog.Warning("To enable NSX-T load balancer support you need to configure section LoadBalancer")
		} else {
			klog.Infof("NSX-T load balancer support enabled. This feature is alpha, use in production at your own risk.")
			// redirect vapi logging from the NSX-T GO SDK to klog
			log.SetLogger(NewKlogBridge())
		}
	} else {
		// explicitly nil the LB interface if ENABLE_ALPHA_NSXT_LB is not set even if the LBConfig is valid
		// ENABLE_ALPHA_NSXT_LB must be explicitly enabled
		lb = nil
	}`

dvonthenen: So basically if someone has the ENV set, and `lb == nil`, then the interface isn't set and we should warn the user.
If both are set, we fall into the else.

The problem with outer else... is someone didn't set the ENVVAR but the interface isn't `nil` and it would cause the CCM to crash.

pkg/cloudprovider/vsphere/config/config_yaml.go
dvonthenen: Yup definitely

pkg/common/config/config_ini_legacy.go
dvonthenen: Yup fixed

pkg/common/config/config_yaml.go
dvonthenen: Fixed

pkg/cloudprovider/vsphere/loadbalancer/config/types_yaml.go
dvonthenen: Oops. Good catch... it's what happens when you copy and paste using INI as the base for the go file.

dvonthenen:
There are some changes (port from string to unit) that I want to retest with before giving my "thumbs up" again for additional review and/or merge.

pkg/cloudprovider/vsphere/loadbalancer/config/config_ini_legacy_test.go
dvonthenen: FYI, the README (which I missed initially but caught in this pass) have been updated to default to YAML and the `go test` for the LB config was originally for INI and is now YAML (passes).

dvonthenen:
Tested on a 1.17.3 config with zones... looks good to me!

pkg/cloudprovider/vsphere/config/config_yaml_test.go
```diff
@@ -0,0 +1,101 @@
...
+	When the INI based cloud-config is deprecated. This file should be deleted.
+*/
+
+const subnetCidrYAMLConfig = `
+global:
+  server: 0.0.0.0
+  port: 443
+  user: user
+  password: password
+  insecure-flag: true
```
andrewsykim, line 34: insecureFlag? or just `insecure`?

pkg/cloudprovider/vsphere/config/config_yaml_test.go
```diff
@@ -0,0 +1,101 @@
...
+  datacenters:
+    - us-west
+  ca-file: /some/path/to/a/ca.pem
```
andrewsykim, line 37: caFile

andrewsykim: or caPath, unless we are trying to line this up with the INI config fields

docs/book/tutorials/deploying_cpi_and_csi_with_multi_dc_vc_aka_zones.md
```diff
@@ -67,35 +67,42 @@ Steps that will be covered in order to setup zones for the vSphere CPI, vSphere
...
+  # set insecure-flag to true if the vCenter uses a self-signed cert
+  insecure-flag: true
```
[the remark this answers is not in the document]
dvonthenen: Will fix

pkg/cloudprovider/vsphere/config/config_yaml_test.go
dvonthenen: Will fix

pkg/cloudprovider/vsphere/config/config_yaml_test.go
dvonthenen: Definitely should be changed

dvonthenen:
Updated based on latest feedback from @andrewsykim

manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml
```diff
@@ -26,7 +30,7 @@ spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
       securityContext:
-        runAsUser: 0
```
andrewsykim, line 29: Are the changes in this file intentional?

dvonthenen: yes, it is. I forgot to pull down the YAML I used to test with. There were discrepancies between the vsphere-cloud-controller-manager-ds.yaml and vsphere-cloud-controller-manager-pod.yaml. This resolves it... the ds.yaml should be running as user 1001 which is a non-root account.

dvonthenen: Should have been changed for this PR:
https://github.com/kubernetes/cloud-provider-vsphere/pull/297

PR #321, opened 2020-04-15
support NSX-T connection with proxy

MartinWeindel:
For the NSX-T load balancer support, the connection should also work, if the NSX-T endpoint is only reachable via a HTTP(S) proxy. This PR uses the standard mechanism, e.g. automatically using the environment variable `HTTPS_PROXY` if defined (see https://golang.org/pkg/net/http/#ProxyFromEnvironment).

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
+++ b/pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
@@ -128,6 +128,7 @@ func NewNsxtBroker(nsxtConfig *config.NsxtConfig) (NsxtBroker, error) {
 	}
 	httpClient := http.Client{
 		Transport: &http.Transport{
+			Proxy:           http.ProxyFromEnvironment,
 			TLSClientConfig: tlsConfig,
 		},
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go
```diff
@@ -128,6 +128,7 @@ func NewNsxtBroker(nsxtConfig *config.NsxtConfig) (NsxtBroker, error) {
 	}
 	httpClient := http.Client{
 		Transport: &http.Transport{
+                        Proxy:           http.ProxyFromEnvironment,
```
andrewsykim, line 131: nit: formatting looks off here

MartinWeindel: fixed, I missed to push the last commit

dvonthenen:
/lgtm

That integration test has been failing for some time now... I haven't gotten around to fixing or root causing what is failing

PR #329, opened 2020-04-29
fix invalid url

wyyxd2017:
i fix a invalid url.thanks


dvonthenen:
I realized this was a mistake... need to revert this change.

PR #343, opened 2020-06-19
return err if VM guest property is nil

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

When fetching VM properties, if the guest property is nil, we should return an err to avoid a panic. Returning an err will also trigger a retry.

**Release note**:

```release-note
Fix a nil panic when guest property is nil
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -181,6 +181,10 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		return err
 	}
 
+	if oVM.Guest == nil {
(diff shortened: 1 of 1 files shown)


dvonthenen:
Hmmmm the unit tests are choking on an test that doesn't seem to affect this PR

dvonthenen:
Manually merging... will need to update the unit test as the requirements for the LoadBalancer have changed underneath us

```
pkg/cloudprovider/vsphere/loadbalancer/nsxt_broker.go:174:2: tlsConfig.BuildNameToCertificate is deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (SA1019)
```

PR #355, opened 2020-07-17
Fixed a bug in InstanceExistsByProviderID()

zuzzas:

There seems to be a bug that does not report an error to the CCM code. That may lead to Node deletion when an API access fails.

I mean... it does look like a bug, but can't shake a feeling that I've missed something.

**Release note**:

```release-note
Errors during InstanceExistsByProviderID() stopped leading to Node object deletion
```

### DIFF
--- a/pkg/cloudprovider/vsphere/instances.go
+++ b/pkg/cloudprovider/vsphere/instances.go
@@ -159,13 +159,14 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 		return true, nil
 	}
 
-	if err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID); err == nil {
+	err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID)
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vsphere/instances.go
```diff
@@ -159,13 +159,14 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 		return true, nil
 	}
 
-	if err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID); err == nil {
+	err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID)
+	if err == nil {
 		klog.V(2).Info("instances.InstanceExistsByProviderID() EXISTS with ", uid)
 		return true, err
```
andrewsykim, line 165: Slightly clearer with:
```
return true, nil
```

zuzzas: Done!

Issue #348, opened 2020-06-30
Support node deletion of terminated VMs


andrewsykim:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

In the past we opted not to delete the node object of a terminated VM due to certain failure scenarios.  However, this behavior is not consistent with other cloud providers. We should re-evaluate if we want to re-introduce this behavior.

Specifically, `InstanceExistsByProviderID` should return `false, nil` when the VM on vSphere no longer exists, not if the Kubernetes node object exists. This cleans up Kubernetes node objects automatically in the event that a VM is deleted.

frapposelli:
I'm fine with this behavior being reinstated as the default, but we could use a feature toggle to enable use cases that require VMs to be persisted even when deleted from the underlying platform

jayunit100:
Is this inflamed by using a CNI like calico that makes a new eth pair for every pod?  In other CNIs like antrea, you only get one eth per pod.  examples below respectively of antrea/calico.

```
# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: tunl0@NONE: <NOARP,UP,LOWER_UP> mtu 1440 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
    inet 192.168.0.192/32 brd 192.168.0.192 scope global tunl0
       valid_lft forever preferred_lft forever
90: eth0@if91: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:12:00:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.18.0.3/16 brd 172.18.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fc00:f853:ccd:e793::3/64 scope global nodad 
       valid_lft forever preferred_lft forever
    inet6 fe80::42:acff:fe12:3/64 scope link 
       valid_lft forever preferred_lft forever

```

But calico... you get alot ... 
```
[...]

PR #359, opened 2020-07-27
InstanceExistsByProviderID Signal Deletion to K8s

dvonthenen:
This PR changes the default behavior to mimic other cloud providers controllers so that if the vSphere VM is no longer accessible/present according to the vCenter Server, the node VM will be deleted.

An option exists (via an environment variable) to prevent the deletion of the VM which preserves the original behavior.

**Which issue this PR fixes**: https://github.com/kubernetes/cloud-provider-vsphere/issues/348

New capability is behind a switch... default behavior is unchanged.

**Release note**:
NA

### DIFF
--- a/pkg/cloudprovider/vsphere/instances.go
+++ b/pkg/cloudprovider/vsphere/instances.go
@@ -19,6 +19,7 @@ package vsphere
 import (
 	"context"
 	"errors"
+	"os"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -26,6 +27,7 @@ import (
 	"k8s.io/klog"
 
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
 )
 
 // Error constants
@@ -154,19 +156,25 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 
 	// Check if node has been discovered already
 	uid := GetUUIDFromProviderID(providerID)
-	if _, ok := i.nodeManager.nodeUUIDMap[uid]; ok {
-		klog.V(2).Info("instances.InstanceExistsByProviderID() CACHED with ", uid)
-		return true, nil
-	}
-
 	err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID)
 	if err == nil {
 		klog.V(2).Info("instances.InstanceExistsByProviderID() EXISTS with ", uid)
 		return true, nil
 	}
 
-	klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid)
-	return false, err
+	if err != vclib.ErrNoVMFound {
+		klog.V(4).Info("instances.InstanceExistsByProviderID() failed with ", uid, ". Err: ", err)
+		return false, err
+	}
+
+	// at this point, err is vclib.ErrNoVMFound
+	if _, ok := os.LookupEnv("SKIP_NODE_DELETION"); ok {
+		klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid, ". Override and prevent deletion.")
+		return false, err
+	}
+
+	klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid, ". Signaling deletion.")
+	return false, nil
 }
 
 // InstanceShutdownByProviderID returns true if the instance is in safe state to detach volumes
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/vsphere/instances.go
```diff
@@ -165,6 +166,11 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 		return true, nil
 	}
 
+	if _, ok := os.LookupEnv("ENABLE_DELETION_OF_VM"); ok {
+		klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid, ". Signaling deletion.")
+		return false, nil
```
andrewsykim, line 171: before returning `false, nil`  here, do we need to check the error type? `DiscoverNode` may have returned an err not related to the node not being found?

dvonthenen: Good call!

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -165,6 +167,13 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 		return true, nil
 	}
 
+	if err == vclib.ErrNoVMFound {
+		if _, ok := os.LookupEnv("ENABLE_DELETION_OF_VM"); ok {
```
andrewsykim, line 171: One more thing, should this be something in the cloud config instead of an env var?

dvonthenen: That's a good question... do we have a feeling what option most users are going to select? Currently, the original behavior is preserved for backward compatibility... if we feel that users going to all go one way, then maybe the environment variable is good enough for the time being. Another way to look at it is, do we think this option is going to be around for a very very long time, if so then we should have it in the config.

If we do have it in the config, then it should go here in a new struct called like CPIOptions or something:
https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/config/types_yaml.go#L45

dvonthenen: Talked about this and came to agreement that we should fall in-line with other cloud provider controller behaviors.

dvonthenen:

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -165,8 +167,19 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
 		return true, nil
 	}
 
-	klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid)
-	return false, err
+	if err != vclib.ErrNoVMFound {
```
andrewsykim, line 170: Should we remove the cache check in line 158?

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -165,8 +167,19 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
...
+		klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid, ". Err: ", err)
+		return false, err
+	}
+
+	// at this point, err is vclib.ErrNoVMFound
+	if _, ok := os.LookupEnv("PREVENT_DELETION_OF_VM"); ok {
```
andrewsykim, line 176: Thoughts on `SKIP_NODE_DELETION` instead?

pkg/cloudprovider/vsphere/instances.go
dvonthenen: Removed. Should not need to cache.

pkg/cloudprovider/vsphere/instances.go
dvonthenen: Changed

dvonthenen:
Latest changes based on feedback!

dvonthenen:
Also noticed a small documentation error as well. "443" (string) -> 433 (int)

pkg/cloudprovider/vsphere/instances.go
```diff
@@ -154,19 +156,25 @@ func (i *instances) InstanceExistsByProviderID(ctx context.Context, providerID s
...
 	err := i.nodeManager.DiscoverNode(uid, cm.FindVMByUUID)
 	if err == nil {
 		klog.V(2).Info("instances.InstanceExistsByProviderID() EXISTS with ", uid)
 		return true, nil
 	}
 
-	klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid)
-	return false, err
+	if err != vclib.ErrNoVMFound {
+		klog.V(4).Info("instances.InstanceExistsByProviderID() NOT FOUND with ", uid, ". Err: ", err)
```
andrewsykim, line 166: the "NOT FOUND" should be removed from this log line

dvonthenen: yup, replaced with failed

PR #361, opened 2020-07-27
Log does not print node initialization success when zone labels are not configured

dvonthenen:
When the zone/region labels are not configured in the cloud config, the node initialization process will fail to print a "successfully initialized node" in the log but the node will still function correctly. 

Does not change the behavior of the CPI

**Release note**:
No end-user doc or interface changes

### DIFF
--- a/pkg/cloudprovider/vsphere/zones.go
+++ b/pkg/cloudprovider/vsphere/zones.go
@@ -43,6 +43,10 @@ func (z *zones) GetZone(ctx context.Context) (cloudprovider.Zone, error) {
 
 	zone := cloudprovider.Zone{}
 
+	if len(z.region) == 0 || len(z.zone) == 0 {
+		return zone, nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/vsphere/zones.go
```diff
@@ -90,6 +94,10 @@ func (z *zones) GetZoneByNodeName(ctx context.Context, nodeName k8stypes.NodeNam
 
 	zone := cloudprovider.Zone{}
 
+	if len(z.region) == 0 || len(z.zone) == 0 {
```
andrewsykim, line 97: Maybe for a future PR, should this be an && check? Do we support only setting regions?

dvonthenen: I believe both are required for out-of-tree. I know in-tree only requires one of them (forgot which one). I definitely have only ever tested when both the zone and region are set.

PR #365, opened 2020-07-30
instances: return InstanceNotFound error in InstanceID

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

Returns InstanceNotFound error in InstanceID which ensures non-existent nodes are deleted even before their provider ID is set. 

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/instances.go
+++ b/pkg/cloudprovider/vsphere/instances.go
@@ -111,7 +111,8 @@ func (i *instances) InstanceID(ctx context.Context, nodeName types.NodeName) (st
 		return node.UUID, nil
 	}
 
-	if err := i.nodeManager.DiscoverNode(string(nodeName), cm.FindVMByName); err == nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/vsphere/instances.go
```diff
@@ -120,8 +121,13 @@ func (i *instances) InstanceID(ctx context.Context, nodeName types.NodeName) (st
...
 	}
 
+	if err != vclib.ErrNoVMFound {
+		klog.V(4).Infof("instances.InstanceID() failed with err: %v", err)
+		return "", err
+	}
+
 	klog.V(4).Info("instances.InstanceID() NOT FOUND with ", string(nodeName))
-	return "", ErrNodeNotFound
+	return "", cloudprovider.InstanceNotFound
```
andrewsykim, line 130: I guess we need the env var gate here too?

andrewsykim: fixed

Issue #368, opened 2020-08-03
cloud controller is unable to patch node with IP addresses from cloud provider


deshui123:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
> /kind bug

**What happened**:
For IPv6(both internal and external are IPv6), cloud controller is unable to patch node with IP addresses from cloud provider on a multi-NIC VM.

I have added **internal-vm-network-name**, but this can not work for all nodes, even there is some logs look like 'Adding Internal IP by NetworkName: 2001:da8:8001:801::115'

```
[root@m6-g1-01 ~]# kubectl get node -o wide
NAME       STATUS   ROLES    AGE     VERSION   INTERNAL-IP              EXTERNAL-IP         OS-IMAGE                                      KERNEL-VERSION                CONTAINER-RUNTIME
m6-g1-01   Ready    <none>   5h27m   v1.18.6   2001:da8:8001:801::111   <none>              Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
m6-g1-02   Ready    <none>   5h27m   v1.18.6   2001:da8:8001:801::112   <none>              Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
m6-g1-03   Ready    <none>   5h27m   v1.18.6   2600:0:0:810::108        2600:0:0:810::108   Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
m6-g2-01   Ready    <none>   5h26m   v1.18.6   2600:0:0:810::109        2600:0:0:810::109   Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
m6-g2-02   Ready    <none>   5h26m   v1.18.6   2600:0:0:810::110        2600:0:0:810::110   Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
m6-g3-01   Ready    <none>   5h26m   v1.18.6   2001:da8:8001:801::116   <none>              Red Hat Enterprise Linux Server 7.8 (Maipo)   4.18.0-147.8.1.el8_1.x86_64   docker://1.13.1
[...]

PR #378, opened 2020-08-11
External/Internal Address Discovery Fix

dvonthenen:
Fixes a logic issue when using the address CIDR selection and VM Network name selection when determining internal/external IP addresses.

**Which issue this PR fixes**: https://github.com/kubernetes/cloud-provider-vsphere/issues/368

Case 1: Tested using CIDR selection for a private range of IPs
Case 2: Tested using two different VM Networks with dual NIC for a subset of VMs/kubernetes nodes
Case 3: Testing using internal IP on 1 VM Network and the external IP on another VM Network so both internal and external have different IPs.

**Release note**:
NA

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -221,12 +221,8 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		externalVMNetworkName = nm.cfg.Nodes.ExternalVMNetworkName
 	}
 
-	var addressMatchingEnabled bool
-	if internalNetworkSubnet != nil && externalNetworkSubnet != nil {
-		addressMatchingEnabled = true
-	}
-
-	found := false
+	foundInternal := false
+	foundExternal := false
 	addrs := []v1.NodeAddress{}
 
 	klog.V(2).Infof("Adding Hostname: %s", oVM.Guest.HostName)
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+							Type:    v1.NodeExternalIP,
+							Address: ip,
+						},
+					)
+					foundExternal = true
+				}
+			}
+
+			// At least one of the Internal or External addresses has been found.
+			// Both technically need to exist for the node to function correctly.
```
andrewsykim, line 311: I think it's valid for only one of them to be defined, preferable InternalIP.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+				)
+				foundInternal = true
+				foundExternal = true
 				break
 			}
 		}
 	}
 
-	if !found {
+	if !foundInternal || !foundExternal {
```
andrewsykim, line 346: I think we want this to be `!foundIInternal && !foundExternal` since it should be allowed for only 1 address to be set.

andrewsykim:
@gwang550 can you help review this please?

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: ok

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: Will update comment

dvonthenen:

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+					klog.V(2).Infof("Adding Internal IP by AddressMatching: %s", ip)
 					v1helper.AddToNodeAddresses(&addrs,
 						v1.NodeAddress{
 							Type:    v1.NodeInternalIP,
 							Address: ip,
 						},
 					)
-					found = true
-					break
+					foundInternal = true
```
andrewsykim, line 274: break here?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+					klog.V(2).Infof("Adding External IP by AddressMatching: %s", ip)
 					v1helper.AddToNodeAddresses(&addrs,
 						v1.NodeAddress{
 							Type:    v1.NodeExternalIP,
 							Address: ip,
 						},
 					)
-					found = true
-					break
+					foundExternal = true
```
andrewsykim, line 284: break here?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
-							Type:    v1.NodeExternalIP,
-							Address: ip,
-						}, v1.NodeAddress{
 							Type:    v1.NodeInternalIP,
 							Address: ip,
 						},
 					)
-					found = true
-					break
+					foundInternal = true
```
andrewsykim, line 296: break here?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -260,81 +256,94 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
 				}
+				if !foundExternal && externalVMNetworkName != "" && strings.EqualFold(externalVMNetworkName, v.Network) {
+					klog.V(2).Infof("Adding External IP by NetworkName: %s", ip)
+					v1helper.AddToNodeAddresses(&addrs,
+						v1.NodeAddress{
+							Type:    v1.NodeExternalIP,
+							Address: ip,
+						},
+					)
+					foundExternal = true
```
andrewsykim, line 306: break here?

Issue #381, opened 2020-08-21
CPI 1.2 deletes a node object right after its creation


yastij:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

the CAPV CI started to flake intensively after we upgraded to 1.2 CCM. after some debugging here's what we found:

after the bootstrap the kubelet is able to register the node sucessfully:

``` 
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: I0821 11:54:33.633787    1060 kubelet_node_status.go:294] Setting node annotation to enable volume controller attach/detach
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: I0821 11:54:33.635044    1060 kubelet_node_status.go:70] Attempting to register node cluster-lkdffl-md-0-5459fbd7cb-lcdz2
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: I0821 11:54:33.649212    1060 kubelet_node_status.go:73] Successfully registered node cluster-lkdffl-md-0-5459fbd7cb-lcdz2
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: E0821 11:54:33.669511    1060 kubelet.go:2267] node "cluster-lkdffl-md-0-5459fbd7cb-lcdz2" not found
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: E0821 11:54:33.769659    1060 kubelet.go:2267] node "cluster-lkdffl-md-0-5459fbd7cb-lcdz2" not found
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: E0821 11:54:33.869790    1060 kubelet.go:2267] node "cluster-lkdffl-md-0-5459fbd7cb-lcdz2" not found
Aug 21 11:54:33 cluster-lkdffl-md-0-5459fbd7cb-lcdz2 kubelet[1060]: E0821 11:54:33.969975    1060 kubelet.go:2267] node "cluster-lkdffl-md-0-5459fbd7cb-lcdz2" not found
[...]

PR #383, opened 2020-08-25
nodemanager: wait for hostname before registering node

andrewsykim:
Follow-up bug fix from https://github.com/kubernetes/cloud-provider-vsphere/pull/359

* Updates node manager to only discover a node if vmtools is reporting a valid hostname. 
* Only check VM by IP if not found by node name.
* Only check VM by reverse UUID if not found by normal UUID
* return explicit nil for better readability. 

**Release note**:

```release-note
Fix a bug where a node can be reported as 'not found' prior to vmtools running on the host.
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -28,6 +28,7 @@ import (
 	pb "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/proto"
 	vcfg "k8s.io/cloud-provider-vsphere/pkg/common/config"
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
 	v1helper "k8s.io/cloud-provider/node/helpers"
 	"k8s.io/klog"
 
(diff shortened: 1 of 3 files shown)


andrewsykim:
@gwang550 can you help review this please?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -113,6 +114,10 @@ func (nm *NodeManager) shakeOutNodeIDLookup(ctx context.Context, nodeID string,
 			return vmDI, err
 		}
 
+		if err != vclib.ErrNoVMFound {
```
dvonthenen, line 117: This will short circuit the lookup by IP because if the hostname is overridden to be the IP, you will definitely fail the first with no VM found. I think if you remove this, you will see get the desired results.

andrewsykim:
/hold

For reviews

pkg/cloudprovider/vsphere/nodemanager.go
andrewsykim: Not sure I follow, this only short circuits if the err is not `vclib.ErrNoVMFound`. So if the name check above fails with `vclib.ErrNoVMFound`, we would still run the FindVMByIP below right?

dvonthenen: Sorry my bad... that's right!

andrewsykim:
Fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/381

PR #389, opened 2020-09-02
fix dual stack is not supported

gwang550:
This PR fixes the problem where the dual-stack logic is implemented but not work.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -17,6 +17,7 @@ limitations under the License.
 package vsphere
 
 import (
+	"errors"
 	"io"
 	"io/ioutil"
 	"os"
@@ -200,11 +201,10 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 
 	// add alpha dual stack feature
 	for tenant := range cfg.VirtualCenter {
-		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) == 2 {
-			print(len(cfg.VirtualCenter[tenant].IPFamilyPriority))
+		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) > 1 {
 			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok{
-				klog.Errorf("ALPHA DUAL STACK feature enabled, number of  ip family provided not matched for %s", tenant)
-				panic("Two IP families provides, but dual stack feature is not enabled!")
+				klog.Errorf("number of ip family provided for VCenter %s is 2, ENABLE_ALPHA_DUAL_STACK env var is not set", tenant)
+				return nil, errors.New("two IP families provided, but dual stack feature is not enabled")
 			}
 		}
 	}
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -202,7 +202,7 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 	// add alpha dual stack feature
 	for tenant := range cfg.VirtualCenter {
 		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) > 1 {
-			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok{
+			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok {
 				klog.Errorf("number of ip family provided for VCenter %s is 2, ENABLE_ALPHA_DUAL_STACK env var is not set", tenant)
 				return nil, errors.New("two IP families provided, but dual stack feature is not enabled")
 			}
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -198,6 +198,17 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 		lb = nil
 	}
 
+	// add alpha dual stack feature
+	for tenant := range cfg.VirtualCenter {
+		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) == 2 {
+			print(len(cfg.VirtualCenter[tenant].IPFamilyPriority))
+			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok{
+				klog.Errorf("ALPHA DUAL STACK feature enabled, number of  ip family provided not matched for %s", tenant)
+				panic("Two IP families provides, but dual stack feature is not enabled!")
+			}
+		}
+	}
+
 	vs := VSphere{
 		cfg:          cfg,
 		cfgLB:        lbcfg,
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -360,7 +360,6 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 			if !foundInternal && !foundExternal {
 				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
 			}
-
 		}
 	}
 
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -234,8 +234,8 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		externalVMNetworkName = nm.cfg.Nodes.ExternalVMNetworkName
 	}
 
-	foundInternal := false
-	foundExternal := false
+	var foundInternal bool
+	var foundExternal bool
 	addrs := []v1.NodeAddress{}
 
 	klog.V(2).Infof("Adding Hostname: %s", oVM.Guest.HostName)
@@ -267,6 +267,8 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		// Must break out of loop in the event of ipv6,ipv4 where the NIC does
 		// contain a valid IPv6 and IPV4 address
 		for _, family := range ipFamily {
+			foundInternal = false
+			foundExternal = false
 			ips := returnIPsFromSpecificFamily(family, v.IpAddress)
 
 			for _, ip := range ips {
@@ -330,7 +332,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				} else if !foundInternal && foundExternal {
 					klog.Warning("External address found, but internal address not found. Returning what addresses were discovered.")
 				}
-				break
+				continue
 			}
 
 			// Neither internal or external addresses were found. This defaults to the old
@@ -353,12 +355,14 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				foundExternal = true
 				break
 			}
+
+			if !foundInternal && !foundExternal {
+				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
+			}
 		}
 	}
 
-	if !foundInternal && !foundExternal {
-		return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
-	}
+
 
 	klog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
 		nodeID, vmDI.VM, vmDI.VcServer, vmDI.DataCenter.Name())
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"os"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
@@ -268,6 +267,8 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		// Must break out of loop in the event of ipv6,ipv4 where the NIC does
 		// contain a valid IPv6 and IPV4 address
 		for _, family := range ipFamily {
+			foundInternal = false
+			foundExternal = false
 
 			ips := returnIPsFromSpecificFamily(family, v.IpAddress)
 
@@ -332,13 +333,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				} else if !foundInternal && foundExternal {
 					klog.Warning("External address found, but internal address not found. Returning what addresses were discovered.")
 				}
-				if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); ok {
-					foundExternal = false
-					foundInternal = false
-					continue
-				} else {
-					break
-				}
+				continue
 			}
 
 			// Neither internal or external addresses were found. This defaults to the old
@@ -366,13 +361,6 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
 			}
 
-			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); ok {
-
-				foundExternal = false
-				foundInternal = false
-			} else {
-				break
-			}
 		}
 	}
 
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -364,8 +364,6 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		}
 	}
 
-
-
 	klog.V(2).Infof("Found node %s as vm=%+v in vc=%s and datacenter=%s",
 		nodeID, vmDI.VM, vmDI.VcServer, vmDI.DataCenter.Name())
 	klog.V(2).Info("Hostname: ", oVM.Guest.HostName, " UUID: ", oVM.Summary.Config.Uuid)
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 
 	v1 "k8s.io/api/core/v1"
@@ -234,8 +235,8 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		externalVMNetworkName = nm.cfg.Nodes.ExternalVMNetworkName
 	}
 
-	var foundInternal bool
-	var foundExternal bool
+	foundInternal := false
+	foundExternal := false
 	addrs := []v1.NodeAddress{}
 
 	klog.V(2).Infof("Adding Hostname: %s", oVM.Guest.HostName)
@@ -267,8 +268,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		// Must break out of loop in the event of ipv6,ipv4 where the NIC does
 		// contain a valid IPv6 and IPV4 address
 		for _, family := range ipFamily {
-			foundInternal = false
-			foundExternal = false
+
 			ips := returnIPsFromSpecificFamily(family, v.IpAddress)
 
 			for _, ip := range ips {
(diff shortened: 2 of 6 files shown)


dvonthenen:
We only support 1 IP family type currently. Any other configuration or attempt to run with a dual-stack is not supported and is documented. https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/types_yaml.go#L67-L70

If we did want to add code to address that, the code provided is insufficient because the call to `AddToNodeAddresses` is already made which adds the address to the list which makes that `continue` logically not correct.

Also, it is valid to only have just an Internal address for airgapped/private clusters. There is a comment to reflect that use case:
https://github.com/kubernetes/cloud-provider-vsphere/pull/389/files#diff-b7f52468613fc2d2fc4b257bed84a46cR325-R328

There is an effort to add dual-stack support very soon, so we might want to just leave the code as is for now.

andrewsykim:
> We only support 1 IP family type currently. Any other configuration or attempt to run with a dual-stack is not supported and is documented. https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/types_yaml.go#L67-L70

I think future looking we wanted to support `ipv4,ipv6` or `ipv6,ipv4` right? That's why ipFamily is currently a string array.

> If we did want to add code to address that, the code provided is insufficient because the call to AddToNodeAddresses is already made which adds the address to the list which makes that continue logically not correct.

Can you point where exactly? Having a hard time finding which point in the code you're referring to.

gwang550:
> Can you point where exactly? Having a hard time finding which point in the code you're referring to.
I think Daivd is referring to the code starting from [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/780bec4e840352cbb1b6fd2461a62f371456db9c/pkg/cloudprovider/vsphere/nodemanager.go#L283) that's where AddToNodeAddresses Happens.

dvonthenen:
> > We only support 1 IP family type currently. Any other configuration or attempt to run with a dual-stack is not supported and is documented. https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/types_yaml.go#L67-L70
> 
> I think future looking we wanted to support `ipv4,ipv6` or `ipv6,ipv4` right? That's why ipFamily is currently a string array.

That is correct.


> > If we did want to add code to address that, the code provided is insufficient because the call to AddToNodeAddresses is already made which adds the address to the list which makes that continue logically not correct.
> 
> Can you point where exactly? Having a hard time finding which point in the code you're referring to.

See below...

dvonthenen:
> > Can you point where exactly? Having a hard time finding which point in the code you're referring to.
> 
> I think Daivd is referring to the code starting from [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/780bec4e840352cbb1b6fd2461a62f371456db9c/pkg/cloudprovider/vsphere/nodemanager.go#L283) that's where AddToNodeAddresses Happens.

Yup, that's where it is.

gwang550:
> If we did want to add code to address that, the code provided is insufficient because the call to `AddToNodeAddresses` is already made which adds the address to the list which makes that `continue` logically not correct.

The logic here is to add ip for the other ip-family if present, previously, it will just break out of the loop [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/22e85632bfe62f9a6afff8d2fc269f80eb9e7c78/pkg/cloudprovider/vsphere/nodemanager.go#L333)

andrewsykim:
> then checking to see if the cluster has got the IP in both format by kubectl get node <node-name> -o jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}' and it shows two format of the IP address

@gwang550 can you show the output from fetching all the node addresses? Does this mean there are 4 address types in total? 2 InternalIP (IPv4, IPv6) and 2 ExternalIP (IPv4, IPv6)?

gwang550:
> > then checking to see if the cluster has got the IP in both format by kubectl get node  -o jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}' and it shows two format of the IP address
> 
> @gwang550 can you show the output from fetching all the node addresses? Does this mean there are 4 address types in total? 2 InternalIP (IPv4, IPv6) and 2 ExternalIP (IPv4, IPv6)?

Here is result for internalIP 
10.185.35.245 2620:124:6020:c00b:0:a:0:153
 
and this is the result for ExternalIP
10.185.35.245 2620:124:6020:c00b:0:a:0:153

andrewsykim (commented):
Can we alpha feature gate the use of dual-stack addresses similar to how we feature gate NSX-T support?

https://github.com/kubernetes/cloud-provider-vsphere/blob/44a75efaef85b259f75a2cf729b60149dbb9353f/pkg/cloudprovider/vsphere/cloud.go#L187-L194

I think dual-stack in Kubernetes may change as @dvonthenen mentioned and I want to ensure we're allowed to break it later.

andrewsykim:
Can we also add some unit tests for this? There are some examples in `nodemanager_tes.go` you can reference.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -353,12 +361,22 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				foundExternal = true
 				break
 			}
+
+			if !foundInternal && !foundExternal {
+				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
+			}
+
+			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); ok {
```
andrewsykim, line 369: I think we want the alpha gate check to happen when checking the ipFamily field in the config.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -330,7 +332,13 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				} else if !foundInternal && foundExternal {
 					klog.Warning("External address found, but internal address not found. Returning what addresses were discovered.")
 				}
-				break
+				if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); ok {
```
andrewsykim, line 335: Would prefer the check for `ENABLE_ALPHA_DUAL_STACK` to happen in the same place we check the NSX-T gate in `buildVSphereFromConfig`:

https://github.com/kubernetes/cloud-provider-vsphere/blob/22e85632bfe62f9a6afff8d2fc269f80eb9e7c78/pkg/cloudprovider/vsphere/cloud.go#L179-L211

andrewsykim: `DiscoverNode` in `nodemanager.go` should assume dualstack if  `ipFamily` has multiple entries. But configuration of `ipFamily` should be gated by `ENABLE_ALPHA_DUAL_STACK`.

gwang550: yeah, that makes sense. So, how does the ipFamily look like when `ENABLE_ALPHA_DUAL_STACK` is disabled. Should ipfamily just become a one-entry array(i.e. leaving only the first entry in the array)?

andrewsykim: I think if `ENABLE_ALPHA_DUAL_STACK` is disabled but the config has multiple IP families, we should treat that as an error and exit. What do you think?

gwang550: I think what you proposed has a better behavior and informs the user to provide the correct ipfamily. Going to get started on that direction.

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -353,12 +356,15 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				foundExternal = true
 				break
 			}
+
+			if !foundInternal && !foundExternal {
```
dvonthenen, line 360: Did we want to error out here? If error and dual stack is enabled, this won't attempt to try using the second ipfamily right?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -330,7 +333,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 				} else if !foundInternal && foundExternal {
 					klog.Warning("External address found, but internal address not found. Returning what addresses were discovered.")
 				}
-				break
+				continue
```
dvonthenen, line 336: Based on the comment, should we be using a `continue`? If the comment isn't valid anymore, we should update the comment to reflect the behavior.

Also if `continue` is correct, we should reset `addrs = []v1.NodeAddress{}` unless having either the external/internal is valid with IPv4/IPv6 and the having an internal/external with a IPv6/IPv4 is valid.

In others words through an example:
- external has an IPv6 address
- internal has an IPv4 address

pkg/cloudprovider/vsphere/nodemanager.go
gwang550: yeah, if neither is found, then we want to error out here. If error and dual-stack is enabled, the loop will exit here and won't attempt to use the second ipfamily.

pkg/cloudprovider/vsphere/nodemanager.go
gwang550: Regarding the comment, that doesn't reflect the behavior here, the comment needs to be changed. However, I am not sure why must break out the loop when NIC does contain a valid IPv6 and IPV4 address, in IPv6, IPv4 case. 

My understanding is if both IPv6 and IPv4 contain valid addresses, then both will be added to the Node address, which is the behavior we wanted if dual-stack is enabled.

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: Ok cool

pkg/cloudprovider/vsphere/nodemanager.go
dvonthenen: ahhhh I gotcha now. I thought for some reason we were supported internal/external IPv6 and a internal/external IPv4. but here we are saying literally mixed mode dual stack when you can have an external IPv6 and an internal IPv4.

we should update the comment to reflect this change and I think we are good to go

dvonthenen:
Left the one comment/update... but code wise I think we are good to go.

Going to try and cut releases based on branches before merging this...
/hold

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -198,6 +198,17 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 		lb = nil
 	}
 
+	// add alpha dual stack feature
+	for tenant := range cfg.VirtualCenter {
+		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) == 2 {
+			print(len(cfg.VirtualCenter[tenant].IPFamilyPriority))
+			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok{
+				klog.Errorf("ALPHA DUAL STACK feature enabled, number of  ip family provided not matched for %s", tenant)
+				panic("Two IP families provides, but dual stack feature is not enabled!")
```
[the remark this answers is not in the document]
dvonthenen: I totally missed that... definitely a Fatal

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,71 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+		{
+			testName: "Vcenter ip family should be length of 2 when dual stack feature is enabled",
+			conf: `[Global]
+			user = user
+			password = password
+			datacenters = us-west
+			[VirtualCenter "0.0.0.0"]
+			user = user
+			password = password
+			ipFamily = ipv6,ipv4
```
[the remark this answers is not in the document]
gwang550: That's was not the latest commit, updated the changes.

pkg/cloudprovider/vsphere/cloud.go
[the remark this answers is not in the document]
gwang550: yeah, totally agreed. Changed the code and please review, and let me know. Thanks

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -198,6 +198,17 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 		lb = nil
 	}
 
+	// add alpha dual stack feature
+	for tenant := range cfg.VirtualCenter {
+		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) == 2 {
+			print(len(cfg.VirtualCenter[tenant].IPFamilyPriority))
+			if _, ok := os.LookupEnv("ENABLE_ALPHA_DUAL_STACK"); !ok{
+				klog.Errorf("ALPHA DUAL STACK feature enabled, number of  ip family provided not matched for %s", tenant)
```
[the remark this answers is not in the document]
gwang550: yeah, nice catch, changed the log message.

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -127,6 +127,88 @@ func TestDiscoverNodeByName(t *testing.T) {
 	}
 }
 
+func TestAlphaDualStack(t *testing.T) {
+	cfg, ok := configFromEnvOrSim(true)
+	defer ok()
+
```
andrewsykim, line 133: remove the extra lines here

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -127,6 +127,88 @@ func TestDiscoverNodeByName(t *testing.T) {
...
+				SystemUUID: k8sUUID,
+			},
+		},
+	}
+
+	// get node registered so node can be exported
+	nm.RegisterNode(node)
+
+	err = nm.DiscoverNode(name, cm.FindVMByName)
+
```
andrewsykim, line 179: remove this extra line

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -127,6 +127,88 @@ func TestDiscoverNodeByName(t *testing.T) {
...
+	size := len(ipv4Ips)
+	if size != 1 {
+		t.Errorf("Should only return single IPv4 address. expected: 1, actual: %d", size)
+	} else if !strings.EqualFold(ipv4Ips[0], "10.0.0.1") {
+		t.Errorf("IPv6 does not match. expected: 10.0.0.1, actual: %s", ipv4Ips[0])
+	}
+
+	//check ipv6 ip
+
```
andrewsykim, line 200: remove this line

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -127,6 +127,88 @@ func TestDiscoverNodeByName(t *testing.T) {
...
+	ipv6Ips := returnIPsFromSpecificFamily(vcfg.IPv6Family, ips)
+
+	size = len(ipv6Ips)
+	if size != 1 {
+		t.Errorf("Should only return single IPv6 address. expected: 1, actual: %d", size)
+	} else if !strings.EqualFold(ipv6Ips[0], "fd01:0:101:2609:bdd2:ee20:7bd7:5836") {
+		t.Errorf("IPv6 does not match. expected: fd01:0:101:2609:bdd2:ee20:7bd7:5836, actual: %s", ipv6Ips[0])
```
andrewsykim, line 207: Might be good to put this IPv6 address in a variable

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,113 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+	var (
+		username = "user"
+		password = "password"
+	)
+	var testcases = []struct {
+		testName                 string
+		conf                     string
+		enableDualStackFeature	 bool
+		expectedIsSecretProvided bool
+		expectedUsername         string
```
andrewsykim, line 231: expectedUsername and expectedPassword are not actually checked anywhere in the test, is that expected?

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,113 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+				t.Fatalf("readConfig: unexpected error returned: %v", err)
+			}
+		}
+
+		if testcase.enableDualStackFeature {
+			_ :os.Setenv("ENABLE_ALPHA_DUAL_STACK", "1")
+		}
+
+		vs, err = buildVSphereFromConfig(cfg, nil)
+		print(err)
```
andrewsykim, line 308: Use `t.Log` to print the error and we should probably only print it when not nil

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,113 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+					t.Fatalf("buildVSphereFromConfig: expected err: %s, receiver err: %s", testcase.expectedError, err)
+				} else {
+					continue
+				}
+			}  else {
+				t.Fatalf("buildVSphereFromConfig: Should succeed when a valid config is provided: %v", err)
+			}
+		}
+		if testcase.enableDualStackFeature {
+			_ : os.Unsetenv("ENABLE_ALPHA_DUAL_STACK")
```
andrewsykim, line 321: This can use `defer` and put it on line 305

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,113 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+			
+			`,
+			enableDualStackFeature: false,
+			expectedUsername: username,
+			expectedPassword: password,
+			expectedError:    nil,
+		},
+	}
+	for _, testcase := range testcases {
+		t.Logf("Executing Testcase: %s", testcase.testName)
```
andrewsykim, line 289: Use t.Run to wrap the test case instead of logging the name here, something like:
```
t.Run(testcase.testName, func(t *testing.T) {
    ....
})
```

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -198,6 +199,16 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
 		lb = nil
 	}
 
+	// add alpha dual stack feature
+	for tenant := range cfg.VirtualCenter {
+		if len(cfg.VirtualCenter[tenant].IPFamilyPriority) == 2 {
```
andrewsykim, line 204: I think > 1 is more accurate here?

pkg/common/config/consts_and_errors.go
```diff
@@ -66,4 +66,6 @@ var (
 
 	// ErrInvalidIPFamilyType is returned when an invalid IPFamily type is encountered
 	ErrInvalidIPFamilyType = errors.New("Invalid IP Family type")
+
+	ErrMissingDualStackEnvVar = errors.New("Two IP families provides, but dual stack feature is not enabled")
```
andrewsykim, line 70: This is only called in 1 place so don't think a constant is really needed.

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,113 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+		cfg, err := ccfg.ReadCPIConfig([]byte(testcase.conf))
```
andrewsykim, line 290: Instead of `expectedUsername` and `expectedPassword` there should be a field in the test table for the entire expected config

pkg/cloudprovider/vsphere/cloud.go
gwang550: >1 is accurate, so should we also check the case when ipfamilies is greater than 2?

pkg/common/config/consts_and_errors.go
gwang550: Yeah, that's the only place, I can get that removed.

pkg/cloudprovider/vsphere/vsphere_test.go
gwang550: Those two are not used in that test case, going to remove that.

pkg/cloudprovider/vsphere/vsphere_test.go
gwang550: In the testcases, the ipfamily number is not fixed to simulate different use cases. So how should the config be checked in different cases?

andrewsykim:
@gwang550 can you fix the fmt error please?

pkg/common/config/consts_and_errors.go
```diff
@@ -66,4 +66,4 @@ var (
 
 	// ErrInvalidIPFamilyType is returned when an invalid IPFamily type is encountered
 	ErrInvalidIPFamilyType = errors.New("Invalid IP Family type")
-)
+)
```
andrewsykim, line 69: Let's undo this extra line diff here

andrewsykim (commented):
LGTM pending go-fmt fixes

pkg/common/config/consts_and_errors.go
gwang550: Fixed

pkg/cloudprovider/vsphere/vsphere_test.go
```diff
@@ -216,6 +217,98 @@ func TestVSphereLoginByToken(t *testing.T) {
...
+					if err.Error() != testcase.expectedError {
+						t.Fatalf("readConfig: expected err: %s, received err: %s", testcase.expectedError, err)
+					}
+				} else {
+					t.Fatalf("readConfig: unexpected error returned: %v", err)
+				}
+			}
+
+			if testcase.enableDualStackFeature {
+			_:
```
andrewsykim, line 287: Remove this?

andrewsykim (commented):
Can we revert the go modules changes? I don't think this PR needs to bump any dependencies

andrewsykim (commented):
/approve
/lgtm

andrewsykim:
Thanks @gwang550!

PR #395, opened 2020-10-08
optionally check for orphan, unmanaged load balancer service

MartinWeindel:
The load balancer cleanup code of cloud-provider-vsphere is used as dependency in [gardener-extension-provider-vsphere](https://github.com/gardener/gardener-extension-provider-vsphere).
If there are no virtual servers, currently the method `LBProvider.CleanupServices` does not delete the orphan load balancer service (only relevant in the "managed" case). We need the changes in this PR to deal with this corner case.
There is no change in the behaviour of the load balancer provider, if the cloud-provider-vsphere is used directly.

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
+++ b/pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -88,10 +89,10 @@ func (p *lbProvider) doCleanupStep(clusterName string, client clientcorev1.Servi
 		}
 	}
 
-	return p.CleanupServices(clusterName, services)
+	return p.CleanupServices(clusterName, services, false)
 }
 
-func (p *lbProvider) CleanupServices(clusterName string, validServices map[types.NamespacedName]corev1.Service) error {
+func (p *lbProvider) CleanupServices(clusterName string, validServices map[types.NamespacedName]corev1.Service, ensureLBServiceDeleted bool) error {
 	ipPoolIds := sets.NewString()
 	for _, name := range p.classes.GetClassNames() {
 		class := p.classes.GetClass(name)
@@ -99,7 +100,7 @@ func (p *lbProvider) CleanupServices(clusterName string, validServices map[types
 	}
 
 	lbs := map[types.NamespacedName]struct{}{}
-	servers, err := p.access.ListVirtualServers(ClusterName)
+	servers, err := p.access.ListVirtualServers(clusterName)
 	if err != nil {
 		return err
 	}
@@ -162,6 +163,14 @@ func (p *lbProvider) CleanupServices(clusterName string, validServices map[types
 			if err != nil {
 				return err
 			}
+		}
+	}
+
+	// check for orphan unmanaged load balancer service if there are no virtual servers and flag ensureLBServiceDeleted == true
+	if len(lbs) == 0 && ensureLBServiceDeleted {
+		err = p.removeLoadBalancerServiceIfUnused(clusterName)
+		if err != nil && !isNotFoundError(err) {
+			return errors.Wrap(err, "removeLoadBalancerServiceIfUnused failed")
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


dvonthenen:
this looks pretty straightforward. in the process of cutting versioned releases... will merge right afterwards

pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
```diff
@@ -88,18 +89,18 @@ func (p *lbProvider) doCleanupStep(clusterName string, client clientcorev1.Servi
...
-func (p *lbProvider) CleanupServices(clusterName string, validServices map[types.NamespacedName]corev1.Service) error {
+func (p *lbProvider) CleanupServices(clusterName string, validServices map[types.NamespacedName]corev1.Service, ensureLBServiceDeleted bool) error {
 	ipPoolIds := sets.NewString()
 	for _, name := range p.classes.GetClassNames() {
 		class := p.classes.GetClass(name)
 		ipPoolIds.Insert(class.ipPool.Identifier)
 	}
 
 	lbs := map[types.NamespacedName]struct{}{}
-	servers, err := p.access.ListVirtualServers(ClusterName)
```
andrewsykim, line 102: What was `ClusterName` pointing to previously?

pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
```diff
@@ -164,5 +165,13 @@ func (p *lbProvider) CleanupServices(clusterName string, validServices map[types
 			}
 		}
 	}
+
+	// check for orphan unmanaged load balancer service if there are no virtual servers and flag ensureLBServiceDeleted == true
+	if len(lbs) == 0 && ensureLBServiceDeleted {
```
andrewsykim, line 170: Finalizer support should address this use-case and we should pick up finalizer logic in the service controller (see https://github.com/kubernetes/kubernetes/pull/85023) once we bump our kubernetes depedencies to v1.18. You think we should wait for that or add custom logic now?

andrewsykim (commented):
Sorry for the delay reviewing this

pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
andrewsykim: In other words, I _think_ this should be fixed automatically when we release v1.18.0 but can't say without testing it fully

pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
MartinWeindel: `ClusterName` is the global variable which is set by by the `--cluster-name` flag.
(see https://github.com/kubernetes/cloud-provider-vsphere/blob/44a75efaef85b259f75a2cf729b60149dbb9353f/cmd/vsphere-cloud-controller-manager/main.go#L93-L95)
This change makes only a difference, if the function CleanupServices is used externally. Otherwise the passed `clusterName` is the same as before
(see https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/cloud.go#L118)

pkg/cloudprovider/vsphere/loadbalancer/cleanup.go
MartinWeindel: We saw this problem only very rarely. So I agree, the finalizer should fix the problem in most cases, but it will be hard to prove it by testing.
And there is still a minor chance, that something goes wrong. The LB is deleted, as soon as the last virtual server has been deleted. If something went wrong with deleting the LB service, it will not be tried a second time without this custom logic.

andrewsykim: > The LB is deleted, as soon as the last virtual server is being deleted. If something went wrong with deleting the LB service, it will not be tried a second time without this custom logic.

So even if the Service deletion is blocked by the finalizer, there's still a corner case where failure to delete an LB Service will leave it dangling as well? Wouldn't deletion of the LB Service also be gated by the finalizer?

MartinWeindel: A K8s service is representated by one or multiple virtual server. With the finalizer you can only assure that all related virtual servers are deleted. But if the LB service is managed, it has to be deleted after the last virtual server has gone. You would need some kind of a transaction to enforce the deletion of LB service with the last virtual server. 
Here an example, how the LB service may leak without this check:
1. The `EnsureLoadBalancerDeleted` method is called for the last service.
  a) The last virtual server is deleted successfully.
  b) Deleting the LB service fails for some reason.
  c) The finalizer is *not* removed.
2. The `EnsureLoadBalancerDeleted` method is called for the last service a second time
  a) All virtual servers have already been deleted, nothing to do
  b) The finalizer is removed

MartinWeindel:
@andrewsykim 
friendly ping: what is the status of your review?

andrewsykim:
@lubronzhan @nicolehanjing can you review this please?

PR #400, opened 2020-11-05
Add test for dual stack with internal/external subnet and fix for not discovering the ip when specify internal/external with ipv4/ipv6

gwang550:
This PR adds additional tests for dual stack feature. Mainly for uses case with internal/external subnets

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager_test.go
+++ b/pkg/cloudprovider/vsphere/nodemanager_test.go
@@ -187,7 +187,8 @@ func TestAlphaDualStack(t *testing.T) {
 
 	ipv4Ips := returnIPsFromSpecificFamily(vcfg.IPv4Family, ips)
 	size := len(ipv4Ips)
-	if size != 1 {
+	// size need to equal to 2, one for internal and one for external, in this case they are the same
+	if size != 2 {
 		t.Errorf("Should only return single IPv4 address. expected: 1, actual: %d", size)
 	} else if !strings.EqualFold(ipv4Ips[0], ipv4Ip) {
 		t.Errorf("IPv6 does not match. expected: %s, actual: %s", ipv4Ip, ipv4Ips[0])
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -201,6 +202,96 @@ func TestAlphaDualStack(t *testing.T) {
...
+	nm.RegisterNode(node)
+	err = nm.DiscoverNode(name, cm.FindVMByName)
+	if err != nil {
+		t.Errorf("Failed DiscoverNode: %s", err)
+	}
+
+	nodeList := make([]*pb.Node, 0)
+	_ = nm.ExportNodes("", "", &nodeList)
+
+	ips := nodeList[0].Addresses
```
andrewsykim, line 276: Need to compare the AdressType in this test, to check if the right subnets were assigned InternalIP and ExternalIP

gwang550: There isn't a property where I can check the type of the ips. When the addresses are added, it's type agnostic as you can see [here.](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/nodemanager.go#L390)

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -184,6 +185,97 @@ func TestAlphaDualStack(t *testing.T) {
...
+
+	// get node registered so node can be exported
+	nm.RegisterNode(node)
+	err = nm.DiscoverNode(name, cm.FindVMByName)
+	if err != nil {
+		t.Errorf("Failed DiscoverNode: %s", err)
+	}
+
+	nodeList := make([]*pb.Node, 0)
+	_ = nm.ExportNodes("", "", &nodeList)
```
andrewsykim, line 275: Updating datacenterToNodeList makes this test pass since we're calling `ExportNodes`.  But does the node registration code path call this? IIRC this is only called by the internal gRPC server which is not on by default.

andrewsykim: IIRC the gRPC server was an experimental feature that was never turned on by default.

gwang550: My understanding of the purpose of calling `ExportNodes` is to get the Nodelist so I can check against the test case. I don't think that the node registration code path calls this function. I only found one use case for `ExportNodes` which is [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/3058adc81bd0a3aa59517b3fb2160f04a81bb0ac/pkg/cloudprovider/vsphere/server/server.go#L87).

dvonthenen: @andrewsykim That is correct. The code is there to support a gRPC endpoint which is experimental and is disabled by default.

PR #403, opened 2020-11-11
remove latest tagging during releases

gwang550:
This PR fixes a problem which a newly built release is always tagged as the latest, which causes confusion and it's error-prone for the alpha releases.

**Release note**:

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,7 +287,7 @@ release: | $(DOCKER_SOCK)
 ################################################################################
 .PHONY: release-push
 release-push: | $(DOCKER_SOCK)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -287,7 +287,7 @@ release: | $(DOCKER_SOCK)
 ################################################################################
 .PHONY: release-push
 release-push: | $(DOCKER_SOCK)
-	hack/release.sh -p -l
```
andrewsykim, line 290: Also remove the `-l` option from the release script.

PR #407, opened 2020-11-17
Add route interfaces

### LINKED ISSUE #404: Pod route exchange with NSX-T
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
When adding node to NSX-T networking, container data plane will be associated with NSX-T encapsulation protocols.

**What you expected to happen**:
Exchange routes with NSX-T tier 1 router, so that:
  - a non-encapsulation transit mode can be used when hosts are not layer-2 adjacent and must have traffic forwarded by a router between them, and/or
  - pods IP addresses are routable and accessible externally from the cluster.

**How to reproduce it (as minimally and precisely as possible)**:
With NSX-T connection info and Tier1 router UUID config, vsphere cloud provider will add static route in T1 router when there is a node added event, and will delete related static route for a node deleted event.

**Anything else we need to know?**:
N/A

**Environment**:
- vsphere-cloud-controller-manager version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

dantingl:
TODO: use NSXT config based on #405

Add static routes on NSXT router for pod CIDR.

fixes #404 

**Release note**:

```release-note
```


andrewsykim (commented):
/ok-to-test

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -216,14 +228,21 @@ func buildVSphereFromConfig(cfg *ccfg.CPIConfig, lbcfg *lcfg.LBConfig) (*VSphere
...
-		loadbalancer: lb,
-		instances:    newInstances(nm),
-		zones:        newZones(nm, cfg.Labels.Zone, cfg.Labels.Region),
-		server:       server.NewServer(cfg.Global.APIBinding, nm),
+		cfg:           cfg,
+		cfgLB:         lbcfg,
+		cfgRoute:      routecfg,
+		nodeManager:   nm,
+		loadbalancer:  lb,
+		routeProvider: routeProvider,
```
andrewsykim, line 242: nit: just name this `routes` to be consistent with other fields

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (p *RouteProvider) DeleteRoute(ctx context.Context, clusterName string, route *cloudprovider.Route) error {
+	klog.V(6).Infof("Deleting static route %s on router %s in cluster %s",
+		route.Name, p.routerPath, clusterName)
+	p.broker.tier1StaticRoutesClient.Delete(p.routerPath, route.Name)
+	return nil
+}
+
+func (p *RouteProvider) GetNodeIPAddress(nodeName string) (string, error) {
+	nodes, err := p.kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
andrewsykim, line 113: This should probably call a node lister and fetch from an informer cache.

andrewsykim: Also, this can be a `Get` since we know the node name

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,144 @@
...
+		Tags:     tags,
+	}
+	p.broker.tier1StaticRoutesClient.Patch(p.routerPath, routerId, staticRoute)
+	return nil
+}
+
+func (p *RouteProvider) DeleteRoute(ctx context.Context, clusterName string, route *cloudprovider.Route) error {
+	klog.V(6).Infof("Deleting static route %s on router %s in cluster %s",
+		route.Name, p.routerPath, clusterName)
+	p.broker.tier1StaticRoutesClient.Delete(p.routerPath, route.Name)
```
andrewsykim, line 108: This doesn't return an error?

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,144 @@
...
+	tags = append(tags, model.Tag{Scope: &clusterNameScope, Tag: &clusterName})
+	tags = append(tags, model.Tag{Scope: &nodeNameScope, Tag: &nodeName})
+	var nexthops []model.RouterNexthop
+	nexthops = append(nexthops, model.RouterNexthop{IpAddress: &nodeIp})
+	staticRoute := model.StaticRoutes{
+		Network:  route.DestinationCIDR,
+		NextHops: nexthops,
+		Tags:     tags,
+	}
+	p.broker.tier1StaticRoutesClient.Patch(p.routerPath, routerId, staticRoute)
```
andrewsykim, line 101: Does this return an error?

andrewsykim:
Fixing integration tests here https://github.com/kubernetes/cloud-provider-vsphere/pull/412

andrewsykim:
Fixed in https://github.com/kubernetes/cloud-provider-vsphere/pull/412, can you rebase please?

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,182 @@
...
+		klog.Errorf("Deleting static route %s failed: %s", route.Name, err)
+		return err
+	}
+	return nil
+}
+
+// getNodeIPAddress gets node IP address
+// The order is to choose node internal IP first, then external IP.
+func (p *RouteProvider) getNodeIPAddress(nodeName string) (string, error) {
+	node, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
andrewsykim, line 136: Can we pass in a node informer cache to `RouterProvider` and have this call get from cache? An API call for every route reconcile will get expensive quickly.

andrewsykim: You can pass a shared informer through `Initialize`. Or alternatively we can start the cache informer there and worry about using shared informers in a follow-up PR.

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,182 @@
+package route
```
andrewsykim, line 1: Missing license header

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,182 @@
...
+	}, nil
+}
+
+// Initialize sets kubernetes client in RouteProvider
+func (p *RouteProvider) Initialize(client clientset.Interface) {
+	p.kubeClient = client
+}
+
+// ListRoutes returns a list of routes which have static routes on NSXT
+func (p *RouteProvider) ListRoutes(ctx context.Context, clusterName string) ([]*cloudprovider.Route, error) {
```
andrewsykim, line 49: Any chance we can add some unit tests for this method as well as `CreateRoute` and `DeleteRoute`?

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,182 @@
...
+		for _, resource := range list.Results {
+			if resource.IntentPaths[0] == path && resource.State == config.RealizedState {
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("Waiting for static route %s is timeout", path)
+}
+
+// getRouterId returns router ID from router path
```
andrewsykim, line 178: If we only care about the router ID can we just have the config require router ID instead of the path?

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: OK, let me try to mock NSXAPI response

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: OK, let me check. 
Actually we only need get node when route controller calls CreateRoute(), which happens when there is a new node added. In every route reconcile basically it only calls ListRoutes()

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: I'm thinking if we support Tier0 router later, we could use path to identify it is Tier1 or Tier0 router

pkg/cloudprovider/vsphere/route/routeprovider.go
andrewsykim: > Actually we only need get node when route controller calls CreateRoute(), which happens when there is a new node added. In every route reconcile basically it only calls ListRoutes(

CreateRoute would also be called if ListRoutes doesn't return the route of a node, which can happen if `CreateRoute` errors repeatedly. I agree though that `CreateRoute` is generally not called as frequently as `ListRoutes`. Maybe this can be done in a follow-up PR. 

I think it's worthwhile to consider rate limiting or cachingi `ListRoutes` results though, as that can be an expensive API call, especially when done frequently by the routes controller.

pkg/cloudprovider/vsphere/route/routeprovider.go
andrewsykim: noted, thanks

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: Sure, make sense to me. Based on my test, router controller calls ListRoutes every 10 seconds. We can have a cache in cloud provider to avoid frequent NSXAPI call.

andrewsykim: sgtm, I'd be okay to add caching support in a follow-up, but should be done prior to an official release I think.

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,198 @@
...
+func (p *RouteProvider) getNodeIPAddress(nodeName string) (string, error) {
+	node, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Getting node %s failed: %v", nodeName, err)
+		return "", err
+	}
+
+	addresses := make(map[corev1.NodeAddressType]string)
+	for _, addr := range node.Status.Addresses {
+		addresses[addr.Type] = addr.Address
```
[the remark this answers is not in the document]
dantingl: Thanks, I should consider the dual-stack.
I will use the first `InternalIP`, if there is no `InternalIP`, use the first `ExternalIP`

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,212 @@
...
+	"context"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/vmware/vsphere-automation-sdk-go/runtime/data"
+	"github.com/vmware/vsphere-automation-sdk-go/services/nsxt/model"
```
nicolehanjing, line 29: it might be good to add an empty line here to separate GitHub import / k8s.io import

andrewsykim (commented):
Oops, this was left pending for a few days, forgot to submit it

pkg/cloudprovider/vsphere/route/routeprovider.go
andrewsykim: +1

pkg/cloudprovider/vsphere/route/config/consts_and_errors.go
```diff
@@ -0,0 +1,35 @@
...
+
+package config
+
+import (
+	"time"
+)
+
+const (
+	// ClusterNameTagScope is the scope of clusterName tag
+	ClusterNameTagScope = "vspherecloudprovider/clusterName"
```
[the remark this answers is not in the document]
dantingl: The Tag is in <key:value> format, scope is the key and node name will be the value. For example if a node name is "node1", the tag will be "vspherecloudprovider/clusterName": "node1". Let me add more discription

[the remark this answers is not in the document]
andrewsykim: Thanks, I was specificalyl referring to the tag key `vspherecloudprovider/clusterName`. Wondering if it should be something like `vsphere.k8s.io/cluster-name` or something else. We should be careful about using k8s.io prefixes though since this is technically not an official Kubernetse API.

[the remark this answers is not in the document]
dantingl: I added `vspherecloudprovider` in the scope because I thought it would be better to add the owner in the tag, so other user could know the static routes are create by it. I'm fine to use `vsphere.k8s.io`.
I checked LB tags, looks like it just use `cluster` as the scope, do you think I should follow the same?

```
const (
	// ScopeOwner is the owner scope
	ScopeOwner = "owner"
	// ScopeCluster is the cluster scope
	ScopeCluster = "cluster"
	// ScopeService is the service scope
	ScopeService = "service"
	// ScopePort is the port scope
	ScopePort = "port"
	// ScopeIPPoolID is the IP pool id scope
	ScopeIPPoolID = "ippoolid"
	// ScopeLBClass is the load balancer class scope
	ScopeLBClass = "lbclass"
)
```

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: thanks

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,212 @@
...
+	node, err := p.kubeClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("getting node %s failed: %v", nodeName, err)
+		return "", err
+	}
+
+	allIPs := make([]net.IP, 0, len(node.Status.Addresses))
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == v1.NodeInternalIP {
+			ip := net.ParseIP(addr.Address)
```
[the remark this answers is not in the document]
dantingl: I'm not sure if it is necessary. For dual-stack case, both IPv4 & IPv6 should be reachable. But checking IP family of pod CIDR should be more reliable.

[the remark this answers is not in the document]
dantingl: I updated patch to use Node IP which has the same IP family with podCIDR, but I thought K8s cluster should be fine if node IP and podCIDR are using different IP family.
I did manually test on NSXT, actually the limitation is on NSXT side that static route only allows network and nexthop are using the same IP family.

pkg/cloudprovider/vsphere/route/routeprovider.go
dantingl: Just saw we already have an informer for node manager, maybe I can add logic in `nodeAdded` and `nodeDeleted` to call some route interfaces to cache node info

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestReadINIConfig(t *testing.T) {
+	contents := `
+[Route]
+router-path = /infra/tier-1s/test-router
+
+[NSXT]
+user = admin
+password = secret
+host = nsxt-server
```
andrewsykim, line 36: can we name this field `server` instead of `host`?

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
```diff
@@ -0,0 +1,53 @@
...
+	When the INI based cloud-config is deprecated. This file should be deleted.
+*/
+
+func TestReadINIConfig(t *testing.T) {
+	contents := `
+[Route]
+router-path = /infra/tier-1s/test-router
+
+[NSXT]
+user = admin
```
andrewsykim, line 34: Would it be possible to support `secretName` and `secretNamespace` here in the same way we do for vCenter credentials? I think many users require all credentials are stored in secrets, sometimes the cloud config itself is stored in a ConfigMap.

andrewsykim: Here's one example https://github.com/kubernetes/cloud-provider-vsphere/blob/22e85632bfe62f9a6afff8d2fc269f80eb9e7c78/pkg/common/config/config_yaml_test.go#L52-L53 

It would require more plumbing on the generic NSX-T package to read the secret from an informer cache

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
dantingl: `[NSXT]` session was introduced by LB change, I guess there will be an upgrade issue if we change to use `server`. Or we just introduce a new field `server`?

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
dantingl: Sure, I think we can support secret to store password or certificates. I see there is a `CredentialManager`, let me see if I can leverage it for NSX-T as well.

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
andrewsykim: Oh good point, lets just leave it as host for this PR then

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
dantingl: Could I add NSX-T related cert in a following PR?

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
dantingl: Sure, thanks

pkg/cloudprovider/vsphere/route/config/config_ini_legacy_test.go
andrewsykim: yeah I'm fine with that

andrewsykim:
This getting close to merge, just some minor nits at this point.

pkg/cloudprovider/vsphere/route/nsxtbroker.go
```diff
@@ -0,0 +1,82 @@
...
+	}, nil
+}
+
+func (b *nsxtBroker) QueryEntities(queryParam string) (model.SearchResponse, error) {
+	queryParam = strings.ReplaceAll(queryParam, "/", "\\/")
+	return b.queryClient.List(queryParam, nil, nil, nil, nil, nil)
+}
+
+func (b *nsxtBroker) CreateStaticRoute(routerPath string, routeID string, staticRoute model.StaticRoutes) error {
+	routerID := getRouterID(routerPath)
```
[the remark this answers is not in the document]
dantingl: The parameter `routeID` is static route ID. Maybe I should rename the parameter as `staticRouteID`

pkg/cloudprovider/vsphere/route/nsxtbroker.go
```diff
@@ -0,0 +1,82 @@
...
+	return b.tier1StaticRoutesClient.Patch(routerID, routeID, staticRoute)
```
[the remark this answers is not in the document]
dantingl: The first `routerID` is tier1 router ID. The second one `routeID` is static route ID.

pkg/cloudprovider/vsphere/route/nsxtbroker.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (b *nsxtBroker) DeleteStaticRoute(routerPath string, routeID string) error {
+	routerID := getRouterID(routerPath)
```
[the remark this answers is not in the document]
dantingl: Let me rename `routeID` as `staticRouteID`

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,242 @@
...
+	path := p.routerPath + "/static-routes/" + routeID
+	limit := time.Now().Add(config.RealizedStateTimeout)
+	for time.Now().Before(limit) {
+		time.Sleep(config.RealizedStateSleepTime)
+		list, err := p.broker.ListRealizedEntities(path)
+		if err != nil {
+			return fmt.Errorf("get route %s realized state failed: %s", path, err)
+		}
+		for _, resource := range list.Results {
+			if resource.IntentPaths[0] == path && *resource.State == config.RealizedState {
```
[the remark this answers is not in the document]
dantingl: Sure, thanks

andrewsykim (commented):
/approve

I have a few comments but I think we can iterate on them going forward. I think this is good to merge now. Leaving lgtm to @nicolehanjing 

cc @dvonthenen if you wanted to give this a review as well.

pkg/cloudprovider/vsphere/route/config/config_ini_legacy.go
```diff
@@ -0,0 +1,89 @@
...
+	TODO:
+	When the INI based cloud-config is deprecated. This file should be deleted.
+*/
+
+// CreateConfig generates a common Config object based on what other structs and funcs
+// are already dependent upon in other packages.
+func (rci *RouteConfigINI) CreateConfig() *Config {
+	cfg := &Config{}
+	cfg.Route.RouterPath = rci.Route.RouterPath
+	//NSXT
```
nicolehanjing, line 36: maybe something like "NSXT configurations"?

pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -0,0 +1,245 @@
...
+	queryParam := fmt.Sprintf("resource_type:StaticRoutes AND tags.scope:%s AND tags.tag:%s",
+		config.ClusterNameTagScope, clusterName)
+	staticRoutes, err := p.broker.QueryEntities(queryParam)
+	if err != nil {
+		klog.Errorf("querying static routes for cluster %s failed", clusterName)
+		return nil, err
+	}
+
+	var routes []*cloudprovider.Route
+	for _, item := range staticRoutes.Results {
```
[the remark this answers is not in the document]
nicolehanjing: agreed, learned it 👍🏻

nicolehanjing:
Left a few minor comments, overall /lgtm to me
Thanks @dantingl :)

PR #416, opened 2020-12-04
Change CPI to use klog/v2 (Structured logging)

dvonthenen:
This switches the CPI to use klog/v2 per this thread:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-dev/3d51Jg59ZbI/azv8Hlj5AQAJ

Implements this KEP:
https://github.com/kubernetes/enhancements/pull/1617

NA

Tested using Kubernetes 1.18.12 and vSphere 7.0

**Release note**:
NA

### DIFF
--- a/pkg/common/vclib/connection.go
+++ b/pkg/common/vclib/connection.go
@@ -28,7 +28,7 @@ import (
 	"github.com/vmware/govmomi/sts"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/soap"
-	"k8s.io/klog"
+	klog "k8s.io/klog/v2"
 )
 
 // VSphereConnection contains information for connecting to vCenter
@@ -180,12 +180,6 @@ func (connection *VSphereConnection) NewClient(ctx context.Context) (*vim25.Clie
 	err = connection.login(ctx, client)
 	if err != nil {
 		return nil, err
-	}
-	if klog.V(3) {
-		s, err := session.NewManager(client).UserSession(ctx)
-		if err == nil {
-			klog.Infof("New session ID for '%s' = %s", s.UserName, s.Key)
-		}
 	}
 
 	if connection.RoundTripperCount == 0 {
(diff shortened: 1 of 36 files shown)


pkg/common/vclib/connection.go
```diff
@@ -181,11 +181,10 @@ func (connection *VSphereConnection) NewClient(ctx context.Context) (*vim25.Clie
...
 	}
-	if klog.V(3) {
-		s, err := session.NewManager(client).UserSession(ctx)
-		if err == nil {
-			klog.Infof("New session ID for '%s' = %s", s.UserName, s.Key)
-		}
+
+	s, err := session.NewManager(client).UserSession(ctx)
+	if err == nil {
+		klog.V(3).Infof("New session ID for '%s' = %s", s.UserName, s.Key)
```
andrewsykim, line 187: Can we just remove this log line?

andrewsykim (commented):
lgtm overall, I recall there were some breaking changes in v2 around the way stdio and log files w ere handled. Anyway we can validate if that all works?

pkg/common/vclib/connection.go
dvonthenen: yup, I'm ok with that too

pkg/common/vclib/connection.go
```diff
@@ -181,12 +181,6 @@ func (connection *VSphereConnection) NewClient(ctx context.Context) (*vim25.Clie
 	if err != nil {
 		return nil, err
 	}
-	if klog.V(3) {
-		s, err := session.NewManager(client).UserSession(ctx)
```
andrewsykim, line 185: Is it safe to actually remove this call, I was just meant the log line previously but if `s` is not used anywhere this seems fine.

dvonthenen: yep, it's completely safe to remove. if anything, I don't think we should be printing the session key in the logs anyways. it could be used to hijack a VC session.

andrewsykim (commented):
/lgtm 
/hold

pkg/common/vclib/connection.go
dvonthenen: verified that CPI behaves correctly removing these lines

dvonthenen:
Verified latest changes are good. Removing hold.

PR #421, opened 2021-01-05
Find node connected router automatically

dantingl:

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/route/config/consts_and_errors.go
+++ b/pkg/cloudprovider/vsphere/route/config/consts_and_errors.go
@@ -39,4 +39,9 @@ const (
 
 	// DisplayNameMaxLength is the maximum length of static route display name
 	DisplayNameMaxLength = 255
+
+	// ProviderIDPrefix is the prefix of node's provider ID
+	ProviderIDPrefix = "vsphere://"
+	// BiosUuidPrefix is the prefix of VM BIOS UUID
+	BiosUuidPrefix = "biosUuid:"
 )
--- a/pkg/common/kubernetes/informers.go
+++ b/pkg/common/kubernetes/informers.go
@@ -23,6 +23,8 @@ import (
 	"syscall"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	listerv1 "k8s.io/client-go/listers/core/v1"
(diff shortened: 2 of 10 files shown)


pkg/cloudprovider/vsphere/route/config/consts_and_errors.go
```diff
@@ -39,4 +39,9 @@ const (
 
 	// DisplayNameMaxLength is the maximum length of static route display name
 	DisplayNameMaxLength = 255
+
+	// ProviderIDPrefix is the prefix of node's provider ID
+	ProviderIDPrefix = "vsphere://"
```
dvonthenen, line 44: We should probably consolidate this const with the other one defined so we only need to update in a single location should this ever change:
https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/util.go#L31

Would also be in favor of putting the bios uuid prefix in the same location

pkg/common/kubernetes/informers.go
```diff
@@ -104,3 +106,9 @@ func (im *InformerManager) AddNodeListener(add, remove func(obj interface{}), up
 func (im *InformerManager) Listen() {
 	go im.informerFactory.Start(im.stopCh)
 }
+
+// GetNodeList gets node list
+func (im *InformerManager) GetNodeList() ([]*v1.Node, error) {
+	nodeLister := im.informerFactory.Core().V1().Nodes().Lister()
```
dvonthenen, line 112: Is it possible to reuse the node list capture below instead of creating a new informer with a separate list? or is it a chicken/egg initialization thing?
https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/nodemanager.go#L56

dvonthenen (commented):
Left a couple of questions...

pkg/cloudprovider/vsphere/route/config/consts_and_errors.go
dantingl: Thanks for letting me know this, I should use `ProviderIDPrefix` from vsphere/util.go.
For `BiosUuidPrefix` actually it is a prefix from NSXT object, maybe I should put it in nsxt directory?

pkg/common/kubernetes/informers.go
dantingl: I think `nodeRegUUIDMap` is still empty when initializing route provider. If it is not good to add a func in InformerManager, I can move it into RouteProvider `Initialize`

andrewsykim (commented):
/ok-to-test

PR #423, opened 2021-01-11
Get NSXT username/password from secret

dantingl:

**Release note**:

```release-note
```

### DIFF
--- a/pkg/nsxt/config/config_ini_legacy.go
+++ b/pkg/nsxt/config/config_ini_legacy.go
@@ -16,10 +16,39 @@
 
 package config
 
-import "errors"
+import (
+	"errors"
+	"fmt"
 
-// ValidateConfig checks NSXT configurations
-func (cfg *NsxtConfigINI) ValidateConfig() error {
+	"gopkg.in/gcfg.v1"
+)
+
+/*
+	TODO:
+	When the INI based cloud-config is deprecated. This file should be deleted.
+*/
+
+// CreateConfig generates a common Config object based on what other structs and funcs
+// are already dependent upon in other packages.
+func (nci *NsxtConfigINI) CreateConfig() *Config {
+	cfg := &Config{}
+	cfg.User = nci.NSXT.User
+	cfg.Password = nci.NSXT.Password
+	cfg.Host = nci.NSXT.Host
+	cfg.InsecureFlag = nci.NSXT.InsecureFlag
+	cfg.VMCAccessToken = nci.NSXT.VMCAccessToken
+	cfg.VMCAuthHost = nci.NSXT.VMCAuthHost
+	cfg.ClientAuthCertFile = nci.NSXT.ClientAuthCertFile
+	cfg.ClientAuthKeyFile = nci.NSXT.ClientAuthKeyFile
+	cfg.CAFile = nci.NSXT.CAFile
+	cfg.SecretName = nci.NSXT.SecretName
+	cfg.SecretNamespace = nci.NSXT.SecretNamespace
+
+	return cfg
+}
+
+// validateConfig checks NSXT configurations
+func (cfg *NsxtINI) validateConfig() error {
 	if cfg.VMCAccessToken != "" {
 		if cfg.VMCAuthHost == "" {
 			return errors.New("vmc auth host must be provided if auth token is provided")
@@ -36,6 +65,14 @@ func (cfg *NsxtConfigINI) ValidateConfig() error {
 		if cfg.ClientAuthKeyFile == "" {
 			return errors.New("client key file is required if client cert file is provided")
 		}
+	} else if cfg.SecretName != "" {
+		if cfg.SecretNamespace == "" {
+			return errors.New("secret namespace is required if secret name is provided")
+		}
+	} else if cfg.SecretNamespace != "" {
+		if cfg.SecretName == "" {
+			return errors.New("secret name is required if secret namespace is provided")
+		}
 	} else {
 		return errors.New("user or vmc access token or client cert file must be set")
 	}
@@ -43,4 +80,38 @@ func (cfg *NsxtConfigINI) ValidateConfig() error {
 		return errors.New("host is empty")
 	}
 	return nil
+}
+
+// CompleteAndValidate sets default values, overrides by env and validates the resulting config
+func (nci *NsxtConfigINI) CompleteAndValidate() error {
+	return nci.NSXT.validateConfig()
+}
+
+// ReadRawConfigINI parses vSphere cloud config file and stores it into ConfigINI
+func ReadRawConfigINI(configData []byte) (*NsxtConfigINI, error) {
+	if len(configData) == 0 {
+		return nil, fmt.Errorf("Invalid INI file")
+	}
+
+	cfg := &NsxtConfigINI{}
+
+	if err := gcfg.FatalOnly(gcfg.ReadStringInto(cfg, string(configData))); err != nil {
+		return nil, err
+	}
+
+	err := cfg.CompleteAndValidate()
+	if err != nil {
+		return nil, err
+	}
+	return cfg, nil
+}
+
+// ReadConfigINI parses vSphere cloud config file and stores it into Config
+func ReadConfigINI(configData []byte) (*Config, error) {
+	cfg, err := ReadRawConfigINI(configData)
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg.CreateConfig(), nil
 }
--- a/pkg/nsxt/config/types_ini_legacy.go
+++ b/pkg/nsxt/config/types_ini_legacy.go
@@ -16,8 +16,13 @@
 
 package config
 
-// NsxtConfigINI contains the NSX-T specific configuration
+// NsxtConfigINI  is used to read and store information from the cloud configuration file
 type NsxtConfigINI struct {
+	NSXT NsxtINI `gcfg:"nsxt"`
+}
+
+// NsxtINI contains the NSX-T specific configuration
+type NsxtINI struct {
 	// NSX-T username.
 	User string `gcfg:"user"`
 	// NSX-T password in clear text.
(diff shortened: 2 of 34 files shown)


pkg/nsxt/config/types_ini_legacy.go
```diff
@@ -34,4 +34,7 @@ type NsxtConfigINI struct {
 	ClientAuthCertFile string `gcfg:"client-auth-cert-file"`
 	ClientAuthKeyFile  string `gcfg:"client-auth-key-file"`
 	CAFile             string `gcfg:"ca-file"`
+
```
nicolehanjing, line 37: might be better to also add comments here
```
// SecretName is the secret name for NSX-T username and password
```

pkg/nsxt/config/types_yaml.go
```diff
@@ -34,4 +34,7 @@ type NsxtConfigYAML struct {
 	ClientAuthCertFile string `yaml:"clientAuthCertFile"`
 	ClientAuthKeyFile  string `yaml:"clientAuthKeyFile"`
 	CAFile             string `yaml:"caFile"`
+
```
nicolehanjing, line 37: same here, add comments

pkg/nsxt/nsxt_connector.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/protocol/client"
 	"github.com/vmware/vsphere-automation-sdk-go/runtime/security"
 	"k8s.io/cloud-provider-vsphere/pkg/nsxt/config"
+	"k8s.io/klog"
```
nicolehanjing, line 33: should we use `k8s.io/klog/v2` instead?

dantingl: I see other files are using `k8s.io/klog`, do we plan to use v2 instead of v1?

dvonthenen: klog v2 was just merged. We should avoid mixing between the two. Mixing versions would also require a different init for both libraries. Easier just to avoid it.

andrewsykim: +1 to use klog/v2

pkg/nsxt/nsxt_secret.go
```diff
@@ -0,0 +1,152 @@
...
+		klog.Warningf("NSXT username and password should be both provided in secret")
+		return
+	}
+	klog.V(6).Infof("Updating security context for NSXT connection")
+	securityCtx := core.NewSecurityContextImpl()
+	securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.USER_PASSWORD_SCHEME_ID)
+	securityCtx.SetProperty(security.USER_KEY, user)
+	securityCtx.SetProperty(security.PASSWORD_KEY, password)
+	for _, connector := range nsm.connectors {
+		connector.SetSecurityContext(securityCtx)
```
andrewsykim, line 142: Do you think we need to support per-host Secrets? Or do you think it's sufficient to assume a Secret is valid for every  NSX host?

pkg/nsxt/nsxt_secret.go
```diff
@@ -0,0 +1,152 @@
+/*
```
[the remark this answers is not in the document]
nicolehanjing: +1 for renaming

pkg/nsxt/nsxt_connector.go
dantingl: Got it, thanks

pkg/nsxt/nsxt_connector.go
```diff
@@ -68,18 +69,13 @@ func NewNsxtConnector(nsxtConfig *config.NsxtConfig) (client.Connector, error) {
...
-			if nsxtConfig.Password == "" {
-				return nil, fmt.Errorf("password must be provided")
-			}
-
+		} else if nsxtConfig.User != "" && nsxtConfig.Password != "" {
 			securityCtx.SetProperty(security.AUTHENTICATION_SCHEME_ID, security.USER_PASSWORD_SCHEME_ID)
 			securityCtx.SetProperty(security.USER_KEY, nsxtConfig.User)
 			securityCtx.SetProperty(security.PASSWORD_KEY, nsxtConfig.Password)
+		} else if nsxtConfig.SecretName != "" && nsxtConfig.SecretNamespace != "" {
+			klog.V(6).Infof("Will get username and password from secret")
```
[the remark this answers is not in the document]
dantingl: OK

pkg/nsxt/nsxt_secret.go
```diff
@@ -0,0 +1,152 @@
...
+
+// SecretManager is used to manage NSXT secret
+type SecretManager struct {
+	secretName      string
+	secretNamespace string
+	connectors      []*client.RestConnector
+}
+
+// secretManager is a singleton for all NSXT secret consumers
+var secretManager *SecretManager
```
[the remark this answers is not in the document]
dantingl: Yes, only `NewNsxtConnector` calls `NewSecretManager`, but Route provider and LB provider will both call `NewNsxtConnector`. 
Right now loading nsxt config is inside LB and Route provider, so we need a singleton to update connectors for both of them

pkg/nsxt/nsxt_secret.go
dantingl: I don't think we could assume user will set the same user/password for all NSX hosts. 
For each vsphere-cloud-provider, it only has one NSXT config with one Secret, right? Or do you mean we support multiple NSXT config?

pkg/nsxt/nsxt_secret.go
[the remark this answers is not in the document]
dantingl: Maybe ideally NSXT config should be loaded and initialized in VSphere's `buildVSphereFromConfig` and `Initialize` separately, but for the time being I thought a singleton could work :)

pkg/nsxt/nsxt_connector.go
```diff
@@ -68,18 +69,13 @@ func NewNsxtConnector(nsxtConfig *config.NsxtConfig) (client.Connector, error) {
...
+			NewSecretManager(nsxtConfig.SecretName, nsxtConfig.SecretNamespace)
```
andrewsykim, line 78: We should be handling the returned value here

pkg/nsxt/nsxt_secret.go
andrewsykim: We support multiple vCenters if a cluster could possibly have VMs from mulitple vCenters. Is that scenario likely or possible with NSX? Seems less likely but just double checking

pkg/nsxt/nsxt_secret.go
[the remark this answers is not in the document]
andrewsykim: I'd like to avoid singletons in possible, especially if it's managed through a global state. I do think it's problematic that we are calling `NewNsxtConnector` twice, do you think we can refactor the code to call it once and pass a single instance of the connecter to both the LB and routes controllers?

pkg/nsxt/nsxt_secret.go
dantingl: To have K8s nodes in the same overlay subnet, we should attach VM vNIC to the same segment in one NSX. For VMs in different vCenters, as long as their ESX hosts register to the same NSX, it could still work. So I think every K8s cluster has only one NSX.

pkg/nsxt/nsxt_secret.go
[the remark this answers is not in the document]
dantingl: Calling `NewNsxtConnector` twice because I planed to use different connections with NSXT for Route and LB. But maybe it's better to share the same connector. 
Then in this case connector and secret_manager should be called once, let me try to move NSXT init logic in VSphere directly, and pass them to Route and LB provider.

pkg/nsxt/nsxt_connector.go
andrewsykim, line 33: Use klog v2

pkg/nsxt/nsxt_connector.go
andrewsykim, line 77: Remove this log line

pkg/nsxt/config/config_ini_legacy.go
```diff
@@ -36,6 +65,14 @@ func (cfg *NsxtConfigINI) ValidateConfig() error {
...
+	} else if cfg.SecretName != "" {
+		if cfg.SecretNamespace == "" {
+			return errors.New("secret namespace is required if secret name is provided")
+		}
+	} else if cfg.SecretNamespace != "" {
+		if cfg.SecretName == "" {
+			return errors.New("secret name is required if secret namespace is provided")
+		}
 	} else {
 		return errors.New("user or vmc access token or client cert file must be set")
```
andrewsykim, line 77: this error message needs updating

andrewsykim: Can be done in a follow-up PR

pkg/nsxt/connector_manager.go
```diff
@@ -46,8 +58,13 @@ func (processor remoteBasicAuthHeaderProcessor) Process(req *http.Request) error
 	return nil
 }
 
-// NewNsxtConnector creates a new NSXT connector
-func NewNsxtConnector(nsxtConfig *config.NsxtConfig) (client.Connector, error) {
+// NewConnectorManager creates a new NSXT connector
+func NewConnectorManager(nsxtConfig *config.Config) (*ConnectorManager, error) {
+	cm := &ConnectorManager{}
+	if nsxtConfig == nil {
+		return cm, nil
```
andrewsykim, line 65: This shouldn't be an error?

andrewsykim: please fix in follow-up PR if this should return error

andrewsykim:
/approve

@nicolehanjing has lgtm

nicolehanjing (approved):
/lgtm
Thanks @dantingl :)

pkg/nsxt/connector_manager.go
dantingl: I thought `nsxtConfig == nil` is a valid case that user does not add nsxt/routes/lb configurations in ConfigMap. Maybe I should check `nsxtConfig` before calling `NewConnectorManager` in cloud.go.

PR #415, opened 2020-12-04
Add in 1.19 YAML into master

dvonthenen:
**What this PR does / why we need it**:
Adds in YAML for 1.19 release into master

**Which issue this PR fixes**:
NA

**Special notes for your reviewer**:
NA

**Release note**:
NA

PR #424, opened 2021-01-13
Bump k8s.io/kubernetes to 1.19.7 and add back InstancesV2

nicolehanjing:
Bump k8s.io/kubernetes to 1.19.7 and add back InstancesV2

**Release note**:

```release-note
Bump k8s.io/kubernetes to 1.19.7 and add back InstancesV2
```


pkg/nsxt/connector_manager.go
```diff
@@ -194,6 +194,9 @@ func (cm *ConnectorManager) GetConnector() client.Connector {
 
 // AddSecretListener adds secret informer add, update, delete callbacks
 func (cm *ConnectorManager) AddSecretListener(secretInformer v1.SecretInformer) error {
+	if cm.config == nil {
```
nicolehanjing, line 197: @andrewsykim received nil pointer errors if the config is not set, thinking we should add this here, or should I open another PR for this fix?

dvonthenen:
When this PR gets merged, please also review the PR for including YAML for 1.19 as well.
https://github.com/kubernetes/cloud-provider-vsphere/pull/415

cluster/images/controller-manager/Dockerfile
```diff
@@ -14,7 +14,7 @@
 ##                               BUILD ARGS                                   ##
 ################################################################################
 # This build arg allows the specification of a custom Golang image.
-ARG GOLANG_IMAGE=golang:1.13.14
+ARG GOLANG_IMAGE=golang:1.15.6
```
andrewsykim, line 17: Is this the GO version used in v1.19.7?

pkg/nsxt/connector_manager.go
```diff
@@ -194,6 +194,9 @@ func (cm *ConnectorManager) GetConnector() client.Connector {
...
+		return errors.New("Config is not available for NSXT connector manager")
```
andrewsykim, line 198: this diff can be removed now

cluster/images/controller-manager/Dockerfile
nicolehanjing: @andrewsykim 
I took a look at https://github.com/kubernetes/kubernetes/blob/v1.19.7/go.mod#L9 and it's using `go 1.15` so I just search the latest Golang image version 1.15.6 and use it here, is that the right way to update it?

andrewsykim: yeah that seems good enough, just want to make sure we are on the same minor at least. Patch version is _probably_ not as important.

andrewsykim (commented):
/approve
/lgtm

PR #425, opened 2021-01-13
Remove NSX-T LB alpha feature gate

andrewsykim:
Removes the NSX-T LB alpha feature gate. This feature has been in use by some folks for a while now and vetted enough to not require an alpha feature gate. Use of the feature still requires opt-in via configuration but the feature gate is probably not warranted. 

Also removes the ability to configure NSX-T routes via env vars. This wasn't supported for NSX-T LBs and given almost all users use configuration files I think it's safe to remove.

This PR also refactors the dual-stack validation logic and adds some tests.

**Release note**:

```
Remove NSX-T LB alpha env var feature gate
```

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud_test.go
+++ b/pkg/cloudprovider/vsphere/cloud_test.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsphere
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/vsphere/cloud_test.go
```diff
@@ -0,0 +1,112 @@
...
+		expectedError error
+	}{
+		{
+			name:        "config dual-stack, gate enabled",
+			gateEnabled: true,
+			cfg: &ccfg.CPIConfig{
+				Config: vcfg.Config{
+					VirtualCenter: map[string]*vcfg.VirtualCenterConfig{
+						"vcenter.local": {
+							IPFamilyPriority: []string{"v4", "v6"},
```
nicolehanjing, line 43: should we consider the testcase with an empty` IPFamilyPriority`?

andrewsykim: added a testcase for the default config

gwang550:
Should we also combine the test in the vsphere_test.go and cloud_test.go, some of the tests are similar, the only difference is how the config is read(INI vs *ccfg.CPIConfig) which I believe is already tested.

andrewsykim:
> Should we also combine the test in the vsphere_test.go and cloud_test.go, some of the tests are similar, the only difference is how the config is read(INI vs *ccfg.CPIConfig) which I believe is already tested.

They test similar code paths but the extra coverage probably doesn't hurt.

PR #426, opened 2021-01-13
Fix nil pointer error if the connector manager config is not set

nicolehanjing:
We would receive nil pointer errors if the connector manager config is not set, this PR fix that issue

**Release note**:

```release-note
NONE
```


pkg/nsxt/connector_manager.go
```diff
@@ -194,6 +194,9 @@ func (cm *ConnectorManager) GetConnector() client.Connector {
 
 // AddSecretListener adds secret informer add, update, delete callbacks
 func (cm *ConnectorManager) AddSecretListener(secretInformer v1.SecretInformer) error {
+	if cm.config == nil {
+		return errors.New("Config is not available for NSXT connector manager")
```
andrewsykim, line 198: error message should typically start with lower case since they can be wrapped

andrewsykim (commented):
minor comment

PR #427, opened 2021-01-14
Add one more validation for NSXT config

dantingl:
NSXT config is required for LB and Route

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/cloud.go
+++ b/pkg/cloudprovider/vsphere/cloud.go
@@ -212,25 +212,34 @@ func (vs *VSphere) HasClusterID() bool {
 func buildVSphereFromConfig(cfg *ccfg.CPIConfig, nsxtcfg *ncfg.Config, lbcfg *lcfg.LBConfig, routecfg *rcfg.Config) (*VSphere, error) {
 	nm := newNodeManager(cfg, nil)
 
-	ncm, err := nsxt.NewConnectorManager(nsxtcfg)
-	if err != nil {
-		return nil, err
-	}
+	var ncm *nsxt.ConnectorManager
+	var lb loadbalancer.LBProvider
+	var routes route.RoutesProvider
+	if nsxtcfg != nil {
+		ncm, err := nsxt.NewConnectorManager(nsxtcfg)
+		if err != nil {
+			return nil, err
+		}
 
-	lb, err := loadbalancer.NewLBProvider(lbcfg, ncm.GetConnector())
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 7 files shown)


dvonthenen:
this looks good to me...

just curious why some many changes to the go.sum. just clean up?

dantingl:
> this looks good to me...
> 
> just curious why some many changes to the go.sum. just clean up?

The file was changed automatically, maybe it's related to I running 'go mod tidy'?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -212,6 +213,10 @@ func (vs *VSphere) HasClusterID() bool {
 func buildVSphereFromConfig(cfg *ccfg.CPIConfig, nsxtcfg *ncfg.Config, lbcfg *lcfg.LBConfig, routecfg *rcfg.Config) (*VSphere, error) {
 	nm := newNodeManager(cfg, nil)
 
+	if nsxtcfg == nil && (lbcfg != nil || routecfg != nil) {
```
andrewsykim, line 216: I think it would be better to indent the whole block here based on nsxtcfg being non-nil.
```
if nsxtcfg != nil {
    	ncm, err := nsxt.NewConnectorManager(nsxtcfg)
	if err != nil {
             ....
	}

        if lbcfg != nil {
              ....
        }

        if routcfg != nil {
               ....
        }
}
```

andrewsykim (commented):
/assign @nicolehanjing

pkg/cloudprovider/vsphere/cloud.go
nicolehanjing: +1

pkg/nsxt/config/config_ini_legacy_test.go
```diff
@@ -23,10 +23,14 @@ import (
 )
 
 func TestINIValidateTokenConfig(t *testing.T) {
-	cfg := &NsxtINI{
+	cfg := &NsxtINI{}
+	err := cfg.validateConfig()
+	assert.EqualError(t, err, "user or vmc access token or client cert file or secret for user must be set")
+
+	cfg = &NsxtINI{
```
nicolehanjing, line 30: I think we can add a testcase where VMCAccessToken is empty but VMCAuthHost is provided

dantingl: OK, then we need add validation for this case as well :)

andrewsykim (commented):
/approve

@nicolehanjing for final lgtm

PR #446, opened 2021-03-12
vSphere Paravirtual - Refactor all the tests to not use ginkgo and use standard unit test instead

nicolehanjing:
- refactor loadbalancer_test.go and vmservice_test.go to not use ginkgo and use standard unit tests + fake client instead
- delete vmservice_suite_test.go since it does not actually test anything
- Follow up: we may don't carry the controller-runtime dependency for long because it will conflict with client-go and other Kubernetes modules.

**Release note**:

```release-note
vSphere Paravirtual - Refactor all the tests to not use ginkgo and use standard unit test instead
```


pkg/cloudprovider/vsphereparavirtual/cloud.go
```diff
@@ -140,7 +140,6 @@ func (cp *VSphereParavirtual) Instances() (cloudprovider.Instances, bool) {
 // InstancesV2 returns an implementation of cloudprovider.InstancesV2.
 //  TODO: implement this for v1.20
 func (cp *VSphereParavirtual) InstancesV2() (cloudprovider.InstancesV2, bool) {
-	klog.Warning("The vSphere cloud provider does not support InstancesV2")
```
nicolehanjing, line 143: actually not so sure if we can remove this - but it's a little noisy every time we debug something

nicolehanjing:
cc @andrewsykim @lubronzhan PTAL :)

pkg/cloudprovider/vsphereparavirtual/cloud.go
andrewsykim: I think it's safe, thanks for doing that

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
+			assert.NoError(t, err)
 
-	Describe("GetLoadBalancer", func() {
-		var (
-			err    error
-			exists bool
-		)
+			_, err = NewLoadBalancer(testClusterNameSpace, cfg, &testOwnerReference)
+			assert.NoError(t, err)
+			assert.Equal(t, testCase.err, err)
```
andrewsykim, line 86: This assert and the one above are checking the same thing I think. Can probably remove one of them.

andrewsykim (commented):
/assign @maplain

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
lubronzhan: The upper one should be removed to be generic

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
-			AfterEach(func() {
-				err := lb.EnsureLoadBalancerDeleted(context.Background(), testClustername, testK8sService)
-				Expect(err).ShouldNot(HaveOccurred())
-			})
-			It("should not return error", func() {
-				Expect(exists).Should(Equal(true))
-				Expect(err).ShouldNot(HaveOccurred())
-			})
-		})
+func TestGetLoadBalancer_VMServiceNotFound(t *testing.T) {
```
lubronzhan, line 94: These several `TestGetLoadBalancer` could be refactored to single func with list of testcases

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
+		name      string
+		expectErr bool
+	}{
+		{
+			name:      "when VMService update failed",
+			expectErr: true,
+		},
+		{
+			name:      "when VMService is updated",
+			expectErr: true,
```
lubronzhan, line 154: The original testcase doesn't expect error here.

Maybe add another variable called updateFunc, one return error one doesn't

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
+	lb, fcw := newTestLoadBalancer()
+	testK8sService := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testK8sServiceName,
+			Namespace: testK8sServiceNameSpace,
+		},
+	}
+	// Ensure that the client Create call will fail. The preceding Get call should return NotFound
+	fcw.CreateFunc = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+		return fmt.Errorf("failed to create VirtualMachineService")
```
lubronzhan, line 258: nit: could be replaced with `ErrCreateVMService`

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
+	assert.NoError(t, ensureErr)
 
-		JustBeforeEach(func() {
-			status, ensureErr = lb.EnsureLoadBalancer(context.Background(), testClustername, testK8sService, []*v1.Node{})
-		})
+	err := lb.EnsureLoadBalancerDeleted(context.Background(), testClustername, testK8sService)
+	assert.NoError(t, err)
+}
+
+func TestEnsureLoadBalancer_VMServiceCreatedIPNotFound(t *testing.T) {
```
lubronzhan, line 231: nit: These two ensure could be refactored to use testcases slice with variable called `createFunc`

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
```diff
@@ -44,385 +40,310 @@ import (
...
-					return fmt.Errorf("an error occurred while deleting load balancer")
-				}
-				err := lb.EnsureLoadBalancerDeleted(context.Background(), "test", testK8sService)
-				Expect(err).To(MatchError("an error occurred while deleting load balancer"))
-			})
-		})
-	})
-})
+	// should ignore not found error
+	fcw.DeleteFunc = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
```
lubronzhan, line 336: Nit: would be best if could be several test cases with different DeletFunc

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice_test.go
```diff
@@ -41,669 +40,595 @@ import (
...
+				},
+			},
+		},
+	}
+	vmServiceObj, err := vms.Create(context.Background(), k8sService, testClustername)
+	assert.Equal(t, vmServiceObj, (*vmopv1alpha1.VirtualMachineService)(nil))
+	assert.Error(t, err)
+}
+
+func TestCreateVMService_CreationErr(t *testing.T) {
```
lubronzhan, line 213: Duplicate vmService?

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice_test.go
```diff
@@ -41,669 +40,595 @@ import (
...
+
+	_, ok = vmServiceObj.Annotations[AnnotationServiceHealthCheckNodePortKey]
+	assert.NotEqual(t, ok, true)
+
+	testK8sService.Spec.ExternalTrafficPolicy = ""
+	err = vms.Delete(context.Background(), testK8sService, testClustername)
+	assert.NoError(t, err)
+}
+
+func TestVMService_NotExist(t *testing.T) {
```
lubronzhan, line 350: Maybe still add `CreateOrUpdate` to show it's testing this func

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice_test.go
```diff
@@ -41,669 +40,595 @@ import (
...
+	// Redefine Get in the client to return an error
+	fcw.GetFunc = func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+		return fmt.Errorf("failed to get VirtualMachineService")
+	}
+	_, err := vms.CreateOrUpdate(context.Background(), testK8sService, testClustername)
+	assert.Error(t, err)
+}
+
+func TestVMService_CreationFails(t *testing.T) {
+	testK8sService, vms, fcw := initTest()
```
lubronzhan, line 380: I think the original test cases is not well defined, this should be split into two cases

lubronzhan:
For clean up func that used to be inside `afterEach`, I think it's better to put inside `defer` like 
```
defer cleanup()

func cleanup(
_:= vms.Delete()
)
```
Since we can't guarantee the cleanup will be called if test cases fail in the middle
Example: https://github.com/kubernetes/kubernetes/blob/master/test/integration/service/loadbalancer_test.go#L50

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
nicolehanjing: good catch, updated!

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
nicolehanjing: actually - the test flow is different
the first one only calls GetLB() but the second one also calls EnsureLB and EnsureLBDeleted functions
we can combine them but it may not be readable I think 🤔

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
nicolehanjing: oops I should set expectErr to false, will update!

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
nicolehanjing: got it, yeah I agree

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice_test.go
nicolehanjing: can you explain a little more?
should we remove 
```
// Redefine Get in the client to return an error
	fcw.GetFunc = func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
		return apierrors.NewNotFound(v1alpha1.Resource("virtualmachineservice"), testClustername)
	}
```?

lubronzhan: I think it's two cases  that one failed at Get, one failed at Create, you can create two cases testing this func

pkg/cloudprovider/vsphereparavirtual/loadbalancer_test.go
lubronzhan: ok hm. Then let's keep it

lubronzhan (approved):
LGTM, just that refactor of splitting testcases

nicolehanjing:
@maplain Can you take a look when you get a chance? Thanks! :)

PR #453, opened 2021-03-24
Refactor main.go based on sample main from k8s go module version v0.20.6

nicolehanjing:
Previously we refactor `main.go` to bypass an upstream bug, now the bug has been fixed and **cloud provider repo** provides a new [sample main](https://github.com/kubernetes/cloud-provider/blob/master/sample/basic_main.go).
This refactoring of cloud provider vsphere`main.go` is based on the latest sample main.
This PR also updates k8s go module version to v0.20.6 in order to fetch the latest code of k8s repos.

**Release note**:

```release-note
Refactor main.go based on sample main from k8s go module version v0.20.6
```


cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -165,21 +133,12 @@ func main() {
...
 	// utilflag.InitFlags() (by removing its pflag.Parse() call). For now, we have to set the
 	// normalize func and add the go flag set by hand.
-	pflag.CommandLine.SetNormalizeFunc(flag.WordSepNormalizeFunc)
+	pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	// utilflag.InitFlags()
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
-	// Set cloud-provider flag to vsphere
```
nicolehanjing, line 174: looks like that was just leftover code
I tried changing the flag value to "vsphere" or "vsphere-paravirtual", there is no difference in the log/behavior and didn't see any error log during testing, maybe we could remove this?

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -210,148 +169,24 @@ func main() {
 	}
 }
 
-// initFunc is used to launch a particular controller.  It may run additional "should I activate checks".
```
nicolehanjing, line 213: this deleted code was originally copied from the cloud provider repo to bypass upstream bug, as the upstream fix merged, we could directly using it from the cloud provider repo

go.mod
```diff
@@ -19,22 +19,22 @@ require (
...
-	k8s.io/cloud-provider v0.20.2
-	k8s.io/component-base v0.20.2
-	k8s.io/klog/v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0
+	honnef.co/go/tools v0.1.3 // indirect
+	k8s.io/api v0.21.0-beta.1
+	k8s.io/apimachinery v0.21.0-beta.1
+	k8s.io/apiserver v0.21.0-beta.1
+	k8s.io/client-go v0.21.0-beta.1
+	k8s.io/cloud-provider v0.21.0-beta.1
```
andrewsykim, line 34: I don't think we can bump this into v0.21 until we released v1.20 version of cloud-provider-vsphere.

So instead we should ask Cici to backport the fix into v1.20 and use the next patch release of 1.20 with this fix.

nicolehanjing: got it, I think there is already a PR out for the cherry-pick, once it's merged, I will update my PR

cmd/vsphere-cloud-controller-manager/main.go
andrewsykim: Makes sense to me

andrewsykim: I think this is to cover the case when `--cloud-provider` is not set, and have it then default to `vsphere`. Have you tried that?

cmd/vsphere-cloud-controller-manager/main.go
andrewsykim: sounds good, thanks for the context!

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -210,148 +169,24 @@ func main() {
...
-	}
-
-	// If CIDRs should be allocated for pods and set on the CloudProvider, then start the route controller
-	routes, ok := cloud.Routes()
-	if !ok {
-		klog.Warning("configure-cloud-routes is set, but cloud provider does not support routes. Will not configure cloud provider routes.")
-		return nil, false, nil
-	}
+func cloudInitializer(config *appconfig.CompletedConfig, cloudProvider string) cloudprovider.Interface {
+	cloudConfig := config.ComponentConfig.KubeCloudShared.CloudProvider
```
andrewsykim, line 173: Should this just be:
```
cloudConfig := config.ComponentConfig.KubeCloudShared.CloudProvider.CloudConfigFile?
```

andrewsykim (commented):
minor comments and questions, lgtm otherwise

andrewsykim:
Left a comment in an old thread in case it gets lost https://github.com/kubernetes/cloud-provider-vsphere/pull/453#discussion_r600858049

cmd/vsphere-cloud-controller-manager/main.go
nicolehanjing: I write it in the next line:
`cloud, err := cloudprovider.InitCloudProvider(cloudProvider, cloudConfig.CloudConfigFile)`
basically they are equal

andrewsykim (commented):
/approve
/lgtm

Feel free to cancel hold once this question is addressed https://github.com/kubernetes/cloud-provider-vsphere/pull/453#discussion_r626006399

cmd/vsphere-cloud-controller-manager/main.go
nicolehanjing: ah you are right, I added it back
if --cloud-provider is not set, we will get errors

nicolehanjing:
@andrewsykim for #453 (comment), I added back the code block
you are right, it handles the case when cloud-provider flag is missing
can I get another LGTM label? cuz after pushing the previous label was removed

PR #466, opened 2021-05-25
Add IPPool CRD API and client interfaces

dantingl:
IPPool CRD is used to allocate a list of subnets for K8s cluster IPAM. Every subnet will be added in K8s Node podCIDR.
Cloud provider sets desired subnet list in the CR spec, another CR consumer will allocate subnet on NSX-T side and set realized subnet result in the status. The status would be 'Ready=False' if any of subnets realization fails.
For K8s Node scale up and down, cloud provider needs to update the desired subnet list in the CR spec, new subnet could be allocated and existing one could be deleted on NSX-T side.


pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+
+// IPPoolConditionType describes the IPPool condition type
+type IPPoolConditionType string
+
+const (
+	// IPPoolConditionTypeReady means IPPool is healthy
+	IPPoolConditionTypeReady IPPoolConditionType = "ready"
+	// IPPoolConditionTypeFailure means IPPool is not healthy
+	IPPoolConditionTypeFailure IPPoolConditionType = "failure"
+)
```
lubronzhan, line 65: Could you add these default value here as well?
```
// PrefixLengthDefault defines the default value for prefixLength
	PrefixLengthDefault int = 24
	// IPFamilyDefault defines the default type for IPfamily
	IPFamilyDefault string = "ipv4"

```

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/doc.go
```diff
@@ -0,0 +1,4 @@
+// +k8s:deepcopy-gen=package,register
```
nicolehanjing, line 1: pls add k8s license header

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+// IPPoolStatus defines the observed state of IPPool.
+type IPPoolStatus struct {
+	// Subnets defines subnets allocation result.
+	Subnets []SubnetRes `json:"subnets"`
+	// Conditions defines current state of the IPPool.
+	Conditions []IPPoolCondition `json:"conditions"`
+}
+
+// SubnetReq defines the subnet allocation request.
+type SubnetReq struct {
```
jianjuns, line 35: I assume later we could add a CIDR fields for requesting a specific CIDR?

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// IPPoolConditionType describes the IPPool condition type
+type IPPoolConditionType string
+
+const (
+	// IPPoolConditionTypeReady means IPPool is healthy
+	IPPoolConditionTypeReady IPPoolConditionType = "ready"
+	// IPPoolConditionTypeFailure means IPPool is not healthy
+	IPPoolConditionTypeFailure IPPoolConditionType = "failure"
```
jianjuns, line 64: But does it overlap with IPPoolConditionTypeReady?

jianjuns (approved):

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// IPFamily defines the ip family type for this subnet, could be ipv4 or ipv6.
+	// +optional
+	IPFamily string `json:"ipFamily,omitempty"`
+
+	// Name defines the name of this subnet.
+	Name string `json:"name"`
+}
+
+// SubnetRes defines the subnet allocation result.
+type SubnetRes struct {
```
andrewsykim, line 49: The `Res` short-form is probably more confusing than helpful I think, can this be named just `SubnetResult`?

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// Name defines the name of this subnet.
+	Name string `json:"name"`
+}
+
+// IPPoolConditionType describes the IPPool condition type
+type IPPoolConditionType string
+
+const (
+	// IPPoolConditionTypeReady means IPPool is healthy
+	IPPoolConditionTypeReady IPPoolConditionType = "ready"
```
andrewsykim, line 62: Condition values should start with capital: `Ready`

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
andrewsykim, line 64: Same here, should be `Failure`

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// IPPoolConditionTypeFailure means IPPool is not healthy
+	IPPoolConditionTypeFailure IPPoolConditionType = "failure"
+)
+
+// IPPoolCondition defines the condition for the IPPool
+type IPPoolCondition struct {
+	Type    IPPoolConditionType    `json:"type"`
```
andrewsykim, line 69: Add comments please

andrewsykim:
/approve

@nicolehanjing has lgtm

pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/ippool_types.go
jianjuns: But what is the convention here? If we remove "optional", then it will be incompatible change later when we make it optional.

dantingl: Good point, let me add it back, but I don't think we should use 24 as default value here.

PR #467, opened 2021-05-29
Update RouteSet CRD

dantingl:
1. Add realized routes in RouteSetStatus
2. Update comments for target and destination


pkg/cloudprovider/vsphereparavirtual/apis/nsxnetworking/v1alpha1/routeset_types.go
```diff
@@ -44,8 +44,10 @@ type Route struct {
 	// Name is the name of this route entry.
 	Name string `json:"name"`
 	// Destination is the CIDR block used for the destination match.
+	// It is network address in CIDR format, support both IPv4 and IPv6.
 	Destination string `json:"destination"`
 	// Target is the IP address used to determine where traffic goes to.
+	// Support both IPv4 and IPv6, and it needs to be consistent with Destination.
 	Target string `json:"target"`
 }
```
jianjuns, line 53: Do you think we need to track which routes are realized in RouteSetStatus? I assume it is possible some routes are accepted, while some others are not. Or maybe we should just introduce a RoutingTable resource to track that?

dantingl: Let me add realized routes in `RouteSetStatus`. Previously I did not add that because right now from cloud provider side it creates `RouteSet` per Node, so the `Ready=True` in `status.conditions` is good enough for them to check realization state.

nicolehanjing: I see, for now we keep using `Conditions` field to determine realized state but in the future, we would edit the corresponding logic to use `Routes` field instead

PR #468, opened 2021-05-31
Implement Route interface and add the script to generate client code

nicolehanjing:
This PR does the following items:
1. add the script / Makefile used to generate the generated client code for nsx apis
2. implement routes interface
3. add routes unit tests
4. move some common utils into centralized folder 'util'

**TODO:**
- use the dynamic client-go client and remove all generated client code. Only need the Go type for RouteSet (blocked by unit testing)
- when backend finishes, check realized state of static routes and do manual testing again

**Release note**:

```release-note
Implement Route interface and add the script to generate client code
Update k8s components version to v0.21.0
```


pkg/cloudprovider/vsphere/route/routeprovider.go
```diff
@@ -115,7 +116,7 @@ func (p *routeProvider) CreateRoute(ctx context.Context, clusterName string, nam
 	nodeName := string(route.TargetNode)
 	klog.V(6).Infof("Creating static route for node %s", nodeName)
 
-	nodeIP, err := p.getNodeIPAddress(nodeName, IsIPv4(route.DestinationCIDR))
+	nodeIP, err := p.getNodeIPAddress(nodeName, util.IsIPv4(route.DestinationCIDR))
```
nicolehanjing, line 119: move `IsIPv4` into a common folder 'util' (unit tests added) so vsphere and vsphere-paravirtual can both use

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+	return r.createCPRoutes(routeSets, clusterName), nil
+}
+
+// createCPRoutes creates cloudprovider Routes based on RouteSet CR
+func (r *routesProvider) createCPRoutes(routeSets *v1alpha1.RouteSetList, clusterName string) []*cloudprovider.Route {
+	var routes []*cloudprovider.Route
+	for _, routeSet := range routeSets.Items {
+		routeSetClusterName := routeSet.Annotations[AnnotationClusterName]
+		if routeSetClusterName == clusterName {
+			// only return cloudprovider.Route if RouteSet CR status 'Ready' is true
```
nicolehanjing, line 150: I commented the check for the realized state of routes for now, we could either 1. uncomment & merge them now, and fix/test later once backend finishes, or 2. uncomment only after we test the logic after backend finishes

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+	kubeClient  clientset.Interface
+	nodeLister  listerv1.NodeLister
+}
+
+var _ RoutesProvider = &routesProvider{}
+
+const (
+	// AnnotationClusterName label is used to annotate GC name for Routes CR
+	AnnotationClusterName = v1alpha1.GroupName + "/clusterName"
+	// RealizedStateTimeout is the timeout duration for realized state check
```
lubronzhan, line 72: Remove these if not used

nicolehanjing: https://github.com/kubernetes/cloud-provider-vsphere/pull/468/files#r643316448 I would need to uncomment those fields if we plan to merge the check for realized state with this PR

pkg/cloudprovider/vsphereparavirtual/cloud.go
```diff
@@ -111,6 +111,16 @@ func (cp *VSphereParavirtual) Initialize(clientBuilder cloudprovider.ControllerC
 		klog.Fatalf("Failed to get cluster namespace: %v", err)
 	}
 
+	nodeLister := cp.informMgr.GetNodeLister()
+	routes, err := NewRoutes(clusterNS, kcfg, *cp.ownerReference, cp.client, nodeLister)
```
dantingl, line 115: Do we have a flag in cloud provider to enable/disable this feature?

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+	}, nil
+}
+
+// ListRoutes implements Routes.ListRoutes
+// Get RouteSet CR from SC namespace and then filters routes that belong to the specified clusterName
+// Only return cloudprovider.Route if RouteSet CR status 'Ready' is true
+func (r *routesProvider) ListRoutes(ctx context.Context, clusterName string) ([]*cloudprovider.Route, error) {
+	klog.V(6).Infof("Attempting to list Routes for cluster %s", clusterName)
+
+	routeSets, err := r.routeClient.NsxV1alpha1().RouteSets(r.namespace).List(ctx, metav1.ListOptions{})
```
dantingl, line 130: If adding clusterName in label, then you can add a labelSelector in `ListOptions`, then no need to filter RouteSet CRs which are belong to this cluster.

pkg/cloudprovider/vsphereparavirtual/route.go
dantingl: I don't think you need to wait for backend, you can always check the CR status.
Or do you mean after we add realized routes in CR status, you'll switch to get realized routes there?

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	_, err := r.routeClient.NsxV1alpha1().RouteSets(r.namespace).Create(ctx, routeSet, metav1.CreateOptions{})
+	if err != nil {
+		klog.ErrorS(ErrCreateRouteSet, fmt.Sprintf("%v", err))
+		return nil, err
+	}
+
+	klog.V(6).Infof("Successfully created RouteSet CR for node %s", nodeName)
+	err = r.updateNetworkingCondition(nodeName)
```
dantingl, line 230: Is it necessary to update `NodeNetworkUnavailable` here?
If I remember it correctly, route controller will update it after it gets `ListRoutes` back?
If route controller does not do that, I think it's better to update `NodeNetworkUnavailable` only for realized routes.

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+		klog.V(6).Infof("Error updating NetworkingCondition for node %s, error: %v", types.NodeName(nodeName), err)
+	}
+	return routeSet, nil
+}
+
+// updateNetworkingCondition removes "NodeNetworkUnavailable" condition on the Node once the Route CR is successfully created
+func (r *routesProvider) updateNetworkingCondition(nodeName string) error {
+	klog.V(6).Infof("updating NetworkingCondition for node %s", nodeName)
+
+	nodes, err := r.nodeLister.List(labels.Everything())
```
dantingl, line 241: How about using `r.nodeMap` as you have the list already, I'm not exactly sure whether `r.nodeLister.List` will have a extra call to apiserver?

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+func (r *routesProvider) DeleteRoute(ctx context.Context, clusterName string, route *cloudprovider.Route) error {
+	routeSetName := string(route.TargetNode)
+	klog.V(6).Infof("Deleting RouteSet CR %s in cluster %s", routeSetName, clusterName)
+	return r.DeleteRouteSetCR(routeSetName)
+}
+
+// GetRouteName returns Route name as <nodeName>-<cidr>-<clusterName>
+// e.g. nodeName-100.96.0.0-24-clusterName
+func (r *routesProvider) GetRouteName(nodeName string, cidr string, clusterName string) string {
+	return strings.Replace(nodeName+"-"+cidr+"-"+clusterName, "/", "-", -1)
```
dantingl, line 318: Is it better to put `clusterName` in the front?

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+	annotations := map[string]string{
+		AnnotationClusterName: clusterName,
+	}
+	nodeRef := metav1.OwnerReference{
+		APIVersion: "v1",
+		Kind:       "Node",
+		Name:       nodeName,
+		UID:        types.UID(nameHint),
+	}
+	r.ownerRefs = append(r.ownerRefs, nodeRef)
```
lubronzhan, line 202: Should it be a tmp ownerRef other than directly assigning to this slice used for all nodes?

pkg/cloudprovider/vsphereparavirtual/route.go
lubronzhan: I think lister could do something like this
```
ippool, err := c.ippoolLister.IPPools(namespace).Get(name)
```

pkg/cloudprovider/vsphereparavirtual/route.go
lubronzhan: Yeah the cloud-provider does add this condition

https://github.com/kubernetes/cloud-provider/blob/master/controllers/route/route_controller.go#L270

pkg/cloudprovider/vsphereparavirtual/route.go
lubronzhan: Because can't end the string with `.` or number, that's why the cluster name is put at the end

pkg/cloudprovider/vsphereparavirtual/route.go
lubronzhan: https://github.com/kubernetes/cloud-provider/blob/master/controllers/route/route_controller.go#L111-L115
The upstream is doing the resync every 5 minutes. maybe that's too long, but still it should be fixed upstream

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,415 @@
...
+				}
+			}
+		}
+	}
+
+	klog.V(6).Infof("Successfully updated NetworkingCondition for node %s", nodeName)
+	return err
+}
+
+// checkStaticRouteRealizedState checks static route realized state
```
lubronzhan, line 283: It's a good to have. Since we can't reflect the status inside TKC

pkg/cloudprovider/vsphereparavirtual/cloud.go
nicolehanjing: yes, if `controllers=route` is set, then we would enable route controller and its corresponding logic

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: ah I mean after backend finishes, I could test the logic and verify if it's working correctly, but I could also merge this logic before testing it with backend, if there are any bugs I could fix later.

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: ah good catch, let me fix it

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: nvm, I uncomment the code, we could merge them together

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: yeah I agree, this is something we need to fix in the upstream - 5 mins is too long
sometimes I see cloud provider Pod pending because of this, that's why I added the update here
but I agree that it might not be good to add `updateNetworkingCondition` here, let me remove this for now and we can fix it in the upstream later

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: yeah, we can only change it to something like `<clusterName>-<cidr>-<nodeName>`, but I think clusterName is only added for passing format validation, and nodeName should be put first as it maps to the corresponding route

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: yeah that's another option, I think either way works
if we use the labelSelector, is that okay for us to remove `clusterName` from annotations, does backend need this for other uses?

pkg/cloudprovider/vsphereparavirtual/route.go
dantingl: https://github.com/kubernetes/cloud-provider/blob/5a236942a3ca6f0454bfcb979cc31207af4894b2/config/v1alpha1/defaults.go#L66
The comment about 5 minutes may be wrong, it should be 10 seconds. So I think removing updateNetworkingCondition here should be fine.

pkg/cloudprovider/vsphereparavirtual/route.go
dantingl: It's fine, right now backend does not read RouteSet annotations.

pkg/cloudprovider/vsphereparavirtual/route.go
dantingl: Got it, thanks

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,353 @@
...
+		return err
+	}
+
+	routeSet, err := r.createRouteSetCR(ctx, clusterName, nameHint, nodeName, route.DestinationCIDR, nodeIP)
+	if err != nil {
+		klog.Errorf("creating RouteSet CR for node %s failed: %s", nodeName, err)
+		return err
+	}
+	// check realized state of static routes
+	return r.checkStaticRouteRealizedState(routeSet.Name)
```
dantingl, line 174: Do we need to check realized state after creating RouteSet CR here? I mean cloud provider returns realized routes in `ListRoutes` interface

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -0,0 +1,353 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            nodeName,
+			OwnerReferences: owners,
+			Namespace:       r.namespace,
+			Labels:          labels,
+		},
+		Spec: routeSetSpec,
+	}
+
+	_, err := r.routeClient.NsxV1alpha1().RouteSets(r.namespace).Create(ctx, routeSet, metav1.CreateOptions{})
```
dantingl, line 211: If the RouteSet CR is already created, will there be any error here? Or it just apply to the existing CR

nicolehanjing: yes, the error will be caught and `ErrCreateRouteSet` will be returned, in that case, we won't create a duplicate CR

dantingl: Do you think we should not throw `ErrCreateRouteSet` if the CR is already created?  especially in cloud provider restart case, maybe it is a little confusing to report these errors.

nicolehanjing: but I think if the routeSet CR already exists, maybe we should not regard it as an error, we just do nothing afterwards

dantingl: I agree, it should not be an error.

dantingl (approved):

pkg/cloudprovider/vsphereparavirtual/route.go
nicolehanjing: if we don't check realized state here, it's possible that we've created a routeSet CR, but the static route is not actually being created
in that case, next time when we do `ListRoutes`, one route is missing for a node, then we will need to try `CreateRoute` one more time: https://github.com/kubernetes/cloud-provider/blob/master/controllers/route/route_controller.go#L194
but if we check the realized state at this point, we could retry as well: https://github.com/kubernetes/cloud-provider/blob/master/controllers/route/route_controller.go#L188
But I agree that either way works, cuz the route controller does retry within a few seconds

lubronzhan (approved):

PR #477, opened 2021-06-10
Use Ippool's status as source of truth. Regardless cidr value, assign…

lubronzhan:
… to node anyway

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphereparavirtual/cloud.go
+++ b/pkg/cloudprovider/vsphereparavirtual/cloud.go
@@ -125,7 +125,6 @@ func (cp *VSphereParavirtual) Initialize(clientBuilder cloudprovider.ControllerC
 	cp.routes = routes
 
 	cp.informMgr.AddNodeListener(cp.nodeAdded, cp.nodeDeleted, nil)
-	cp.informMgr.Listen()
 
 	lb, err := NewLoadBalancer(clusterNS, kcfg, cp.ownerReference)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ippool/ippool_controller.go
```diff
@@ -215,10 +215,7 @@ func (c *Controller) processIPPoolCreateOrUpdate(ippool *ippoolv1alpha1.IPPool)
 	// make map of allocated subnets
 	subs := make(map[string]string)
 	for _, sub := range ippool.Status.Subnets {
-		// if ippool is exhausted, the cidr value could be empty
-		if sub.CIDR != "" {
-			subs[sub.Name] = sub.CIDR
-		}
+		subs[sub.Name] = sub.CIDR
```
nicolehanjing, line 218: does this aim to avoid 'null'?
and we just display empty values in status (if CIDR is empty)?

pkg/cloudprovider/vsphereparavirtual/cloud.go
```diff
@@ -125,7 +125,6 @@ func (cp *VSphereParavirtual) Initialize(clientBuilder cloudprovider.ControllerC
 	cp.routes = routes
 
 	cp.informMgr.AddNodeListener(cp.nodeAdded, cp.nodeDeleted, nil)
-	cp.informMgr.Listen()
```
nicolehanjing, line 128: why should we remove this?
if we intend to remove this, should we not call `cp.informMgr.AddNodeListener`?

lubronzhan: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphereparavirtual/cloud.go#L150
Because there is another one here, I intended to call it after the handler registration inside `StartControllers`

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ippool/ippool_controller.go
lubronzhan: There are cases like cidr exhaustion, the value of CIDR in status will be empty string. I set this to make sure that status.cidr will be source of truth, regardless it's value.
If NCP fail to allocate cidr, then the podCIDR should have the same value.

PR #481, opened 2021-07-01
Fix link in in_tree_vs_out_of_tree.md

bostrt:

Fixes a bad link in doc


nicolehanjing (approved):
Thanks for the fix!

Issue #521, opened 2021-11-03
CPI is not adding proper 'ProviderID:' on the node with UUID


nicolehanjing:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
We met an issue that when some nodes were deleted by the vSphere cloud provider(CPI), and later they got added back on the k8s cluster, CPI cannot set the proper providerID when the de-registered node was added back on the k8s cluster.

We've seen this case recently when the VM is undergoing failover, then when the VM is back, the UUID is empty. The UUID is fetched from vSphere API

The context is that CPI generates providerID by combining "provider_name + system UUID" e.g. "vsphere://UUID", but the system UUID we fetched from vCenter (via govmomi) is empty. So the final providerID we get is "vsphere://", which is invalid. 

**What you expected to happen**:
When registering the node, UUID should not be empty and a proper ProviderID should be assigned.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- vsphere-cloud-controller-manager version: 1.21.1
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

nicolehanjing:
after checking with vSphere team, there is a short period of time while a VM is being created that `config.uuid` will be empty. 
To fix this, we'd like to defer node registration until the UUID is set.

PR #522, opened 2021-11-04
disable node registration when UUID is empty

nicolehanjing:
CPI generates providerID by combining "provider_name + system UUID" e.g. "vsphere://UUID", but the system UUID we fetched from vCenter (via govmomi) is empty. So the final providerID we get is "vsphere://", which is invalid.
In such case, we should disable node registration when UUID is empty

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/521

**Release note**:

```release-note
disable node registration when UUID is empty
```

### DIFF
--- a/pkg/cloudprovider/vsphere/instances.go
+++ b/pkg/cloudprovider/vsphere/instances.go
@@ -36,9 +36,6 @@ var (
 	// ErrNotFound is returned by NodeAddresses, NodeAddressesByProviderID,
 	// and InstanceID when a node cannot be found.
 	ErrNodeNotFound = errors.New("Node not found")
-
-	// ErrNodeUUIDEmpty is returned by InstanceID when UUID of a node is empty.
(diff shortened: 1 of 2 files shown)


nicolehanjing:
cc @dougm @lubronzhan

lubronzhan:
You can just error out here if vmDI has empty UUID
https://github.com/kubernetes/cloud-provider-vsphere/blob/6ba6d0e10e8aaa30ce5520944faf218c27ddcf14/pkg/cloudprovider/vsphere/nodemanager.go#L190

Also could you update the log here https://github.com/kubernetes/cloud-provider-vsphere/blob/6ba6d0e10e8aaa30ce5520944faf218c27ddcf14/pkg/cloudprovider/vsphere/nodemanager.go#L315

Issue #302, opened 2020-03-11
Does cloud-provider-vsphere support dual stack?


deshui123:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> /kind feature

**What happened**:

At present, kubernetes 1.17 and cloud-provider-openstack can support dual-stack.
When tested cloud-provider-vsphere with dual-stack and version 1.17, I got some errors and   the taint **node.cloudprovider.kubernetes.io/uninitialized** still exists

**What you expected to happen**:

cloud-provider-vsphere with dual-stack can work well, can get node's IPv4/IPv6 interfaces.
```
# kubectl get nodes henry-dual-we-01 -o go-template --template='{{range .status.addresses}}{{printf "%s: %s \n" .type .address}}{{end}}'  
InternalIP: 10.75.72.170
InternalIP: 192.16.1.15
InternalIP: 2001:282:4000:2000::9
Hostname: henry-dual-we-01
```
**How to reproduce it (as minimally and precisely as possible)**:

As I know, cloud controller manager for openstack has already support dual-stack.
I used image **docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest** and it work well.

Now I tested CPI for vsphere using kubeadm and am according to the guide https://raw.githubusercontent.com/kubernetes/cloud-provider-vsphere/master/manifests/controller-manager/vsphere-cloud-controller-manager-ds.yaml

I added **IPv4 and IPv6 CIDRs** for podsubnet, but get some errors
```
networking:
  serviceSubnet: 10.254.0.0/16,fd01:abcd::/112
  podSubnet: 192.168.1.0/22,2001:283:4000:2001::/62
```
```
[root@henry-master exec]# kubectl logs -f -n kube-system   vsphere-cloud-controller-manager-6kc69
Flag --address has been deprecated, see --bind-address instead.
I0310 06:24:02.777223       1 flags.go:33] FLAG: --address="127.0.0.1"
I0310 06:24:02.777856       1 flags.go:33] FLAG: --allocate-node-cidrs="false"
I0310 06:24:02.777870       1 flags.go:33] FLAG: --allow-untagged-cloud="false"
I0310 06:24:02.777879       1 flags.go:33] FLAG: --alsologtostderr="false"
I0310 06:24:02.777937       1 flags.go:33] FLAG: --authentication-kubeconfig=""
I0310 06:24:02.777950       1 flags.go:33] FLAG: --authentication-skip-lookup="false"
[...]

PR #524, opened 2021-11-04
Internal & External subnets accepts multiple CIDRs

tylerschultz:
internal-network-subnet-cidr and external-network-subnet-cidr config
variables accept a list of comma separated CIDRs. Encountered IPs are
matched against each list of CIDRs.

**Which issue this PR fixes**

This is related to issue #302

Verification steps:
 - added unit tests for new functionality
 - built and pushed a ccm image to a dualstack env.
 - updated the internal-network-subnet-cidr and external-network-subnet-cidr to contain comma separated lists of CIDRs
 - verified the Node objects spec settings contained the desired IPs
 - verified the logs showed the IPs were assigned via subnet matching. 

**Release note**:

```release-note
internal-network-subnet-cidr and external-network-subnet-cidr config
variables accept a list of comma separated CIDRs
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -222,23 +222,19 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		klog.Warningf("Unable to find vcInstance for %s. Defaulting to ipv4.", tenantRef)
 	}
 
-	var internalNetworkSubnet *net.IPNet
-	var externalNetworkSubnet *net.IPNet
+	var internalNetworkSubnets []*net.IPNet
+	var externalNetworkSubnets []*net.IPNet
 	var internalVMNetworkName string
 	var externalVMNetworkName string
 
 	if nm.cfg != nil {
-		if nm.cfg.Nodes.InternalNetworkSubnetCIDR != "" {
-			_, internalNetworkSubnet, err = net.ParseCIDR(nm.cfg.Nodes.InternalNetworkSubnetCIDR)
-			if err != nil {
-				return err
-			}
+		internalNetworkSubnets, err = parseCIDRs(nm.cfg.Nodes.InternalNetworkSubnetCIDR)
+		if err != nil {
+			return err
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -494,14 +507,19 @@ func findSubnetMatch(ipAddrNetworkNames []*ipAddrNetworkName, networkSubnet *net
...
+		return findFirst(ipAddrNetworkNames, func(candidate *ipAddrNetworkName) bool {
 			return strings.EqualFold(networkName, candidate.networkName)
 		})
+	}
+	return nil
+}
 
-		if len(networkNameMatches) > 0 {
-			return networkNameMatches[0]
+// findFrist returns the first occurance that matches the given predicate
```
christianang, line 517: typo: First

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -494,14 +507,19 @@ func findSubnetMatch(ipAddrNetworkNames []*ipAddrNetworkName, networkSubnet *net
...
+func findFirst(ipAddrNetworkNames []*ipAddrNetworkName, predicate func(*ipAddrNetworkName) bool) *ipAddrNetworkName {
```
christianang, line 518: I see this function replaces the `filter` function, but the filter function still exists. Is there somewhere it is still being used or can it be deleted?

pkg/cloudprovider/vsphere/nodemanager.go
tylerschultz: doh! good catch

pkg/cloudprovider/vsphere/nodemanager.go
tylerschultz: filter is still used in 2 places

MaxRink:
@mcwumbly This seems to adress my issue as a whole, given that i can hereby just specify small CIDRs in a list. 
Also DualStack :+1: :partying_face:

nicolehanjing (approved):

PR #532, opened 2021-11-25
Add notifier for config map updates.

### LINKED ISSUE #531: Reload pod if cloud secret is updated

Currently, CPI reads data from a configMap and that configMap also refers to a secret:
For example:
https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.22/vsphere-cloud-controller-manager.yaml#L11-L65

We should have a controller that watches these resource. Also the client shouldn't cache the CM and Secret since there could be tons of them.

### Why is this needed?

When the data in these config is updated, the process won't be triggered to restarted to read the new data.

HanFa:

Notify the app to stop and reload pod if there's an update to `vsphere-cloud-config` config map.

**Which issue this PR fixes**: fixes #531

This feature has been tested manually.

**Release note**:

```release-note
Cloud controller manager pods will restart whenever there is update to the vsphere-cloud-config configmap.
```

### DIFF
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -27,7 +27,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere"
 	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/loadbalancer"
@@ -44,6 +43,7 @@ import (
 	"k8s.io/component-base/version/verflag"
 	klog "k8s.io/klog/v2"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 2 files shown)


nicolehanjing:
@HanFa can you add the release notes in the PR description and also share testing steps?

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -74,7 +74,7 @@ func main() {
 
 			c, err := ccmOptions.Config(app.ControllerNames(app.DefaultInitFuncConstructors), app.ControllersDisabledByDefault.List())
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "%v\n", err)
+				_, _ = fmt.Fprintf(os.Stderr, "%v\n", err)
```
nicolehanjing, line 77: wondering why we are making this change 🤔

HanFa: we are ignoring any error thrown by `Fprintf`, adding _ will make it more explicit

go.mod
```diff
@@ -3,6 +3,7 @@ module k8s.io/cloud-provider-vsphere
 go 1.16
 
 require (
+	github.com/fsnotify/fsnotify v1.4.9 // indirect
```
lubronzhan, line 6: did you run go mod tidy? I don't see go.sum change

cmd/vsphere-cloud-controller-manager/main.go
lubronzhan: Good question, could you add comment to the cod as well?

go.mod
HanFa: yes, this is after `go mod tidy`. You can see there is fsnotify at go.sum
```
gopkg.in/fsnotify.v1 v1.4.7/go.mod h1:Tz8NjZHkW78fSQdbUxIjBTcgA1z1m8ZHf0WmKUhAMys=
```

cmd/vsphere-cloud-controller-manager/main.go
HanFa: sure!

lubronzhan (approved):
/lgtm

Issue #533, opened 2021-12-02
vsphere-cloud-controller-manager version is not shown in the log correctly


alex-vmw:
### What happened?

vsphere-cloud-controller-manager version is shown as `${VERSION}` in the log:
```
I1201 10:00:32.996286       1 main.go:81] vsphere-cloud-controller-manager version: ${VERSION}
```

### What did you expect to happen?

Expected to see a version number instead of `${VERSION}`.

### How can we reproduce it (as minimally and precisely as possible)?

Start vsphere-cloud-controller-manager 1.22.3 and look at its log.

### Kubernetes version

<details>
1.22.3
</details>

### Cloud provider or hardware configuration

<details>
vSphere
</details>

### OS version

<details>
Ubuntu 20.04.3 LTS
</details>

### Kernel (e.g. `uname -a`)

<details>
5.4.0-88-generic
</details>

### Install tools

<details>

</details>

### Container runtime (CRI) and and version (if applicable)

<details>

</details>

### Related plugins (CNI, CSI, ...) and versions (if applicable)

<details>

</details>

### Others

<details>

</details>

nicolehanjing:
I think the value comes from https://github.com/kubernetes/cloud-provider-vsphere/blob/master/cmd/vsphere-cloud-controller-manager/main.go#L81, where [makefile](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/Makefile#L76) passes the version value. 

I checked the value by running: 
```
$ make version
v1.22.3
```

So even the version value is correct, the problem is that we cannot pass it via MakeFile. Do you have any context for it? @akutz

akutz:
Hi @nicolehanjing ,

I guarantee at one point in time a Go build flag was used to set `main.version` at https://github.com/kubernetes/cloud-provider-vsphere/blob/master/cmd/vsphere-cloud-controller-manager/main.go#L54.

akutz:
In fact, it should still set the version via https://github.com/kubernetes/cloud-provider-vsphere/blob/0ce006108f01994e3c7b4b35dde1162262807c93/Makefile#L76.

What I bet is happening is the official binary is being built where there are no git tags and thus no version is calculated.

akutz:
It still embeds the version when built locally:

1. Clone the repo
2. Run `make` and one can see the following in the output:
    ```bash
[...]

PR #535, opened 2021-12-03
show vsphere-cloud-controller-manager version in the log correctly

nicolehanjing:
vsphere-cloud-controller-manager version is shown as ${VERSION} in the log:
```
I1201 10:00:32.996286       1 main.go:81] vsphere-cloud-controller-manager version: ${VERSION}
```
we should see a version number instead of `${VERSION}`.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/kubernetes/cloud-provider-vsphere/issues/533

verified the fix:
```
I1203 00:09:19.942725       1 main.go:81] vsphere-cloud-controller-manager version: 1.22.3
```

**Release note**:

```release-note
show vsphere-cloud-controller-manager version in the log correctly
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -327,4 +327,5 @@ squash:
 docker-image:
 	docker build \
 	-f cluster/images/controller-manager/Dockerfile \
-	-t "$(IMAGE):$(BRANCH_NAME)" . \
+	-t "$(IMAGE):$(BRANCH_NAME)" \
+	--build-arg "VERSION=${VERSION}" . \
--- a/cluster/images/controller-manager/Dockerfile
+++ b/cluster/images/controller-manager/Dockerfile
@@ -33,7 +33,7 @@ ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:9b60270ec0991bc4f14bda475e8
 FROM ${GOLANG_IMAGE} as builder
 
 # This build arg is the version to embed in the CPI binary
-ARG VERSION=unknown
+ARG VERSION=1.22.3
 
 # This build arg controls the GOPROXY setting
 ARG GOPROXY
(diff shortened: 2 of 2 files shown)


akutz:
I think this fix is a good start, but shouldn't the fix be passing the version into the Docker build command as a build arg instead of hard coding it?

nicolehanjing:
> I think this fix is a good start, but shouldn't the fix be passing the version into the Docker build command as a build arg instead of hard coding it?

I think either works - using the fixed value and updating it when we have a new version released; or fetch the version value via git commands
I think using a fixed value can help to prevent possible issues in some other environments, and I'm planning to add this in my WIP release doc to inform the maintainer to update the version value when a new release version is available.

Makefile
```diff
@@ -327,4 +327,5 @@ squash:
 docker-image:
 	docker build \
 	-f cluster/images/controller-manager/Dockerfile \
-	-t "$(IMAGE):$(BRANCH_NAME)" . \
+	-t "$(IMAGE):$(BRANCH_NAME)" \
+	--build-arg "VERSION=${VERSION}" . \
```
nicolehanjing, line 331: pass version info dynamically when building docker image

cluster/images/controller-manager/Dockerfile
```diff
@@ -33,7 +33,7 @@ ARG DISTROLESS_IMAGE=gcr.io/distroless/static@sha256:9b60270ec0991bc4f14bda475e8
 FROM ${GOLANG_IMAGE} as builder
 
 # This build arg is the version to embed in the CPI binary
-ARG VERSION=unknown
+ARG VERSION=1.22.3
```
nicolehanjing, line 36: for local development, if you run `docker build` command without version info, the log is still not correct. Probably we can leave the fixed value and update it when we have a new release version

PR #544, opened 2021-12-07
chart: allow configuration of configmap and secret name

jwhb:
**What this PR does / why we need it**: This PRs allow to set the name of created `Secret` and `ConfigMap` objects.

Regarding ConfigMap: The chart had a mismatch between DaemonSet volume (requesting ConfigMap volume by `name: cloud-config`) and the ConfigMap creation (with `name: vsphere-cloud-config`). With this PR the ConfigMap name will default to `cloud-config` if config is disabled. With config enabled it will default to  `vsphere-cloud-config` and create this ConfigMap.

Regarding Secret: A Kubernetes platform might create the vSphere secret in the correct syntax, so we may want to reuse this secret and not create a duplicate with this Helm chart. For OpenShift on vSphere I can set `config.secret.create: false` and `config.secret.name: vsphere-creds` when deploying into `kube-system` namespace.

**Which issue this PR fixes**
No issue reported.

Please **make a Helm Chart release**. I can create another PR if that is desired.

**Release note**:

```release-note
allow setting name of secret and configmap with config.name, config.secret.name
allow disabling creation of vSphere secret with config.secret.create
```

### DIFF
--- a/charts/vsphere-cpi/templates/configmap.yaml
+++ b/charts/vsphere-cpi/templates/configmap.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: vsphere-cloud-config
+  name: {{ .Values.config.name | default "cloud-config" }}
   labels:
     app: {{ template "cpi.name" . }}
     vsphere-cpi-infra: cloud-config
@@ -16,7 +16,7 @@ data:
       # set insecure-flag to true if the vCenter uses a self-signed cert
       insecureFlag: true
       # settings for using k8s secret
-      secretName: vsphere-cloud-secret
+      secretName: {{ .Values.config.secret.name }}
       secretNamespace: {{ .Release.Namespace }}
 
     # vcenter section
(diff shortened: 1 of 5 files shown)


charts/vsphere-cpi/templates/configmap.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: vsphere-cloud-config
+  name: {{ .Values.config.name }}
```
lubronzhan, line 5: Since we have cloud-config as the default name for configmap, could you also add it here?

lubronzhan:
Thank you @jwhb!
We will also try add some helm chart unit test as well!

charts/vsphere-cpi/templates/configmap.yaml
jwhb: Just added static inline defaults in secret.yaml and configmap.yaml. Thanks!

lubronzhan (approved):
/lgtm

PR #557, opened 2021-12-21
Adds implementation for returning zones for nodes.

jvrahav:
This change implements zone interface
Here we rely on the labels on the virtual machine object returned from VMOperator to determine the zone.
we use the zone returned from the label to populate the faultdomain in zone instance and return the zone.

**Release note**:
```
Adds zones support for nodes
```


dougm (approved):
/ok-to-test
/lgtm

PR #558, opened 2021-12-21
Bump cloud-provider-vsphere to k8s v1.23

Fedosin:
Bump cloud-provider-vsphere to k8s v1.23

**Release note**:
```release-note
Bump cloud-provider-vsphere to k8s v1.23
```


go.mod
```diff
@@ -1,14 +1,14 @@
 module k8s.io/cloud-provider-vsphere
 
-go 1.16
+go 1.17
```
lubronzhan, line 3: Could you set it to 1.16 still. Since I saw k8s is still referring to 1.16
https://github.com/kubernetes/kubernetes/blob/v1.23.1/go.mod#L9

Fedosin: Unfortunately I can't, because some libraries explicitly require golang 1.17: https://github.com/kubernetes-sigs/controller-runtime/blob/master/go.mod#L3

Without this bump the code wouldn't compile.

PR #560, opened 2022-01-11
Return an error if multiple VMs with the same hostname was found

lobziik:
In case if multiple VMs with the same hostname exists
within the same scope, picking the wrong VM is quite possible,
which might lead to hard-diagnosable issues with
Nodes initialization, or even to set up incorrect values (ProviderID, addresses) on a Node object.

For avoid ambiguity `FindByDnsName` call was replaced to `FindAllByDnsName`,
`GetVMByDNSName` function now returns an error in case if
multiple VMs with the same HostName was found.

--
**Release note**:

```release-note
Added check for multiple VMs with the same hostname during initial node discovery. In case if multiple VMs with the same hostname exists this would be considered as an error.
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager_test.go
+++ b/pkg/cloudprovider/vsphere/nodemanager_test.go
@@ -31,6 +31,7 @@ import (
 
 	pb "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/proto"
 	cm "k8s.io/cloud-provider-vsphere/pkg/common/connectionmanager"
+	"k8s.io/cloud-provider-vsphere/pkg/common/vclib"
 )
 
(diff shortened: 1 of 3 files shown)


lubronzhan (commented):
/lgtm
@HanFa could you help take a look

lubronzhan:
It's probably failing because of our recent golang bump. @HanFa could you fix this in another PR

lubronzhan:
Hi @lobziik could you rebase your branch again? @HanFa just fixed the failing test

HanFa:
thanks for the PR @lobziik! I've fixed the staticcheck. could you rebase your PR to the latest as @lubronzhan suggests?

lobziik:
Rebased

PR #573, opened 2022-01-25
[Cherry-Pick] Deprecate pod priority annotation in cpi deployment

### LINKED ISSUE #525: Need `priorityClassName: system-node-critical` on pod spec
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
There is no priority class in the pod spec 
```
priorityClassName: system-node-critical
```

When there is high load on the control plane node, the cpi pod will be evicted because it doesn't have high priority

**What you expected to happen**:
cpi pod shouldn't be evicted when there is high load on the control plane node

**How to reproduce it (as minimally and precisely as possible)**:
Deploy a cluster with any cpi version  <= 1.22, scale coredns pod to 500

**Anything else we need to know?**:

**Environment**:
- vsphere-cloud-controller-manager version: 
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

XudongLiuHarold:
Signed-off-by: Xudong Liu <xudongl@vmware.com>

Since the priority annotation has been removed from k8s 1.12+ and we have already adopt the priority class in this PR: #569, deprecate all the pod priority annotations in CPI manifest.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: follow up fixes #525

**Release note**:

```release-note
Deprecate `scheduler.alpha.kubernetes.io/critical-pod` annotation in cpi manifest.
```

### DIFF
--- a/charts/vsphere-cpi/templates/daemonset.yaml
+++ b/charts/vsphere-cpi/templates/daemonset.yaml
@@ -9,7 +9,6 @@ metadata:
     tier: control-plane
   namespace: {{ .Release.Namespace }}
   annotations:
-    scheduler.alpha.kubernetes.io/critical-pod: ""
     {{- if .Values.daemonset.annotations }}
     {{- toYaml .Values.daemonset.annotations | nindent 4 }}
     {{- end }}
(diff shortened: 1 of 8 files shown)


lubronzhan:
Could you also update 1.23 manifest?

PR #643, opened 2022-08-23
Use full resource path for ipclaim and ipaddress in e2e CI.

HanFa:

Use full resource path for IPClaim and IPAddress in the CPI e2e CI, to avoid CR name conflicts with other group such as MetalLB or Calico

**Release note**:

```release-note
Use full resource path for ipclaim and ipaddress in e2e CI.
```


srm09 (changes requested):
This is missing the definition of `append_api_group` bash function

PR #702, opened 2023-03-13
Move credentials from configmap to secret

erkanerol:

Configmap shouldn't contain any sensitive data. Now, the chart renders credentials in the configmap by default, which is a severe security issue.

This PR replaces the config map with a secret.

Alternative fix: https://github.com/kubernetes/cloud-provider-vsphere/pull/703

**Release note**:
```release-note
Use secret instead of configmap for configuration including credentials.
```


erkanerol:
closing in favor of https://github.com/kubernetes/cloud-provider-vsphere/pull/703

PR #725, opened 2023-05-17
prefer addresses that are statically configured in VM metadata

### LINKED ISSUE #723: Statically configured addresses for VMs should be preferred for External/Internal Node IPs
Currently the [nodemanager code](https://github.com/kubernetes/cloud-provider-vsphere/blob/950d57d6945df5c8308a9a03d096de0f3a0e6316/pkg/cloudprovider/vsphere/nodemanager.go#L376) will look at IPs as they are on the VMs Guest.Net property to find which IP should be assigned for the External or Internal Node IPs. This will evaluate in order they are returned, preferring whichever is the first to meet any match criteria in the API response.

This causes us some trouble when using IPv6 and static addresses. The devices get several autoconfigured addresses that can take precedence over a statically configured addresses, even when DHCP6 is disabled.

We would like to adjust the current behavior to prefer any addresses that are statically configured by looking at the ExtraConfig `guestinfo.metadata` for the VM. Any addresses in the `network.ethernets[*].addresses` field for a given device should be preferred as the External or Internal Node IP, so long as it also meets any configured criteria in the CPI.

This could represent a change in the existing behavior as we are changing the precedence for address selection. We wanted to clear any concerns with this approach here. Though in the IPv6 case this current behavior can lead to issues with the cluster and could be considered a bug. It isn't clear to us if there is an existing case where this behavior would be expected.

adobley:
lookup guestinfo metadata from vSphere and prefer those addresses when setting the External and Internal Node IPs

This fixes an issue around autoconf'd addresses being assigned to the External and Internal when using IPv6

Fixes #723

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -18,11 +18,14 @@ package vsphere
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net"
+	"sort"
 	"strings"
 
+	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
 	ccfg "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphere/config"
 	vcfg "k8s.io/cloud-provider-vsphere/pkg/common/config"
@@ -48,6 +51,15 @@ var (
 	// ErrVMNotFound is returned when the specified VM cannot be found.
 	ErrVMNotFound = errors.New("VM not found")
 )
+
+type cloudInitConfig struct {
+	Network struct {
+		Ethernets map[string]struct {
+			Name      string   `yaml:"set-name"`
+			Addresses []string `yaml:"addresses"`
+		} `yaml:"ethernets"`
+	} `yaml:"network"`
+}
 
 func newNodeManager(cfg *ccfg.CPIConfig, cm *cm.ConnectionManager) *NodeManager {
 	return &NodeManager{
@@ -191,7 +203,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 	}
 
 	var oVM mo.VirtualMachine
-	err = vmDI.VM.Properties(ctx, vmDI.VM.Reference(), []string{"guest", "summary"}, &oVM)
+	err = vmDI.VM.Properties(ctx, vmDI.VM.Reference(), []string{"guest", "summary", "config"}, &oVM)
 	if err != nil {
 		klog.Errorf("Error collecting properties for vm=%+v in vc=%s and datacenter=%s: %v",
 			vmDI.VM, vmDI.VcServer, vmDI.DataCenter.Name(), err)
@@ -291,10 +303,17 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		return fmt.Errorf("unable to find suitable IP address for node after filtering out localhost IPs")
 	}
 
+	sortedNonLocalhostIPs, err := sortStaticallyConfiguredAddressesFirst(oVM.Config.ExtraConfig, nonLocalhostIPs)
+	if err != nil {
+		klog.Errorf("Error sorting statically configured addresses for vm=%+v in vc=%s and datacenter=%s: %v",
+			vmDI.VM, vmDI.VcServer, vmDI.DataCenter.Name(), err)
+		return err
+	}
+
 	for _, ipFamily := range ipFamilies {
-		klog.V(6).Infof("ipFamily: %q nonLocalhostIPs: %q", ipFamily, nonLocalhostIPs)
+		klog.V(6).Infof("ipFamily: %q nonLocalhostIPs: %q", ipFamily, sortedNonLocalhostIPs)
 		discoveredInternal, discoveredExternal := discoverIPs(
-			nonLocalhostIPs,
+			sortedNonLocalhostIPs,
 			ipFamily,
 			internalNetworkSubnets,
 			externalNetworkSubnets,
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -284,6 +286,58 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+
+	if guestInfo != "" && encoding == "base64" {
+		value, err := base64.StdEncoding.DecodeString(guestInfo)
+		if err != nil {
+			klog.Errorf("Error decoding guestinfo for vm=%+v in vc=%s and datacenter=%s: %v",
+				vmDI.VM, vmDI.VcServer, vmDI.DataCenter.Name(), err)
+			return err
+		}
+
+		type cloudInitConfig struct {
```
XudongLiuHarold, line 299: Could we move this declaration out of the function?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -284,6 +286,58 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
...
+			for _, address := range eth.Addresses {
+				guestInfoAddresses = append(
+					guestInfoAddresses,
+					net.ParseIP(strings.Split(address, "/")[0]))
+			}
+		}
+
+		// We want to prefer addresses that are from the guestInfo but only if
+		// they are on a nic already
+		ips := []*ipAddrNetworkName{}
```
XudongLiuHarold, line 327: #L327 to #L337 is trying to find the `guestInfoAddress` in `nonLocalhostIPs`:

 I'd suggest that we use a `map[net.IP]bool` to store the `guestInfoAddresses` and check if `nonLocalhostIPs` contains the IP or not which can reduce the time complexity and make the code easier to understand.

pkg/cloudprovider/vsphere/nodemanager.go
flawedmatrix: Done 👍

pkg/cloudprovider/vsphere/nodemanager.go
flawedmatrix: We decided to rewrite the code to prioritize readability so we changed the implementation to a sort to better describe the intent and extracted the code into a separate function.

XudongLiuHarold (commented):
/lgtm

The PR looks good to me, but I'd like to ask @lubronzhan  also taking a look.

lubronzhan:
Hi @adobley , is this trying to resolve issue when customer use IPAM to provide IP for Node while IPV6 also allocate slaac IP which could appear before the static ip?

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -652,3 +671,59 @@ func (nm *NodeManager) getNodeNameByUUID(UUID string) string {
...
+
+		guestInfo := &cloudInitConfig{}
+		err = yaml.Unmarshal(value, guestInfo)
+		if err != nil {
+			return nil, err
+		}
+
+		// Map of guestInfo IP -> index that describes the order they appear in the guestInfo
+		guestInfoAddresses := make(map[string]int)
+		for _, eth := range guestInfo.Network.Ethernets {
```
lubronzhan, line 710: Could we check whether Network is also not nil

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -652,3 +671,59 @@ func (nm *NodeManager) getNodeNameByUUID(UUID string) string {
 	}
 	return ""
 }
+
+func extractGuestInfo(extraConfig []types.BaseOptionValue) (string, string) {
```
lubronzhan, line 675: nit: I think it could be more accurate if renamed to `guestInfoMetadata`

pkg/cloudprovider/vsphere/nodemanager_test.go
```diff
@@ -1883,3 +2105,23 @@ func TestExcludeLocalhostIPs(t *testing.T) {
 		t.Errorf("failure: expected ipAddr to equal fd00:100:64::1, but was %s", actual[1].ipAddr)
 	}
 }
+
+func guestInfoWithAddresses(addresses string) string {
```
lubronzhan, line 2109: Maybe add another case for DHCP case? 
This is the guestInfo with DHCP machine
```
kubo@OrBVACRlEwEIY:~$ echo "Cmluc3RhbmNlLWlkOiAidGtnLW1nbXQtdmMtbWQtMC1uNWp4Zy01YzZiODc3NGN4a245NDYtNHdtN3ciCmxvY2FsLWhvc3RuYW1lOiAidGtnLW1nbXQtdmMtbWQtMC1uNWp4Zy01YzZiODc3NGN4a245NDYtNHdtN3ciCndhaXQtb24tbmV0d29yazoKICBpcHY0OiBmYWxzZQogIGlwdjY6IHRydWUKbmV0d29yazoKICB2ZXJzaW9uOiAyCiAgZXRoZXJuZXRzOgogICAgaWQwOgogICAgICBtYXRjaDoKICAgICAgICBtYWNhZGRyZXNzOiAiMDA6NTA6NTY6OTM6ZTI6NDIiCiAgICAgIHNldC1uYW1lOiAiZXRoMCIKICAgICAgd2FrZW9ubGFuOiB0cnVlCiAgICAgIGRoY3A0OiBmYWxzZQogICAgICBkaGNwNjogdHJ1ZQo=" |base64 -d

instance-id: "tkg-mgmt-vc-md-0-n5jxg-5c6b8774cxkn946-4wm7w"
local-hostname: "tkg-mgmt-vc-md-0-n5jxg-5c6b8774cxkn946-4wm7w"
wait-on-network:
  ipv4: false
  ipv6: true
network:
  version: 2
  ethernets:
    id0:
      match:
        macaddress: "00:50:56:93:e2:42"
      set-name: "eth0"
      wakeonlan: true
      dhcp4: false
      dhcp6: true
```

pkg/cloudprovider/vsphere/nodemanager.go
christianang: Network is not a pointer so it shouldn't be nil in this case.

christianang:
> Hi @adobley , is this trying to resolve issue when customer use IPAM to provide IP for Node while IPV6 also allocate slaac IP which could appear before the static ip?

Yes, this is trying to fix this case.

pkg/cloudprovider/vsphere/nodemanager_test.go
adobley: We added a test for this case, thanks!

pkg/cloudprovider/vsphere/nodemanager.go
adobley: We've renamed it, thanks for the review!

pkg/cloudprovider/vsphere/nodemanager.go
lubronzhan: Oh sorry you are right

lubronzhan (approved):
/lgtm
/approve

PR #726, opened 2023-05-25
Fix tls thumbprint parameter in the helm chart

erkanerol:
This PR fixes helm chart for tls thumbprint parameter. The current chart is wrong.

The controller respects `thumbprint`, not `tlsThumbprint`.

https://github.com/kubernetes/cloud-provider-vsphere/blob/a7af18bcabc6b79b559aff6a8f84b17b9491ac9c/pkg/common/config/config_ini_legacy.go#L47

Since the current parameter doesn't work because of [wrong templating](https://github.com/kubernetes/cloud-provider-vsphere/pull/726#discussion_r1205388077) and the wrong parameter name, I think it is OK to fix the parameter too. I believe there is no one who relies on this feature since it is already broken. 

**Release note**:
```release-note
Fix tls thumbprint parameter in the helm chart
```

### DIFF
--- a/charts/vsphere-cpi/templates/configmap.yaml
+++ b/charts/vsphere-cpi/templates/configmap.yaml
@@ -13,8 +13,8 @@ data:
     # Global properties in this section will be used for all specified vCenters unless overriden in VirtualCenter section.
     global:
       port: 443
-      {{- if .Values.config.tlsThumbprint }}
-      tlsThumbprint: {{- .Values.config.tlsThumbprint }}
+      {{- if .Values.config.thumbprint }}
+      thumbprint: {{ .Values.config.thumbprint }}
       {{- else }}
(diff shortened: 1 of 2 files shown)


lubronzhan:
Ok I found it in their release note
```
Updated runtime and libraries to v0.6.0
Update runtime to v0.6.0 by @aagrawal3 in https://github.com/vmware/vsphere-automation-sdk-go/pull/19
Updated lib to v0.6.0 by @aagrawal3 in https://github.com/vmware/vsphere-automation-sdk-go/pull/20
Removed the ConverterMode field of bindings.TypeConverter as it is no longer needed. Also removed SetMode(), Mode() and InRestMode().
```
Thanks for the PR
/lgtm
/approve

lubronzhan:
Looks good! Thanks
/lgtm
/approve

PR #734, opened 2023-06-26
Improve helm chart to support global values

erkanerol:
### Background

A helm chart can be used as sub chart in another chart. The parent chart can provide values for subcharts explicitly and it doesn't require anything in the subchart. On the other hand, when the parent chart wants to provide a common set of values to all subcharts, it is not easy. The parent chart has to set `.global` and all subcharts have to respect `.global` explicitly. See https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

While developing a helm chart, if you want 
- Your chart should work standalone
- Your chart should respect `.global` if exists

at the same time, there is no proper solution. You need to implement the logic in your chart.

### Why did I open this PR?

Current chart has code pieces as below to achieve this goal. 

```
 .Values.config.region | default .Values.global.config.region
```

**Problems**
- The order is not right in the current chart. Parent chart should be able to overwrite config via `.global`. Therefore, default should be `.config` and `.global.config` should overwrite it if exists.
- Config merge is done for every parameter in templates. It is error-prone and the logic is missing for some configs right now. See `.config.thumbprint ` and `.config.secret.name`.

### Note for reviewers
I tried to find a better solution. It was great if it was possible to do the merge once and use it in every template but there is no such solution. See https://kubernetes.slack.com/archives/C0NH30761/p1686828259395319

**Release note**:
```release-note
Improve helm chart to support global values so that this chart can be used as subchart.
```

### DIFF
--- a/charts/vsphere-cpi/templates/configmap.yaml
+++ b/charts/vsphere-cpi/templates/configmap.yaml
@@ -1,8 +1,14 @@
-{{- if .Values.config.enabled | default .Values.global.config.enabled -}}
+{{- $config := .Values.config -}}
+{{- if .Values.global }}
+{{- if .Values.global.config }}
+{{- $config = mergeOverwrite (deepCopy .Values.config) .Values.global.config -}}
+{{- end }}
+{{- end }}
+{{- if $config.enabled -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ .Values.config.name | default "cloud-config" }}
+  name: {{ $config.name | default "cloud-config" }}
   labels:
     app: {{ template "cpi.name" . }}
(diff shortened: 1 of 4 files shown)


lubronzhan:
Thanks!
/ok-to-test

erkanerol:
@lubronzhan Thanks for running the tests. Are you OK with this PR? Is there anything I should do?

lubronzhan:
Let me try the new template locally. Thanks

lubronzhan:
e2e test didn't get triggered automatically so I updated the test config here https://github.com/kubernetes/test-infra/pull/30126
And I manually tested your change locally. 
Thanks for the contribution!

Issue #736, opened 2023-07-10
With golang 1.20, govmomi will report issue validating certificate


lubronzhan:
### What happened?

Opened an issue in govmomi as well
https://github.com/vmware/govmomi/issues/3174

Right now if i compile from master branch with golang 1.20, even with correct thumbprint config, govmomi will report error like 

```
E0710 05:47:19.253974       1 connectionmanager.go:147] Cannot connect to vCenter with err: Post "https://10.191.133.186:443/sdk": tls: failed to verify certificate: x509: cannot validate certificate for 10.191.133.186 because it doesn't contain any IP SANs
E0710 05:47:20.258996       1 connection.go:181] Failed to create new client. err: Post "https://10.191.133.186:443/sdk": tls: failed to verify certificate: x509: cannot validate certificate for 10.191.133.186 because it doesn't contain any IP SANs
E0710 05:47:20.259047       1 connection.go:67] Failed to create govmomi client. err: Post "https://10.191.133.186:443/sdk": tls: failed to verify certificate: x509: cannot validate certificate for 10.191.133.186 because it doesn't contain any IP SANs
E0710 05:47:20.259058       1 connectionmanager.go:147] Cannot connect to vCenter with err: Post "https://10.191.133.186:443/sdk": tls: failed to verify certificate: x509: cannot validate certificate for 10.191.133.186 because it doesn't contain any IP SANs
```

### What did you expect to happen?

CPI should be working without certificate validation error when using golang 1.20

### How can we reproduce it (as minimally and precisely as possible)?

1. Bumped Cloud-provider-vsphere to use golang1.20
2. Govmomi version is 0.30.4
3. Then compile and deploy Cloud-provider-vsphere with correct tlsthumbprint config
4. Error shows up

```
E0710 05:47:20.259058       1 connectionmanager.go:147] Cannot connect to vCenter with err: Post "https://10.191.133.186:443/sdk": tls: failed to verify certificate: x509: cannot validate certificate for 10.191.133.186 because it doesn't contain any IP SANs
```

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Cloud provider or hardware configuration

<details>

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

[...]

PR #738, opened 2023-07-18
Bump govmomi to 0.30.6

lubronzhan:
Need govmomi fix https://github.com/vmware/govmomi/issues/3174 for validating certificate's tlsthumbprint to work properly

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/kubernetes/cloud-provider-vsphere/issues/736

**Release note**:

```release-note
Bump govmomi to 0.30.6
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.2
 	github.com/vmware-tanzu/vm-operator-api v0.1.4-0.20201118171008-5ca641b0e126
-	github.com/vmware/govmomi v0.30.4
+	github.com/vmware/govmomi v0.30.6
 	github.com/vmware/vsphere-automation-sdk-go/lib v0.7.0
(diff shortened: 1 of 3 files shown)


liggitt:
Does this also need to be updated in kubernetes/kubernetes for go1.20? That is still using v0.30.0 (see https://github.com/kubernetes/kubernetes/pull/114766 and backports)

liggitt:
cc @MadhavJivrajani

lubronzhan:
cc @wyike it should be needed

MadhavJivrajani:
@liggitt I think we'll have to bump and back-port. This issue didn't surface directly in our CI.

We did a similar fix but in our tests: https://github.com/kubernetes/kubernetes/commit/c31cc5ec46315a02343ec6d6a2ef659e2cc8668e
and the reason it worked in our CI was because the function being relies on text matching in the error string.

PR #788, opened 2023-12-12
Allow setting specific useragent identifier

### LINKED ISSUE #787: Adding a clusterID and using on useragent

Hey folks, 

During some debugging with CSI and CAPV, we've figured out some sessions being opened a lot to vCenter and VAPI, and in fact we got some improvements from it.

One thing that was really helpful on CSI side was the fact that they [have implemented the clusterID](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/02766e6bc4032b233eadf8cd9b42e3b736b46a9f/pkg/common/config/config.go#L776) as part of the userAgent, so listing vc sessions (govc session.ls) allowed us to summarize how many connections each cluster/CSI was opening, if the session was stuck, if it was same userAgent and different nodes (some leader election issue?), etc.

I would like to request that we consider the same for vSphere CPI. I couldn't find on the code any reference on setting the clusterID as a config, so I wanted to understand if:
* We can add this extra configuration on yaml config (and make it clear that right now this information is only used for userAgent)
* We can mutate userAgent to, in case clusterID is set, being part of its composition

Thanks

### Why is this needed?

Facilitate investigation on VC session leaks, stale connections, add better observation on who/what cluster is consuming connections

rikatz:
**What this PR does / why we need it**: During some scalability tests, we have figured out that having a unique identifier per cluster would make it easier to identify how many connections a cluster is opening to vSphere, and improve this workflow. Also it is good to know exactly what cluster has open connections

**Which issue this PR fixes**: fixes #787 

**Release note**:

```release-note
Allow specifying new ClusterID on configuration
```

### DIFF
--- a/pkg/common/config/config_yaml.go
+++ b/pkg/common/config/config_yaml.go
@@ -37,6 +37,7 @@ func (ccy *CommonConfigYAML) CreateConfig() *Config {
 		VirtualCenter: make(map[string]*VirtualCenterConfig),
 	}
 
+	cfg.Global.ClusterID = ccy.Global.ClusterID
 	cfg.Global.User = ccy.Global.User
 	cfg.Global.Password = ccy.Global.Password
 	cfg.Global.VCenterIP = ccy.Global.VCenterIP
@@ -88,7 +89,7 @@ func (vccy *VirtualCenterConfigYAML) isSecretInfoProvided() bool {
 }
 
 func (ccy *CommonConfigYAML) validateConfig() error {
-	//Fix default global values
+	// Fix default global values
 	if ccy.Global.RoundTripperCount == 0 {
 		ccy.Global.RoundTripperCount = DefaultRoundTripperCount
 	}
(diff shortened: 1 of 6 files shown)


lubronzhan:
BTW, i just found out that cloud-provider has a cluster-name field https://github.com/kubernetes/cloud-provider/blob/152995bd25ae20bc618a9ba7039a1b683372c86c/config/types.go#L71
 This was used for serviceController previously. Sorry I didn't found it before. But yeah you make it clear inside the description of the new field.

Could you also add to ini type config as well? Downstream product still use that one https://github.com/rikatz/cloud-provider-vsphere/blob/different-user-agent/pkg/common/config/config_ini_legacy.go
But it just reminds me we should switch over to new type of config

rikatz:
> Or do you think we can just previous clusterName value?
The goal is to identify it on userAgent, so if clusterName does the job, why not? :) I will work on it next week, thanks!!

lubronzhan:
Yeah I think you are right, this is the flag added before https://github.com/kubernetes/cloud-provider-vsphere/blob/master/cmd/vsphere-cloud-controller-manager/main.go?rgh-link-date=2023-12-19T13%3A39%3A42Z#L135
And that is passed to the cloud-provider's ClusterName as a config. 
Sorry I didn't find the right one before

Issue #786, opened 2023-11-23
Restart paravirtual CPI POD if /etc/cloud/ccm-provider related secret changes 


wyike:
### What would you like to be added?

In paravirtual mode, a secret container supervisor cluster access token is mounted to CPI pod.
We want to add a mechanism to watch that secret and restart CPI Pod if its updated

### Why is this needed?

The access token could be rotated for example by kubelet. The new secret watch mechanism can avoid unnecessary failures from CPI Pod when accessing supervisor cluster.

PR #793, opened 2024-01-05
[Fix]Monitor the path of supervisor config and restart if it's updated

lubronzhan:
If cert or token of supervisor is updated, CPI should reload the config

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/kubernetes/cloud-provider-vsphere/issues/786

**Release note**:

```release-note
Restart CPI container If cert or token of supervisor is updated, 
```

### DIFF
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -50,8 +50,12 @@ import (
 	"github.com/spf13/pflag"
 )
 
-// AppName is the full name of this CCM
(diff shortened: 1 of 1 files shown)


lubronzhan:
Tested here  and container restarted if there is change to that folder
```
F0105 01:24:50.975187       1 main.go:253] restarting pod because received event CREATE        "/etc/cloud/ccm-provider/..2024_01_05_01_24_50.689340044"
```

PR #809, opened 2024-01-11
Cleanup VSphere sessions on SIGTERM

flawedmatrix:

This PR adds some cleanup logic for when the cloud-provider-vsphere is sent a SIGTERM signal (like in the case of when a pod is terminated). It will close the stop channel, ensuring that logic that relies on the stop channel is notified of the shutdown event, and it will ensure that VSphere sessions are closed when the program reaches the termination state.

**Release note**:

```release-note
Ensure VSphere sessions are cleaned up when Kubernetes pod is terminated.
```

### DIFF
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -25,7 +25,9 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	cloudprovider "k8s.io/cloud-provider"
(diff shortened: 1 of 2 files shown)


flawedmatrix:
cc @christianang

cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -191,11 +193,23 @@ func main() {
...
+			close(stop)
+		}()
+
 		if err := app.Run(completedConfig, cloud, controllerInitializers, webhookHandlers, stop); err != nil {
 			// explicitly ignore the error by Fprintf, exiting anyway due to app error
 			_, _ = fmt.Fprintf(os.Stderr, "%v\n", err)
 			os.Exit(1)
 		}
+		// Log out of all sessions on exit or signal
+		vsphere.SessionLogout()
```
lubronzhan, line 212: Question: This one is added here, not in the goroutine, is it because it needs to notify the init() func to logout all sessions and can block the main.go? Just curious, if app.Run exit with error will this code still get executed?

pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -113,6 +127,17 @@ func (vs *VSphere) Initialize(clientBuilder cloudprovider.ControllerClientBuilde
 		vs.connectionManager = connMgr
 		vs.nodeManager.connectionManager = connMgr
 
+		logoutWG.Add(1)
+		// Gracefully logout of all VSphere sessions if the stop channel is signaled
+		go func() {
+			// TODO: The stop channel is unusable since the cloud-provider module
```
lubronzhan, line 133: Could you add it's because of this line https://github.com/kubernetes/cloud-provider/blob/1bae60eb89ced16795f81a900b79cb55524ba6f0/app/controllermanager.go#L536
Not using `ctx := wait.ContextForChannel(stopCh)` I assume?
Also would be useful to open an issue in cloud-provider repo.
Good catch!

cmd/vsphere-cloud-controller-manager/main.go
flawedmatrix: It turns out if there are errors after app initialization, the program just exits without giving us a chance to cleanup sessions. This is due to the Fatalf handling of errors k8s.io/cloud-provider, so it's a bit difficult to fix that at the moment.

pkg/cloudprovider/vsphere/cloud.go
flawedmatrix: Added a comment to point to the `context.TODO()`.

PR #905, opened 2024-03-08
run dependabot on three latest release branches

### LINKED ISSUE #865: Dependabot for release branches
# What would you like to be added

I'd like to see dependabot added for release branches. The branches are on N-2 versions of newest K8s version

# Why is this needed?

This would be useful to help automate and facilitate updating dependencies for older releases we continue to support. In particular, this would help keep us up to date with CVEs that get fixed in any of our dependencies that we should be releasing patches for.

skriss:
Configures dependabot to run on the latest three
release branches in addition to master. Only
patch bumps will be performed on the release
branches.

Also updates the docker dependabot configuration
to reference /cluster/images/controller-manager
which is where the main Dockerfile is stored.

Fixes #865

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,8 +2,12 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 version: 2
 updates:
+# #############
+# master branch
+# #############
 # GitHub Actions
-- package-ecosystem: "github-actions"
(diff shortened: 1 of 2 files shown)


skriss:
cc @lubronzhan

lubronzhan:
Thanks! Looks good to me
/ok-to-test

PR #1140, opened 2024-07-03
⚙️ fix: downgrade vm op api to fix ci

XudongLiuHarold:

**Release note**:

```release-note
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,10 @@
 module k8s.io/cloud-provider-vsphere
 
-go 1.22.4
+go 1.22.0
 
-replace github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels => github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels v1.8.7-0.20240625173706-a93aa9f0bda0
+toolchain go1.22.3
+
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,8 +1,10 @@
 module k8s.io/cloud-provider-vsphere
 
-go 1.22.4
+go 1.22.0
 
-replace github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels => github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels v1.8.7-0.20240625173706-a93aa9f0bda0
+toolchain go1.22.3
+
+replace github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels => github.com/vmware-tanzu/vm-operator/pkg/constants/testlabels v1.8.7-0.20240530162623-88d9b71d9741
```
DanielXiao, line 7: This should also be v1.8.6

XudongLiuHarold: Is it okay to have any random version, or do you have to set it to align with the vm-operator version?

DanielXiao: It does not report an error but better to align with vm-operator/api version

PR #1148, opened 2024-07-16
Move DaemonSet affinities and tolerations to values.yaml to make them user manageable.

FutureMatt:
Moves the CPI DaemonSet affinities and tolerations from daemonset.yaml to values.yaml whilst preserving the defaults.

Some environments, in our case RKEv1, apply different node taints and labels for example, ours are:
```
node-role.kubernetes.io/controlplane=true:NoExecute
node-role.kubernetes.io/etcd=true:NoSchedule
```

This change would allow us to easily adjust them.

### DIFF
--- a/charts/vsphere-cpi/values.yaml
+++ b/charts/vsphere-cpi/values.yaml
@@ -82,9 +82,6 @@ daemonset:
     - key: node.cloudprovider.kubernetes.io/uninitialized
       value: "true"
       effect: NoSchedule
-    - key: node-role.kubernetes.io/master
-      effect: NoSchedule
-      operator: Exists
     - key: node-role.kubernetes.io/control-plane
       effect: NoSchedule
       operator: Exists
--- a/charts/vsphere-cpi/values.yaml
+++ b/charts/vsphere-cpi/values.yaml
@@ -77,5 +77,31 @@ daemonset:
   ## Additional pod labels
   ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
   podLabels: {}
-  ## Allows for the default tolerations to be replaced with user-defined ones
-  tolerations: []
+  ## Allows for the default tolerations to be updated or replaced with user-defined ones
+  tolerations:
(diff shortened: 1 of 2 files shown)


charts/vsphere-cpi/values.yaml
```diff
@@ -77,5 +77,28 @@ daemonset:
...
   ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
   podLabels: {}
-  ## Allows for the default tolerations to be replaced with user-defined ones
-  tolerations: []
+  ## Allows for the default tolerations to be updated or replaced with user-defined ones
+  tolerations:
+    - key: node.cloudprovider.kubernetes.io/uninitialized
+      value: "true"
+      effect: NoSchedule
+    - key: node-role.kubernetes.io/master
```
XudongLiuHarold, line 85: Could you remove this since it is already deprecated?

XudongLiuHarold (commented):
Hi @FutureMatt, thanks for your PR!  

Do you have the requirement to not only add some tolerations but also replace them? Cuz now you can already add more tolerations by setting tolerations in `values.yaml`

/ok-to-test

FutureMatt:
> Do you have the requirement to not only add some tolerations but also replace them? Cuz now you can already add more tolerations by setting tolerations in values.yaml

Yes, annoyingly, RKE v1 by default creates control plane nodes with taints and  labels of `controlplane` not `control-plane` and we have ~20 legacy clusters that we now need to apply this to.

XudongLiuHarold:
> > Do you have the requirement to not only add some tolerations but also replace them? Cuz now you can already add more tolerations by setting tolerations in values.yaml
> 
> Yes, annoyingly, RKE v1 by default creates control plane nodes with taints and labels of `controlplane` not `control-plane` and we have ~20 legacy clusters that we now need to apply this to.

make senses to me, but we are about to release 1.31 beta, to make the release stable, I will approve this one after the beta released. :)

XudongLiuHarold (commented):
/lgtm

/approve

PR #1225, opened 2024-08-28
Removed error logging from config loader

### LINKED ISSUE #1224: Remove config yaml error log messages
### What happened?

When deploying clusters, the cluster may be using either INI or YAML configuration for the vsphere cloud provider.  All of our existing clusters (OpenShift) haven't migrated to using the newer YAML format.  This is resulting in the logging of error messages that can be misleading since the INI is present and loads correctly.  The logging in this case should either be removed in the YAML module or be made configurable so clients using the config module can suppress these messages.

An example of loading INI config with log output.

```
E0828 14:08:33.586780       1 config_yaml.go:208] Unmarshal failed: yaml: unmarshal errors:
  line 1: cannot unmarshal !!seq into config.CommonConfigYAML
W0828 14:08:33.586794       1 config.go:264] ReadConfigYAML failed: yaml: unmarshal errors:
  line 1: cannot unmarshal !!seq into config.CommonConfigYAML
I0828 14:08:33.586894       1 config.go:272] ReadConfig INI succeeded. INI-based cloud-config is deprecated and will be removed in 2.0. Please use YAML based cloud-config.
I0828 14:08:33.586924       1 config.go:283] Config initialized
```

It would be ideal to just have the error log calls be removed in [config_yaml.go](https://github.com/kubernetes/cloud-provider-vsphere/blob/de12674c6db56a858856567fc18db8b6122dad32/pkg/common/config/config_yaml.go#L197) like it is done in the [config_ini_legacy.go](https://github.com/kubernetes/cloud-provider-vsphere/blob/de12674c6db56a858856567fc18db8b6122dad32/pkg/common/config/config_ini_legacy.go#L234).

An alternative approach would be to make an Options object that can pass in various options that can trigger whether or not the logging is performed.  The default behavior can be to log all messages as we do today while allowing developers using this module to filter them out via the Options object.

### What did you expect to happen?

I would like it if the YAML config loader did not log the errors in the log like the INI module does.  This will remove the Exx message from the logs preventing any confusion of it being an error.

```
W0828 14:08:33.586794       1 config.go:264] ReadConfigYAML failed: yaml: unmarshal errors:
  line 1: cannot unmarshal !!seq into config.CommonConfigYAML
I0828 14:08:33.586894       1 config.go:272] ReadConfig INI succeeded. INI-based cloud-config is deprecated and will be removed in 2.0. Please use YAML based cloud-config.
I0828 14:08:33.586924       1 config.go:283] Config initialized
```

### How can we reproduce it (as minimally and precisely as possible)?

Create a cluster with the vSphere cloud provider config using the INI format

### Anything else we need to know (please consider providing level 4 or above logs of CPI)?

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
Kubernetes Version: v1.30.3
```

</details>

### Cloud provider or hardware configuration

<details>
VMware vSphere 8.0.2.00300
</details>

### OS version

_No response_

### Kernel (e.g. `uname -a`)

_No response_

### Install tools

_No response_

### Container runtime (CRI) and and version (if applicable)

_No response_

### Related plugins (CNI, CSI, ...) and versions (if applicable)

_No response_

### Others

_No response_

vr4manta:

- Reduces misleading error message that may cause confusion when users glance through the logs

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1224 

This PR adds a check to see if byte data is YAML.  If so, it will call the parse yaml like before allowing any errors to be logged.  I believe this is enough to just state the YAML attempt failed and is falling back onto INI config load.

**Release note**:

```
None
```

### DIFF
--- a/pkg/common/config/config.go
+++ b/pkg/common/config/config.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -255,23 +256,41 @@ func (cfg *Config) FromEnv() error {
 // ReadConfig parses vSphere cloud config file and stores it into VSphereConfig.
 // Environment variables are also checked
 func ReadConfig(byConfig []byte) (*Config, error) {
+	var cfg *Config
+
 	if len(byConfig) == 0 {
 		return nil, fmt.Errorf("Invalid YAML/INI file")
 	}
 
-	cfg, err := ReadConfigYAML(byConfig)
-	if err != nil {
-		klog.Warningf("ReadConfigYAML failed: %s", err)
+	// Check to see if yaml file before calling parser to prevent unnecessary error messages.
+	yamlErr := isConfigYaml(byConfig)
+	if yamlErr == nil {
+		cfg, yamlErr = ReadConfigYAML(byConfig)
 
-		cfg, err = ReadConfigINI(byConfig)
-		if err != nil {
-			klog.Errorf("ReadConfigINI failed: %s", err)
-			return nil, err
+		// ReadConfigYAML can fail for other config errors not related to YAML.  So it is ok for yamlErr == nil initially
+		// and a new error occurs after read.
+		if yamlErr != nil {
+			klog.Errorf("ReadConfigYAML failed: %s", yamlErr)
+			return nil, yamlErr
 		}
-
-		klog.Info("ReadConfig INI succeeded. INI-based cloud-config is deprecated and will be removed in 2.0. Please use YAML based cloud-config.")
+		klog.V(4).Info("ReadConfig YAML succeeded")
 	} else {
-		klog.Info("ReadConfig YAML succeeded")
+		klog.V(4).Infof("YAML not detected.  Attempting INI config load.")
+		var iniErr error
+		cfg, iniErr = ReadConfigINI(byConfig)
+		if iniErr != nil {
+			var cfgErr Err
+			if errors.As(iniErr, &cfgErr) {
+				// This error means it was valid INI but something was wrong w/ the config
+				return nil, iniErr
+			}
+
+			// If we are here, YAML and INI failed.  Create an error with both failure messages so user can review to determine cause
+			klog.Errorf("ReadConfigYAML failed: %v", yamlErr)
+			klog.Errorf("ReadConfigINI failed: %v", iniErr)
+			return nil, fmt.Errorf("ReadConfig failed.  YAML=[%v], INI=[%v]", yamlErr, iniErr)
+		}
+		klog.Info("ReadConfig INI succeeded. INI-based cloud-config is deprecated and will be removed in 2.0. Please use YAML based cloud-config.")
 	}
 
 	// Env Vars should override config file entries if present
--- a/pkg/common/config/config_yaml.go
+++ b/pkg/common/config/config_yaml.go
@@ -227,3 +227,14 @@ func ReadConfigYAML(byConfig []byte) (*Config, error) {
 
 	return cfg.CreateConfig(), nil
 }
+
+func isConfigYaml(byConfig []byte) error {
+	cfg := CommonConfigYAML{
+		Vcenter: make(map[string]*VirtualCenterConfigYAML),
+	}
+
+	if err := yaml.Unmarshal(byConfig, &cfg); err != nil {
+		return err
+	}
+	return nil
+}
--- a/pkg/common/config/config_yaml_test.go
+++ b/pkg/common/config/config_yaml_test.go
@@ -65,6 +65,28 @@ vcenter:
     secretNamespace: kube-system
 `
 
+const missingServerConfigYAML = `
+global:
+  port: 443
+  user: user
+  password: password
+  insecureFlag: true
+  datacenters:
+    - us-west
+  caFile: /some/path/to/a/ca.pem
+`
+
+const badConfigYAML = `
+global:
+  server: 0.0.0.0
+  port: 443
+  user: password: 
+  insecureFlag: true
+  datacenters:
+    - us-west
+  caFile: /some/path/to/a/ca.pem
+`
+
 func TestReadConfigYAMLGlobal(t *testing.T) {
 	_, err := ReadConfigYAML([]byte(""))
 	if err == nil {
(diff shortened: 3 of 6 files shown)


XudongLiuHarold:
Hi @vr4manta 

Thank you for your PR! My cents:
- It would be better to begin deprecating the INI configuration, as we plan to remove support for it in the future.

- IMO, a better approach to resolving this issue might be to add a method that checks the configuration file format: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/config.go#L261, rather than removing the error log from the YAML parser.

vr4manta:
> Hi @vr4manta
> 
> Thank you for your PR! My cents:
> 
> * It would be better to begin deprecating the INI configuration, as we plan to remove support for it in the future.
> * IMO, a better approach to resolving this issue might be to add a method that checks the configuration file format: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/common/config/config.go#L261, rather than removing the error log from the YAML parser.

Hi @XudongLiuHarold.  Thanks for the feedback.  We are migrating all our new installs to YAML, but all existing clusters are up to the admins to update the config.  If they are doing something like multi vCenter, we are requiring them to upgrade from INI to YAML by hand for now.  But all customers with one failure domain or all failure domains in one vCenter are not currently required to migrate the cloud provider config.

Based on that, I think it would be perfectly fine to do a new function that checks to see if the file is YAML format in the above linked function.  I can look into putting something there that can do a quick check.

pkg/common/config/config_yaml.go
```diff
@@ -227,3 +227,14 @@ func ReadConfigYAML(byConfig []byte) (*Config, error) {
...
+
+func isConfigYaml(byConfig []byte) (bool, error) {
+	cfg := CommonConfigYAML{
+		Vcenter: make(map[string]*VirtualCenterConfigYAML),
+	}
+
+	if err := yaml.Unmarshal(byConfig, &cfg); err != nil {
+		return false, err
+	}
+	return true, nil
```
XudongLiuHarold, line 239: change this function to only return one boolean value:

```go

err := yaml.Unmarshal(byConfig, &cfg)
return error == nil
```

pkg/common/config/config.go
```diff
@@ -255,11 +255,18 @@ func (cfg *Config) FromEnv() error {
...
 func ReadConfig(byConfig []byte) (*Config, error) {
+	var cfg *Config
+
 	if len(byConfig) == 0 {
 		return nil, fmt.Errorf("Invalid YAML/INI file")
 	}
 
-	cfg, err := ReadConfigYAML(byConfig)
+	// Check to see if yaml file before calling parser to prevent unnecessary error messages.
+	isYAML, err := isConfigYaml(byConfig)
```
XudongLiuHarold, line 265: It is better to change this to
```go
if isConfigYaml(byConfig) {
    cfg, err = ReadConfigYAML(byConfig)
   if err != nil {
   // handle error
   }
} else {
   cfg, err = ReadConfigINI(byConfig)
   if err != nil {
   // handle error
   }
}

File: pkg/common/config/config_yaml_test.go [hunk 3]
```diff
@@ -65,6 +65,17 @@ vcenter:
     secretNamespace: kube-system
 `
 
+const badConfigYAML = `
```
XudongLiuHarold, line 68: Please add more test cases to cover different inputs

- valid yaml
- valid ini
- invalid yaml
- invalid ini
- invalid format

vr4manta: ok, I'll add more tests.

pkg/common/config/config.go
vr4manta: Ok, I like that.  I'll make that change.

pkg/common/config/config.go
```diff
@@ -255,23 +255,29 @@ func (cfg *Config) FromEnv() error {
...
 		return nil, fmt.Errorf("Invalid YAML/INI file")
 	}
 
-	cfg, err := ReadConfigYAML(byConfig)
-	if err != nil {
-		klog.Warningf("ReadConfigYAML failed: %s", err)
-
+	// Check to see if yaml file before calling parser to prevent unnecessary error messages.
+	if isConfigYaml(byConfig) {
+		cfg, err = ReadConfigYAML(byConfig)
```
gnufied, line 267: So one of the problems is - may be the config was indeed YAML, but may be YAML was invalid. We can't assume that cfg was ini in that case. That is why IMO - if config parsing fails using YAML and then fails using INI too, then. we should log errors from both YAML and INI parsers.

vr4manta: I like that idea.  I can update to have both errors printed after all attempts.  I'll mock that up and see how it goes.

pkg/common/config/config.go
```diff
@@ -255,23 +256,41 @@ func (cfg *Config) FromEnv() error {
...
-		if err != nil {
-			klog.Errorf("ReadConfigINI failed: %s", err)
-			return nil, err
+		// ReadConfigYAML can fail for other config errors not related to YAML.  So it is ok for yamlErr == nil and a
+		// new error occurs
+		if yamlErr != nil {
+			klog.Errorf("ReadConfigYAML failed: %s", yamlErr)
+			return nil, yamlErr
 		}
+		klog.Info("ReadConfig YAML succeeded")
```
gnufied, line 276: `klog.Info` is pretty high level for this. This should be may be - `klog.V(4)`

pkg/common/config/config.go
```diff
@@ -255,23 +256,41 @@ func (cfg *Config) FromEnv() error {
...
+	} else {
+		klog.Infof("YAML not detected.  Attempting INI config load.")
```
gnufied, line 278: `klog.Info` is pretty high level for this. This should be may be - `klog.V(4)`

vr4manta: I am ok with changing the level.  Let me make those changes.

XudongLiuHarold:
e2e test passed: https://github.com/team-cluster-api/cloud-provider-vsphere/pull/60

elmiko:
code makes sense to me, even if the ini format is being deprecated it would be nice to reduce confusion in the logs.

vr4manta:
/assign @XudongLiuHarold 
I think this is ready for approval.

XudongLiuHarold (commented):
/lgtm

/approve

XudongLiuHarold (approved):

PR #1322, opened 2024-11-21
Added the ability to apply extra volumes, volume mounts and host aliases to the daemonset

FutureMatt:
Adds the ability for the end-user to mount extra volumes into the DaemonSet pod. In our instance we want to leverage the CA file configuration option to validate the vCenter instance we are connecting to and this needs to be mounted in the pod.

I have also added the ability to define `hostAliases` on the DaemonSet. In our case it is because the DNS for our vCenter that resolves to the cert issued is part of a different nameserver from that that our Kube clusters use, and I thought that there may be similar use cases for other teams.

I also spotted that the affinity parameters were missing from the readme so I added that too.


FutureMatt:
Hey, is some one able to take a look at this?

lubronzhan:
Looks good. Thanks @FutureMatt . Have you tried tested it in your own env?
/ok-to-test

PR #1424, opened 2025-02-17
Allow changing the default cluster.role on vmservice

rikatz:
**What this PR does / why we need it**: This PR allows changing the default behavior of VMService creation, to select a different set of Virtual Machines instead of just "node"

As of today, when creating a LoadBalancer service, just VMs with the label of cluster.role=node are selected. On some very specific cases, we may need to run a service on a control plane VM. When this happens, we should be able to select different set of VMs.

**Release note**:
```release-note
Allow creating VMService pointing to a different cluster.role
```


rikatz:
/retest

error seems not related with my change

DanielXiao:
Thank you for the PR but we are going to propagate Node endpoints from guest cluster to Supervisor cluster via new API VMServiceEndpointSlice, which will make any node as endpoint where the target pod lives.

rikatz:
@DanielXiao thanks!

I am closing this for now, as per some of our discussions we will follow a different approach.

/close

PR #1458, opened 2025-03-28
Add support for service annotation propagation

silvery1622:
What this PR does / why we need it:
Enable propagation of Service annotations from workload clusters to influence final Service creation.
Only propagates annotations when enable-service-annotation-propagation is explicitly set true.

Which issue this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close that issue when PR gets merged): fixes #

Special notes for your reviewer:


silvery1622:
/

PR #1514, opened 2025-06-20
Update Supervisor API calls to use FQDN

silvery1622:
Modifies the Supervisor API server REST calls to use the FQDN supervisor.default.svc instead of the Supervisor Endpoint IP address.
The current IP-based approach fails when the Supervisor has Nginx SNI Proxy enabled. Using the FQDN ensures compatibility regardless of whether Nginx SNI Proxy is enabled.


zhanggbj:
@silvery1622 could you please fix fmt error
```
 		Host: "https://" + net.JoinHostPort(SupervisorAPIServerFQDN, svEndpoint.Port),
 		TLSClientConfig: rest.TLSClientConfig{
-			CAData:     rootCA,
+			CAData: rootCA,
 		},
```

PR #1561, opened 2025-08-18
Fix: Tags client not returning tags when session manager is used

BernardMC:
When provisioning workload cluster with session manager enabled the resulting nodes were not getting the 
`topology.kubernetes.io/zone`  label. This meant that pods with affinity settings that relied on this label were never getting provisioned. I checked code and saw when fetching the tags used to populate this label from vsphere the function `withTagsClient` was returning early and not executing the user provided callback function

**Release note**:

```release-note
```


BernardMC:
Can this be cherry picked to 1.34 release? Otherwise that release will have a broken session manager client implementation

PR #1677, opened 2025-12-22
Helm: allow vsphere.conf to be sourced from hostPath

### LINKED ISSUE #1460: Provide option to use `hostPath` as a source for `vsphere.conf` for helm

Currently the source for `vsphere.conf` is `ConfigMap/Secret` but it is sometimes more convenient to store the content in the filesystem of the node. The pod can mount this path with `hostPath` option, but it needs to be configurable in the helm chart options. 

### Why is this needed?

This configuration option will streamline bootstrapping process for the component, when creation of ConfigMap or Secret can’t be automated, but the filesystem access is present, so a file can be created on the node first.

skools-here:
Adds an optional Helm configuration to allow vsphere.conf to be sourced from a hostPath volume.

This is useful for bootstrap and constrained environments where the configuration file already exists on the node and creating a ConfigMap/Secret is not desirable. The default behavior using a ConfigMap remains unchanged.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1460

The change is fully backward compatible (config.source=configMap remains the default).

When config.source=hostPath, the ConfigMap is not created and the DaemonSet mounts the file directly from the node filesystem.

Helm validation is included to prevent misconfiguration.

**Release note**:

```release-note
```

### DIFF
--- a/charts/vsphere-cpi/templates/configmap.yaml
+++ b/charts/vsphere-cpi/templates/configmap.yaml
@@ -1,4 +1,5 @@
 {{- $config := .Values.config -}}
+{{- if and $config.enabled (eq $config.source "configMap") }}
 {{- if .Values.global }}
 {{- if .Values.global.config }}
 {{- $config = mergeOverwrite (deepCopy .Values.config) .Values.global.config -}}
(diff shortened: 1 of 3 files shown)


skools-here:
Hey @chenlin07 , could you please take a look at this PR?
Kindly review it and let me know if any changes are needed..

PR #1717, opened 2026-03-25
Allow supervisor hostname override

100mik:
Allows folks to configure an override for the hardcoded hostname, this enables cases like allowing CPI to speak with the API server through a VCF A proxy.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1709

**Release note**:

```release-note
```

### DIFF
--- a/pkg/cloudprovider/vsphereparavirtual/config/config.go
+++ b/pkg/cloudprovider/vsphereparavirtual/config/config.go
@@ -49,6 +49,8 @@ const (
 	SupervisorServiceAccountNameEnv string = "SUPERVISOR_CLUSTER_SERVICEACCOUNT_SECRET_NAME"
 	// SupervisorAPIServerFQDN reads supervisor service API server's fully qualified domain name from env
 	SupervisorAPIServerFQDN string = "supervisor.default.svc"
+	// InternalSupervisorHostnameOverrideEnv allows overriding the default supervisor hostname
+	InternalSupervisorHostnameOverrideEnv string = "INTERNAL_SUPERVISOR_HOSTNAME_OVERRIDE"
 )
 
 // SupervisorEndpoint is the supervisor cluster endpoint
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vsphereparavirtual/config/config.go
```diff
@@ -131,8 +133,14 @@ func GetRestConfig(svConfigPath string) (*rest.Config, error) {
 		return nil, err
 	}
 
+	supervisorHostname := SupervisorAPIServerFQDN
```
DanielXiao, line 136: This looks like a back door,  I prefer to add hostName and port in CPI Config file. When they are empty, use SupervisorAPIServerFQDN as default.

DanielXiao (changes requested):

100mik:
Folks, I have updated the unit tests to cover the changes we introduced. I was unable to find E2E tests for the paravirtual mode I am afraid. Could use some guidance around how I can run them against a configured cluster.

Addressing other concerns:
1. We mount the desired certificate to the location where CCM expects it. So we do not need to configure it explicitly.
2. I have still stuck to using an env variable, but I have moved the logic closer to the other env variables we are using (for port, primarily). I cannot find existing fields in cloud config for these. I have however, grouped the logic better and extended the structure storing supervisor config to account for the hostname.

Let me know what you think @DanielXiao 🙏🏼

zhanggbj:
> Folks, I have updated the unit tests to cover the changes we introduced. I was unable to find E2E tests for the paravirtual mode I am afraid. Could use some guidance around how I can run them against a configured cluster.

@100mik Regarding E2E tests for paravirtual mode, we don't have automated E2E coverage set up since supervisor is not enabled in infra, for this change, you will need to test it manually against your own configured cluster/infrastructure.

PR #1749, opened 2026-05-11
✨ vsphereparavirtual: add support for VM Operator API version v1alpha6

silvery1622:
Adds a v1alpha6 VM Operator adapter .
Propagates IPFamilies/IPFamilyPolicy from the Kubernetes Service to VirtualMachineService for dual-stack load balancer support, and fixes toStatus to return all ingress entries instead of only the first.

Dual stack fields (ipFamilies, ipFamilyPolicy) are read/written via the unstructured API because the v1alpha6 Go types may lag the CRD schema.

**Release note**:

```
Added VM Operator API v1alpha6 support with dual stack load balancer propagation (`--vm-operator-api-version=v1alpha6`).
```


pkg/cloudprovider/vsphereparavirtual/vmoperator/provider/v1alpha6/client.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), vm); err != nil {
+		return nil, err
+	}
+	return vm, nil
+}
+
+// ListVirtualMachines lists VirtualMachines in the given namespace.
+// ResourceVersion="0" defaults to reading from the API server's watch cache.
+func (c *Client) ListVirtualMachines(ctx context.Context, namespace string, opts metav1.ListOptions) (*vmopv6.VirtualMachineList, error) {
+	if opts.ResourceVersion == "" {
```
DanielXiao: Do not add this option.

pkg/cloudprovider/vsphereparavirtual/loadbalancer.go
```diff
@@ -161,16 +161,17 @@ func (l *loadBalancer) EnsureLoadBalancerDeleted(ctx context.Context, clusterNam
...
+		if ing.IP == "" && ing.Hostname == "" {
+			continue
 		}
+		st.Ingress = append(st.Ingress, v1.LoadBalancerIngress{
+			IP:       ing.IP,
+			Hostname: ing.Hostname,
+		})
 	}
-	return &v1.LoadBalancerStatus{}
+	return st
```
DanielXiao, line 174: Original it return nil but it return empty struct now. What 's the implication?

pkg/cloudprovider/vsphereparavirtual/vmoperator/adapter/v1alpha6/dualstack_unstructured.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphereparavirtual/vmoperator/types"
+)
+
+// readDualStackFromUnstructured populates hub dual-stack fields from the raw object.
+// VM Operator v1alpha6 Go types may lag the CRD; these keys match core Service JSON.
+func readDualStackFromUnstructured(u map[string]interface{}, spec *types.VirtualMachineServiceSpec) {
+	families, found, err := unstructured.NestedStringSlice(u, "spec", "ipFamilies")
+	if err != nil {
+		klog.Warningf("readDualStackFromUnstructured: unexpected type for spec.ipFamilies: %v", err)
```
DanielXiao: Why not raise an error?

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
```diff
@@ -262,6 +268,12 @@ func (s *vmService) Update(ctx context.Context, service *v1.Service, clusterName
 	if !reflect.DeepEqual(existingAnnotations, desiredAnnotations) {
 		needsUpdate = true
 	}
+	if !reflect.DeepEqual(existingIPFams, desiredIPFams) {
```
DanielXiao, line 271: Is it possible existingIPFams or desiredIPFams to be nil or empty slice?

pkg/cloudprovider/vsphereparavirtual/loadbalancer.go
silvery1622: @DanielXiao This is a sharp observation!
The change a subtle but important distinction between returning nil and an empty struct.

In EnsureLoadBalancer, returning nil actually violates the cloudprovider interface contract. If a service is still provisioning and has no IPs yet, the controller expects a valid (but empty) status struct. Previously, the nil return could trigger a nil-pointer dereference downstream. Returning &v1.LoadBalancerStatus{} instead correctly signals that the LB exists but is still pending, keeping the EXTERNAL-IP in the expected <pending> state.

For GetLoadBalancer, it removes ambiguity. A nil return makes it hard to tell if the resource was missing or just unreadable. An empty struct explicitly confirms the resource exists, it just hasn't been assigned an IP yet.

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
silvery1622: Yes, both can be nil or an empty slice at the source. existingIPFams is normalized explicitly with if len(...) == 0 { = nil }. desiredIPFams is normalized implicitly inside cloneIPFamilies, which returns nil whenever len(f) == 0 — so it can never produce an empty non-nil slice. Both sides are always either nil or non-empty before the reflect.

PR #1751, opened 2026-05-11
✨ Add IPv6 and dual-stack support in CPI core

silvery1622:
Adds IPv6 and dual-stack support to the vSphere CPI.

- --cluster-ip-family (paravirtual): controls NodeInternalIP ordering for IPv4-only, IPv6-only, and dual-stack clusters. Non-IPv4 requires --vm-operator-api-version >= v1alpha6; IPv6 with route controller additionally requires --enable-vpc-mode=true.
- Report all interface addresses (deduped, link-local filtered), not just the primary IP.
- Route controller: per-family CR naming (<node> for IPv4, <node>-ipv6 for IPv6); fix silent error discard in DeleteRoute; fix ListRoutes → DeleteRoute double-suffix bug.

**Release note**:

### DIFF
--- a/pkg/cloudprovider/vsphere/nodemanager.go
+++ b/pkg/cloudprovider/vsphere/nodemanager.go
@@ -226,11 +226,6 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		return errors.New("VM Guest hostname is empty")
 	}
 
-	if len(oVM.Guest.Net) == 0 {
-		klog.V(4).Infof("oVM.Guest.Net is empty, skipping node discovery. This could be cauesd by vmtool not reporting correct IP address")
-		return errors.New("VM GuestNicInfo is empty")
-	}
-
 	tenantRef := vmDI.VcServer
 	if vmDI.TenantRef != "" {
 		tenantRef = vmDI.TenantRef
@@ -242,6 +237,11 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		ipFamilies = vcInstance.Cfg.IPFamilyPriority
 	} else {
 		klog.Warningf("Unable to find vcInstance for %s. Defaulting to ipv4.", tenantRef)
+	}
+
+	if len(oVM.Guest.Net) == 0 {
+		klog.V(4).Infof("oVM.Guest.Net is empty, skipping node discovery. This could be caused by vmtool not reporting correct IP address")
+		return errors.New("VM GuestNicInfo is empty")
 	}
 
 	var internalNetworkSubnets []*net.IPNet
@@ -349,7 +349,7 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 		if len(oVM.Guest.Net) > 0 {
 			if discoveredInternal == nil && discoveredExternal == nil {
 				klog.V(4).Infof("oVM.Guest.Net=%v", oVM.Guest.Net)
-				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamilies)
+				return fmt.Errorf("unable to find suitable IP address for node %s with IP family %s", nodeID, ipFamily)
 			}
 		}
 	}
--- a/pkg/cloudprovider/vsphereparavirtual/cluster_ip_family.go
+++ b/pkg/cloudprovider/vsphereparavirtual/cluster_ip_family.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vsphereparavirtual
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphereparavirtual/vmoperator/factory"
+)
+
+// ClusterIPFamilyIPv4 is the default: IPv4-first NodeInternalIP ordering; does
+// not require a VM Operator API with dual-stack VirtualMachineService fields.
+const ClusterIPFamilyIPv4 = "ipv4"
+
+// ClusterIPFamilyIPv6 is IPv6-first ordering and requires
+// --vm-operator-api-version >= v1alpha6 (dual-stack VirtualMachineService fields).
+const ClusterIPFamilyIPv6 = "ipv6"
+
+// ClusterIPFamilyIPv4IPv6 is dual-stack with IPv4 before IPv6 in the reported
+// NodeInternalIP list; requires --vm-operator-api-version >= v1alpha6.
+const ClusterIPFamilyIPv4IPv6 = "ipv4ipv6"
+
+// ClusterIPFamilyIPv6IPv4 is dual-stack with IPv6 before IPv4 (same primary
+// order as ClusterIPFamilyIPv6); requires --vm-operator-api-version >= v1alpha6.
+const ClusterIPFamilyIPv6IPv4 = "ipv6ipv4"
+
+// supportsDualStackVMService records, per --vm-operator-api-version, whether
+// that API persists the dual-stack fields on VirtualMachineService
+// (ipFamilies, ipFamilyPolicy). It must stay in sync with factory.NewAdapter:
+// every version accepted there must appear here. A new API version that
+// supports dual-stack VirtualMachineService fields should be added with the
+// value true; older or partial APIs should be added with false.
+// When adding a new supported version, add a case to TestVmopSupportsDualStackVMServiceAPI.
+var supportsDualStackVMService = map[string]bool{
+	factory.V1alpha2: false,
+	factory.V1alpha5: false,
+	factory.V1alpha6: true,
+}
+
+// ParseClusterIPFamily validates the --cluster-ip-family flag and returns a
+// canonical lowercase value: ipv4, ipv6, ipv4ipv6, or ipv6ipv4.
+func ParseClusterIPFamily(raw string) (string, error) {
+	s := strings.ToLower(strings.TrimSpace(raw))
+	switch s {
+	case "ipv4":
+		return ClusterIPFamilyIPv4, nil
+	case "ipv6":
+		return ClusterIPFamilyIPv6, nil
+	case "ipv4ipv6":
+		return ClusterIPFamilyIPv4IPv6, nil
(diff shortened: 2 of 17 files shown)


pkg/cloudprovider/vsphere/cloud.go
```diff
@@ -59,10 +60,20 @@ const (
 var (
 	logoutCh chan struct{}
 	logoutWG sync.WaitGroup
+
+	// relaxedIPValidation is set by --relaxed-ip-validation. When true,
+	// DiscoverNode logs a warning and continues when an IP family is not found
+	// in vCenter's Guest Info API rather than returning an error. This is
+	// needed in VKR environments where VMware Tools does not populate guest.Net.
+	relaxedIPValidation bool
```
DanielXiao: Why do we need this flag? Is my understanding correct? Please add the reason in comments.

- Raise an error if single stack
- Ignore it if dual stack

pkg/cloudprovider/vsphere/nodemanager.go
```diff
@@ -308,7 +333,11 @@ func (nm *NodeManager) DiscoverNode(nodeID string, searchBy cm.FindVM) error {
 	if len(nonLocalhostIPs) == 0 {
 		klog.V(4).Infof("nonLocalhostIPs is empty")
 		klog.V(4).Infof("oVM.Guest.Net=%v", oVM.Guest.Net)
-		return fmt.Errorf("unable to find suitable IP address for node after filtering out localhost IPs")
+		if nm.relaxedIPValidation {
+			klog.Warningf("No routable IP addresses found for node %s after filtering localhost IPs; continuing with no IP addresses (relaxed-ip-validation enabled)", nodeID)
```
DanielXiao: In this branch, will nonLocalhostIPs be nil or empty? What 's the sortedNonLocalhostIPs value after assignment `sortedNonLocalhostIPs, err := sortStaticallyConfiguredAddressesFirst(oVM.Config.ExtraConfig, nonLocalhostIPs)`?

pkg/cloudprovider/vsphereparavirtual/routemanager/staticroute/routemanager.go
```diff
@@ -62,8 +63,12 @@ func (sr *RouteManager) CreateCPRoutes(staticroutes helper.RouteCRList) ([]*clou
...
 		condition := GetRouteCRCondition(&(staticroute.Status), vpcapisv1.Ready)
 		if condition != nil && condition.Status == v1.ConditionTrue {
-			// one RouteSet per node, so we can use nodeName as the name of RouteSet CR
-			nodeName := staticroute.Name
+			// TargetNode must be the Kubernetes node name, not the CR name. For IPv6
+			// routes the CR carries a helper.SuffixIPv6 suffix; strip it so that the
+			// route controller can match the route to the correct node object and so
+			// that a subsequent DeleteRoute call can recompute the correct CR name via
+			// crNameForRoute without double-appending the suffix.
+			nodeName := strings.TrimSuffix(staticroute.Name, helper.SuffixIPv6)
```
DanielXiao: Is it possible that the ipv4 node name contains suffix `-ipv6`. Is it possible to use annotation or label to determine if this is the route for ipv4 or ipv6

pkg/cloudprovider/vsphereparavirtual/cluster_ip_family.go
```diff
@@ -0,0 +1,109 @@
...
+
+// ClusterIPFamilyIPv6IPv4 is dual-stack with IPv6 before IPv4 (same primary
+// order as ClusterIPFamilyIPv6); requires --vm-operator-api-version >= v1alpha6.
+const ClusterIPFamilyIPv6IPv4 = "ipv6ipv4"
+
+// vmopAPILevel assigns monotonic levels to supported --vm-operator-api-version
+// values. It must stay in sync with factory.NewAdapter: any version accepted there
+// that supports dual-stack VirtualMachineService fields needs level >= vmopAPILevelV1alpha6.
+// When adding a new supported version, add a case to TestVmopSupportsDualStackVMServiceAPI_levels.
+var vmopAPILevel = map[string]int{
```
DanielXiao: The integer levels (20, 50, 60) imply an ordering that future contributors must remember. A boolean supportsDualStack map field, or a typed enum, would convey intent more directly and remove the "must be >= 60" mental arithmetic.

pkg/cloudprovider/vsphereparavirtual/instances.go
```diff
@@ -76,35 +81,116 @@ func (i instances) discoverNodeByName(ctx context.Context, name types.NodeName)
...
+	// Append primary IPs in cluster-preference order so the API server's first
+	// address matches the cluster's intended IP family.
+	ipv6First := false
+	switch clusterIPFamily {
+	case ClusterIPFamilyIPv6, ClusterIPFamilyIPv6IPv4:
+		ipv6First = true
+	case ClusterIPFamilyIPv4, ClusterIPFamilyIPv4IPv6:
+		ipv6First = false
+	default:
+		// Invariant: callers must pass a value returned by ParseClusterIPFamily.
```
DanielXiao: Why don't raise an error?

pkg/cloudprovider/vsphereparavirtual/route.go
```diff
@@ -193,15 +209,28 @@ func (r *routesProvider) getNodeIPAddress(nodeName string, isIPv4 bool) (string,
...
 	}
+
+	// Find the first IP matching the requested family
+	ipFamily := "IPv6"
+	if isIPv4 {
+		ipFamily = "IPv4"
+	}
+
 	for _, ip := range allIPs {
 		if (ip.To4() != nil) == isIPv4 {
```
DanielXiao: A bit confusing here. Are you looking for IPv4 only? How about dual stack?

pkg/cloudprovider/vsphereparavirtual/instances.go
```diff
@@ -76,35 +81,116 @@ func (i instances) discoverNodeByName(ctx context.Context, name types.NodeName)
...
+				Address: vm.PrimaryIP6,
+			})
+			addedIPs[vm.PrimaryIP6] = true
+		}
 	}
 
-	return []v1.NodeAddress{
-		{
+	for _, ip := range vm.NetworkInterfaceAddresses {
+		if addedIPs[ip] {
```
DanielXiao: Two equivalent IPv6 addresses with different textual representations (2001:db8::1 vs 2001:0db8::0001) will both be added. Prefer normalizing via net.ParseIP(ip).String() before keying.

pkg/cloudprovider/vsphere/cloud.go
silvery1622: For VKR environments where VMware Tools may not yet populate IPv6 entries in guest.Net. Without it, an IPv6 /dual stack cluster running on VKR cannot register any node that lacks an IPv6 address.
I have addressed your comments.
The flag is now scoped:
  - IPv4-only cluster:  flag has NO effect; IP discovery failures still error out.
  - IPv6 single stack: flag downgrades failures to warnings.
  - In the dual stack case's per-family loop, the warning is further restricted to the ipv6 family. A missing IPv4 in a dual stack cluster still errors.

pkg/cloudprovider/vsphereparavirtual/routemanager/staticroute/routemanager.go
silvery1622: Thanks. Fixed this by inspecting the CIDR family and only trim the suffix when there is IPv6 CIDR defined.

pkg/cloudprovider/vsphereparavirtual/cluster_ip_family.go
silvery1622: Thanks. I have update the code to use a bool map.

pkg/cloudprovider/vsphereparavirtual/instances.go
silvery1622: Good catch. Fixed this.

pkg/cloudprovider/vsphere/nodemanager.go
silvery1622: nonLocalhostIPs is nil (not empty): filter() declares its accumulator as `var filtered []*ipAddrNetworkName` (zero value), so when nothing matches it returns nil. sortStaticallyConfiguredAddressesFirst passes nil through unchanged when guestInfo is empty. Ranging over a nil slice is a Go no-op, so the per-family loop runs zero iterations without panic.

That said, the fall-through is now only reachable for IPv6/dual-stack clusters:

  relax := nm.relaxedIPValidation && hasIPv6Family(ipFamilies)

For IPv4-only clusters the function returns an error before reaching sortStaticallyConfiguredAddressesFirst, regardless of the flag.

pkg/cloudprovider/vsphereparavirtual/instances.go
silvery1622: Switched to panic with an explicit invariant message.

The default branch is unreachable in normal operation: Initialize() in cloud.go already klog.Fatalf's on a non-canonical --cluster-ip-family before NewInstances is called. So hitting the default means a programming error in a future caller, not a recoverable runtime condition.

Returning an error would change the function signature and ripple into NodeAddresses / NodeAddressesByProviderID. panic keeps the signature unchanged and fails fast — much safer than the previous silent fallback to IPv4-first, which on a dual-stack cluster could route traffic to the wrong family and be hard to diagnose.

pkg/cloudprovider/vsphereparavirtual/route.go
silvery1622: This function intentionally handles a single family per call. Dual stack is handled one layer above by the Kubernetes route controller, which iterates Node.Spec.PodCIDRs and calls CreateRoute once per CIDR:

  CreateRoute(route)
    crName = crNameForRoute(node, route.DestinationCIDR)   // bare name for IPv4, "-ipv6" suffix for IPv6
    nodeIP = getNodeIPAddress(node, util.IsIPv4(route.DestinationCIDR))

For a dual stack node with PodCIDRs = ["10.244.0.0/24", "2001:db8::/64"], CreateRoute is invoked twice: once with wantIPv4=true (CR `node`, IPv4 next-hop) and once with wantIPv4=false (CR `node-ipv6`, IPv6 next-hop). Each call to getNodeIPAddress picks the first registered node address matching the requested family.

Updated the function doc to spell this out explicitly.

pkg/cloudprovider/vsphere/cloud.go
DanielXiao: Can you reminder me why we need to loose validation on IPv6 or dual stack? If an error raised, what happens? Note down the reason in the comment.

pkg/cloudprovider/vsphereparavirtual/routemanager/staticroute/routemanager.go
DanielXiao: Is it possible to use label to identify IPv4 or IPv6 rather by staticroute name? For example:

cluster=foo
node=foo-worker-1
ipfamily=ipv4

Labels can be used in get and list verb.

silvery1622: Thanks for your comment. I will fix this in a forthcoming PR that adds support for IPv6/dual-stack routable pods.

DanielXiao (approved):

PR #1763, opened 2026-05-18
✨  Support IPv6/dual-stack routable pods with NSX-VPC

silvery1622:

- New ipfamily package centralises the four --cluster-ip-family values and their boolean helpers (IPv4Enabled, IPv6Enabled, DualStack, PrimaryIPv4). 
- NSXVPCIPManager creates/deletes one IPAddressAllocation CR per enabled IP family per node (IPv4 keeps the bare node name; IPv6 appends -ipv6). Both CRs carry nodeName and ipfamily labels so downstream consumers never need to parse CR names. 
- ipaddressallocation_controller waits for both partner allocations to be ready, then issues a single atomic PodCIDRs patch ordered by primary family. A resync idempotency guard skips redundant PATCHes when the node already has the correct CIDRs. 
- StaticRoute CRs gain nodeName/ipfamily labels; ListRoutes prefers labels over name-suffix parsing (legacy fallback retained). 
- StartControllers now takes an Options struct to keep the signature stable as new flags are added.

Testing Verification

- Validated the implementation on a vSphere Supervisor environment by deploying four workload clusters covering all IP family configurations (IPv4, IPv6, IPv4/IPv6 Dual-Stack, and IPv6/IPv4 Dual-Stack).
- CR Verification: Verified that IPAddressAllocation and StaticRoute CRs were correctly created, labeled, and populated with proper subnets/next-hops for both single-stack and dual-stack nodes.
- Datapath & Connectivity: Deployed an Nginx echo test application across the clusters. Confirmed successful end-to-end pod connectivity via direct curl requests across different address families (IPv4 and IPv6).
- Lifecycle Management: Validated clean teardown. Deleting the workload clusters successfully cleaned up all associated IPAddressAllocation and StaticRoute resources with no leaking CRs.

**Release note**:

```release-note
Add IPv6 and dual-stack routable pod support for NSX-VPC mode. Set --cluster-ip-family=ipv6, ipv4ipv6, or ipv6ipv4 together with --enable-vpc-mode=true to enable per-family IPAddressAllocation and StaticRoute CRs.
```


pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+}
+
+// processIPAddressAllocationCreateOrUpdate gets the CIDR from the IPAddressAllocation
+// status and patches it to the matching Node's PodCIDRs.
+//
+// In dual stack mode it waits for both the IPv4 and IPv6 partner allocation to
+// be ready before performing a single atomic patch with both CIDRs ordered by
+// primaryIPFamilyIsIPv4.
+func (c *Controller) processIPAddressAllocationCreateOrUpdate(ctx context.Context, ipAddressAllocation *vpcapisv1.IPAddressAllocation) error {
+	nodeName := nodeNameForAllocation(ipAddressAllocation)
```
DanielXiao, line 310: Can we put an annotation of node name when node controller create IPAddressAllocation?

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+	if thisCIDR == "" {
+		return fmt.Errorf("IPAddressAllocation %v does not have an allocated CIDR yet", ipAddressAllocation.Name)
+	}
+
+	node, err := c.nodesLister.Get(nodeName)
+	if err != nil {
+		return err
+	}
+
+	if !c.ipFamily.DualStack() {
```
DanielXiao, line 319: Is ipFamily set per Cluster by addon manager according to Cluster POD/SERVICE cidr ipfamily?

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+		if len(node.Spec.PodCIDRs) == 1 && node.Spec.PodCIDRs[0] == thisCIDR {
```
DanielXiao: What if len(node.Spec.PodCIDRs) > 1 or node.Spec.PodCIDRs[0] != thisCIDR?

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+			klog.V(4).Infof("Node %s already has correct PodCIDR %s, skipping patch", nodeName, thisCIDR)
+			return nil
+		}
+		return utils.PatchNodeCIDRWithRetry(ctx, c.kubeclientset, node, []string{thisCIDR}, c.recorder)
 	}
-	node, err := c.nodesLister.Get(ipAddressAllocation.Name)
+
+	// Dual stack: look up the partner allocation; both must be ready before patching.
+	partnerName := partnerCRNameFor(nodeName, thisIsIPv4)
```
DanielXiao: Let 's put a IPFamily label on IPAddressAllocation to determine it 's IPv4 or IPv6. Do not use name to determine it.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
 	}
+	partnerCIDR := allocationIP(partner)
+	if partnerCIDR == "" {
+		return fmt.Errorf("partner IPAddressAllocation %s does not have an allocated CIDR yet", partnerName)
+	}
 
-	// update node with allocated podCIDR
-	return utils.PatchNodeCIDRWithRetry(ctx, c.kubeclientset, node, podCIDR, c.recorder)
+	// Map (this, partner) into (ipv4, ipv6) then order by primary family.
+	ipv4CIDR, ipv6CIDR := thisCIDR, partnerCIDR
```
DanielXiao: Why does IPv4 must be the primary? Is IPv6,IPv4 valid?

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +254,133 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+	if !thisIsIPv4 {
+		ipv4CIDR, ipv6CIDR = partnerCIDR, thisCIDR
+	}
+	cidrs := []string{ipv4CIDR, ipv6CIDR}
+	if !c.ipFamily.PrimaryIPv4() {
```
DanielXiao: The logic in this method is not straightforward. My understanding is like this:
```
IPAddressAllocations := list AddressAllocation by label cluster=coffee nodename=coffee-worker-1
switch c.ipFamily:
case dual-stack:
   if len(IPAddressAllocations) !=2 
      raise error
   check IPAddressAllocations[].ipFamily one is IPv4 and the other is IPv6
   if c.ipFamily.PrimaryIPv4():
      sort IPAddressAllocations with IPv4 first
   else
      sort IPAddressAllocations with IPv6 first
case IPv4:
    if len(IPAddressAllocations) !=1 
       raise error
    check IPAddressAllocations[].ipFamily is IPv4
case IPv6:
    if len(IPAddressAllocations) !=1 
       raise error
    check IPAddressAllocations[].ipFamily is IPv6

CIDRs := IPAddressAllocations[].status.CIDRs
patch CIDRs to Node
```

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: Yes, it is. Addon manager would be responsible for parsing Cluster Pod/Service CIDRs and set the IPFamily info via argument.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: There is already label for node name upon IPAddressAllocation creation which cab be used for query. Do you think we need both label and annotation?

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: If the node has multiple CIDRs or an incorrect CIDR in a single-stack cluster, that if condition evaluates to false. The code then falls through to utils.PatchNodeCIDRWithRetry, which uses a strategic merge patch to completely overwrite both PodCIDR and PodCIDRs with exactly []string{thisCIDR}.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: Will use the existing isIPv4Allocation(alloc) helper (which checks the Spec) instead of parsing names.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: The ipv4CIDR, ipv6CIDR variables were just local variables to keep track of which CIDR was which. Right below that, the code explicitly checked if !c.ipFamily.PrimaryIPv4() and swapped the slice to []string{ipv6CIDR, ipv4CIDR} before patching the node. So IPv6,IPv4 was completely valid and supported.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: Sure, will use the suggested logic.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +255,142 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+//
+// Identification is taken from Spec.IPAddressType, which is the authoritative
+// source set by the creator. For backward compatibility with pre-dual-stack
+// CRs (which left IPAddressType unset and were always single-stack IPv4), an
+// empty type also resolves to IPv4 — matching the upstream API default.
+//
+// We deliberately do NOT parse the CR name to decide the family: node names
+// that themselves end in "-ipv6" would otherwise be misclassified.
+func isIPv4Allocation(alloc *vpcapisv1.IPAddressAllocation) bool {
+	return alloc.Spec.IPAddressType != vpcapisv1.IPAllocationIPAddressTypeIPv6
```
DanielXiao, line 273: According to method name, `alloc.Spec.IPAddressType == vpcapisv1.IPAllocationIPAddressTypeIPv4` looks better

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +255,142 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+
+		cidr := allocationIP(ipAddressAllocation)
+		if cidr == "" {
+			return fmt.Errorf("IPAddressAllocation %s does not have an allocated CIDR yet", ipAddressAllocation.Name)
+		}
+		cidrs = []string{cidr}
+	} else {
+		// Dual-stack: list all allocations for this node.
+		// Dual-stack clusters are new and guaranteed to have the nodeName label on their CRs.
+		selector := labels.SelectorFromSet(labels.Set{helper.LabelKeyNodeName: nodeName})
```
DanielXiao: Should we also add label for cluster name or vsphere cluster name? There could be same nodeName in different clusters, for example, below 2 clusters all render node name coffee-worker-beijing 
cluster coffee nodepool worker-beijing
cluster coffee-worker nodepool beijing

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
```diff
@@ -23,76 +53,182 @@ type NSXVPCIPManager struct {
...
+	}
+
 	ipAddressAllocation := &vpcapisv1.IPAddressAllocation{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
+			Name:      crName,
 			Namespace: m.svNamespace,
+			Labels: map[string]string{
+				helper.LabelKeyNodeName: nodeName,
+				helper.LabelKeyIPFamily: familyLabel,
```
DanielXiao: The spec.IPAddressType contains ipfamily so I don't think this label is used anywhere. We need to add a label for VSphereCluster

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
```diff
@@ -23,76 +53,182 @@ type NSXVPCIPManager struct {
...
 
-	klog.V(4).Infof("Node %s already requested IPAddressAllocations", node.Name)
+	lister := m.informerFactory.Crd().V1alpha1().IPAddressAllocations().Lister().IPAddressAllocations(m.svNamespace)
+
+	// TODO(danqing.hou): thread context from NSXIPManager.ClaimPodCIDR once the
+	// interface is refactored to accept a context parameter.
+	ctx := context.TODO()
+
+	for _, ipv4 := range families {
+		crName := crNameForFamily(node.Name, ipv4)
```
DanielXiao, line 165: We should use label to get IPAddressAnnotation rather than name.

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
```diff
@@ -23,76 +53,182 @@ type NSXVPCIPManager struct {
...
 	}
+	return errors.Join(errs...)
+}
 
-	klog.V(4).Infof("Deleting IPAddressAllocations %s/%s", m.svNamespace, node.Name)
-	return m.client.CrdV1alpha1().IPAddressAllocations(m.svNamespace).Delete(context.Background(), node.Name, metav1.DeleteOptions{})
+// deleteIPAddressAllocation deletes the IPAddressAllocation CR for the given node and
+// IP family. NotFound is treated as success (idempotent delete).
+func (m *NSXVPCIPManager) deleteIPAddressAllocation(ctx context.Context, nodeName string, ipv4 bool) error {
+	crName := crNameForFamily(nodeName, ipv4)
```
DanielXiao, line 201: same as above

pkg/cloudprovider/vsphereparavirtual/routemanager/staticroute/routemanager.go
```diff
@@ -64,14 +63,17 @@ func (sr *RouteManager) CreateCPRoutes(staticroutes helper.RouteCRList) ([]*clou
...
-			nodeName := staticroute.Name
-			if !util.IsIPv4(staticroute.Spec.Network) {
-				nodeName = strings.TrimSuffix(staticroute.Name, helper.SuffixIPv6)
+			// Prefer the nodeName label written by CreateRoute (Phase 2+).
+			// Fall back to the CR name for legacy CRs that lack the label.
+			// In the fallback path we only strip the IPv6 suffix when the
+			// destination is actually IPv6, so an IPv4 route on a node whose
+			// name happens to end in "-ipv6" is not silently truncated.
+			nodeName := staticroute.Labels[helper.LabelKeyNodeName]
+			if nodeName == "" {
```
DanielXiao, line 70: Raise an error till label is added by node controller.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
```diff
@@ -247,27 +255,142 @@ func (c *Controller) syncHandler(ctx context.Context, key string) error {
...
+// existed in dual-stack mode before this change, so the fallback is only
+// reached for legacy single-stack IPv4 CRs whose names are the bare node
+// name and so the strip is a no-op.
+func nodeNameForAllocation(alloc *vpcapisv1.IPAddressAllocation) string {
+	if name, ok := alloc.Labels[helper.LabelKeyNodeName]; ok && name != "" {
+		return name
 	}
-	if podCIDR == "" {
-		return fmt.Errorf("IPAddressAllocation %v does not get CIDR allocated", ipAddressAllocation.Name)
+	return helper.StripFamilySuffix(alloc.Name)
```
DanielXiao: Do not fallback, Raise an error till label is added by node controller.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: We cannot use == vpcapisv1.IPAllocationIPAddressTypeIPv4 because of backward compatibility. Older CRs created before dual stack support have IPAddressType unset (empty string). By checking != IPv6, we correctly classify both explicitly set IPv4 CRs and older empty-type CRs as IPv4, which matches the upstream API default.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: Thanks for the comment. I have updated the code and add clusterName into IPAddressAllocation's label. 
I also passed clusterName into IPAddressAllocation Controller and matches both nodeName and clusterName when doing label selector.

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
silvery1622: Agreed. I have replaced the IPFamily label with clusterName label.

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
silvery1622: We need continue using Get(crName) to ensure backward compatibility for existing IPv4 single stack clusters during upgrades, as their CRs lack the new labels. Since dual stack is greenfield only, the deterministic CR name crNameForFamily(nodeName, ipv4) is guaranteed to be unique and safe to use without label filtering.

pkg/cloudprovider/vsphereparavirtual/nsxipmanager/nsx_vpc.go
silvery1622: We need continue using Get(crName) to ensure backward compatibility for existing IPv4 single stack clusters during upgrades, as their CRs lack the new labels. Since dual stack is greenfield only, the deterministic CR name crNameForFamily(nodeName, ipv4) is guaranteed to be unique and safe to use without label filtering.

pkg/cloudprovider/vsphereparavirtual/routemanager/staticroute/routemanager.go
silvery1622: We cannot raise an error here because it would break route syncing for existing IPv4 single stack clusters during upgrades, as their legacy CRs lack the label. However, since dual stack is greenfield only, any CR without the label is guaranteed to be a legacy IPv4 CR. I have simplified the code to safely fall back to the CR name directly.

pkg/cloudprovider/vsphereparavirtual/controllers/routablepod/ipaddressallocation/ipaddressallocation_controller.go
silvery1622: Similar to StaticRoute, raising an error would break backward compatibility for existing IPv4 single stack clusters. Since dual stack is greenfield only, any CR lacking the label is guaranteed to be a legacy IPv4 CR. I have simplified the fallback logic to just return the CR name directly.

DanielXiao (approved):

PR #1780, opened 2026-06-05
✨Support token reload without restarting

### LINKED ISSUE #1779: vsphere-paravirtual CCM should not restarts the pod on every projected SA token rotation

Let client-go handle Supervisor service-account token reloading instead of restarting the CCM pod whenever the token file changes. Concretely, two coordinated changes:

In config.GetRestConfig (pkg/cloudprovider/vsphereparavirtual/config/config.go), stop reading the token into rest.Config.BearerToken (a static string) and instead set rest.Config.BearerTokenFile to the token path (<svConfigPath>/token). This lets client-go's transport own token reload: when BearerTokenFile is set, requests are wrapped with NewBearerAuthWithRefreshRoundTripper → NewCachedFileTokenSource, which re-reads the token file periodically (~1 min period, 10s leeway) and on 401s, picking up rotated tokens transparently.

In initializeWatch (cmd/vsphere-cloud-controller-manager/main.go), remove the projected service-account token from the fsnotify watch list (drop SupervisorServiceAccountPath, or at minimum exclude the token file) so a token update no longer triggers klog.Fatalf and a pod restart.

Note: these two changes must land together. Removing the token from the watch without switching to BearerTokenFile would leave the CCM authenticating with a stale token after the first rotation.
The CA cert (ca.crt) is also read once into TLSClientConfig.CAData. If CA rotation is expected, that path still warrant a watch/restart 

### Why is this needed?

The current implementation injects a token that never expires, so today the file rarely (if ever) changes and the restart-on-change behavior is effectively dormant. We are planning to optimize this by moving to a short-lived token that is periodically refreshed. Once that lands, the token file will be rewritten on every refresh.

With the current code, that creates a problem: GetRestConfig reads the token once into the static BearerToken field, which client-go never reloads. The only mechanism that picks up a new token is the fsnotify watch, which restarts the pod. That means every token refresh would restart the CCM, causing leader-election loss, controller resync, and transient downtime — all for a routine, expected rotation.

Switching to BearerTokenFile makes client-go reload the token in place (this is exactly the mechanism designed for kubelet-rotated projected SA tokens), so the short-lived token rollout becomes a no-op for pod lifecycle: no restarts, no leader-election flaps, no churn.

12345lcr:
In paravirtual mode, the CCM watches the mounted supervisor credentials and restarts the pod on any change. Since the credentials are delivered as a single projected Secret volume, routine token rotation triggered an unnecessary restart, even though client-go can reload the token on its own.

- Configures the supervisor rest.Config with BearerTokenFile instead of an inlined static BearerToken, so client-go transparently reloads the (short-lived, periodically refreshed) token from disk.
- Reworks the file watcher to skip restarts on token-only rotations. Because the projected volume updates atomically via a ..data symlink swap (fsnotify never names individual keys), the decision is content-based: a baseline hash of the restart-worthy keys (ca.crt, namespace) is captured at startup and re-checked on each event. Changes to those keys—or any change outside the credentials mount—still trigger a restart; an unreadable key fails safe by restarting.
- Adds unit tests for the watcher decision logic (token-only rotation, ca.crt/namespace changes, missing key, chmod, out-of-mount events) and updates the rest-config test for BearerTokenFile.

Backward compatible with the current non-expiring token and forward compatible with the planned short-lived token.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1779 

Manual Verification Complete

Step 1: Feature Deployment
- Update the CPI deployment to utilize the specific build containing the short-lived token feature.

Step 2: Basic Functionality Validation
- Provision a sample unprivileged LoadBalancer service deployment to verify normal end-to-end routing behavior.

Step 3: Resource Generation and Connectivity Verification
- Verify that the guest cluster LoadBalancer service successfully receives a valid external IP.
- Confirm that the corresponding VirtualMachineService resource is successfully generated in the Supervisor namespace.
- Validate traffic flow via a curl test against the external IP, ensuring the expected payload is returned.

Step 4: Token Rotation and Refresh Testing
- Manually delete the ProviderServiceAccount 's corresponding service account on the Supervisor plane to simulate a token rotation event.
- Monitor the system and verify that the resource successfully regenerates and issues a refreshed token.

Step 5: Temporary Disruption Observation
- During the transient window between service account deletion and recreation, the CPI pod expectedly reported brief Unauthorized API errors.

Note: This represents the expected systemic delay before service account regeneration.

Step 6: Recovery Validation
- Once the newly refreshed token synchronized down to the guest cluster, the CPI pod successfully authenticated, and all Unauthorized errors ceased.
- Concluded that recovery occurs gracefully without requiring a pod restart.

Step 7: Post-Rotation Lifecycle Verification
- Cycle the sample LoadBalancer service (delete and recreate) to ensure operational continuity post-rotation.
- Observed the expected asynchronous event loop timing (initial VirtualMachineService IP not found warnings while the infrastructure was provisioning), followed quickly by a successful EnsuredLoadBalancer status confirmation.

Step 8: Pod Restart Testing
- Manually modify the token secret's ca.crt field in the guest cluster to an invalid value to simulate a corrupted state.
- Confirm that the CPI pod automatically triggers a restart loop as expected when the ca.crt field change is detected.

Step 9: Environmental Recovery
- Revert the token secret's ca.crt field back to its original valid value.
- Confirm that the CPI pod restarts, successfully recovers its authenticated state, and returns to a healthy Running status.

**Release note**:

```release-note
```

### DIFF
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -20,7 +20,7 @@ limitations under the License.
 package main
 
 import (
-	"crypto/sha256"
+	"bytes"
 	"flag"
 	goflag "flag"
 	"fmt"
@@ -330,18 +330,18 @@ type restartGuard struct {
 	// volume. It is empty when not running in paravirtual mode, in which case
 	// every non-Chmod event triggers a restart.
 	secretMountDir string
-	// baselines maps a restart-worthy file name to the content hash captured at
+	// baselines maps a restart-worthy file name to the raw content captured at
 	// startup.
-	baselines map[string][sha256.Size]byte
+	baselines map[string][]byte
 }
 
-// newRestartGuard captures the startup content hashes of the restart-worthy
+// newRestartGuard captures the startup raw contents of the restart-worthy
 // keys under secretMountDir. Pass an empty secretMountDir to restart on every
 // (non-Chmod) event.
 func newRestartGuard(secretMountDir string, restartKeys []string) *restartGuard {
 	g := &restartGuard{
 		secretMountDir: secretMountDir,
-		baselines:      make(map[string][sha256.Size]byte, len(restartKeys)),
+		baselines:      make(map[string][]byte, len(restartKeys)),
 	}
 	for _, k := range restartKeys {
 		data, err := os.ReadFile(filepath.Join(secretMountDir, k))
@@ -352,7 +352,7 @@ func newRestartGuard(secretMountDir string, restartKeys []string) *restartGuard 
 			klog.Warningf("unable to read baseline for %s under %s: %v", k, secretMountDir, err)
 			continue
 		}
-		g.baselines[k] = sha256.Sum256(data)
+		g.baselines[k] = data
 	}
 	return g
 }
@@ -382,7 +382,7 @@ func (g *restartGuard) restartKeyChanged() bool {
 			klog.Warningf("failed to re-read %s under %s, triggering restart: %v", name, g.secretMountDir, err)
 			return true
 		}
-		if sha256.Sum256(got) != want {
+		if !bytes.Equal(got, want) {
 			klog.Infof("detected change in %s, triggering restart", name)
 			return true
 		}
--- a/cmd/vsphere-cloud-controller-manager/main.go
+++ b/cmd/vsphere-cloud-controller-manager/main.go
@@ -20,12 +20,14 @@ limitations under the License.
 package main
 
 import (
+	"crypto/sha256"
 	"flag"
 	goflag "flag"
 	"fmt"
 	"math/rand"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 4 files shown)


cmd/vsphere-cloud-controller-manager/main.go
```diff
@@ -300,6 +311,96 @@ func initializeWatch(paths []string) (watch *fsnotify.Watcher, stopCh chan struc
...
+// them differs from the baseline captured at startup. A read failure is treated
+// as a change so we fail safe by restarting.
+func (g *restartGuard) restartKeyChanged() bool {
+	for name, want := range g.baselines {
+		got, err := os.ReadFile(filepath.Join(g.secretMountDir, name))
+		if err != nil {
+			klog.Warningf("failed to re-read %s under %s, triggering restart: %v", name, g.secretMountDir, err)
+			return true
+		}
+		if sha256.Sum256(got) != want {
```
zhanggbj: The files being checked (ca.crt and namespace) are small (usually less than 1.5 KB total), storing them as raw []byte is completely fine and allows comparing them using standard bytes.Equal, suggest to remove the dependency on "crypto/sha256".


In https://github.com/kubernetes/apiserver/blob/master/pkg/server/dynamiccertificates/dynamic_cafile_content.go#L132-L149, it is using `bytes.Equal` to detect CA changes.

zhanggbj (changes requested):

cmd/vsphere-cloud-controller-manager/main.go
12345lcr: Got it, updated. Thanks

zhanggbj:
Thanks for the PR!
/lgtm
/approve

zhanggbj (approved):

PR #1784, opened 2026-06-09
✨Remove the use of the cryptographically weak MD5 hash

### LINKED ISSUE #1783: Remove the use of the cryptographically weak MD5 hash (crypto/md5)

Remove the use of the cryptographically weak MD5 hash (crypto/md5) from the vSphere Paravirtual CPI's VirtualMachineService naming logic in pkg/cloudprovider/vsphereparavirtual/vmservice.

Currently, GetVMServiceName derives a VirtualMachineService resource name by computing an MD5 hash of <service.Name>.<service.Namespace> and truncating it. This should be replaced with a non-cryptographic hash (e.g., FNV-128a, SHA-256) for new resources.

Because existing clusters already have VirtualMachineService resources named with the old MD5-based scheme, the change must remain backward compatible. When a resource is not found under the new name, the CPI should locate the existing (legacy-named) resource via the identifying labels it has always stamped on every VirtualMachineService (run.tanzu.vmware.com/cluster.name, run.tanzu.vmware.com/service.name, run.tanzu.vmware.com/service.namespace) rather than recomputing the old MD5 name. This keeps Get/Update/Delete working against pre-upgrade resources without orphaning or duplicating them, and it allows crypto/md5 to be dropped entirely.

```
- apiGroups: ["vmoperator.vmware.com"]
  resources: ["virtualmachineservices"]
  verbs: ["get", "list", "create", "update", "delete"]  # "list" is the new requirement
```

Note: the label-based fallback requires the list verb on virtualmachineservices in the supervisor namespace; the supervisor-side RBAC for the CPI must grant it.

### Why is this needed?

The FIPS140 security scan flagged the codebase with "Usage of SHA1 and MD5 detected." MD5 is a broken hash algorithm, and its presence in the source/binary causes the CPI to fail security/compliance gates, even though it is used here only to derive a deterministic resource name and not for any security purpose.

Suppressing the finding (e.g., with #nosec) is unreliable: the repo's standard linting is golint and the image scanning is Trivy (which does not perform Go weak-crypto source analysis), so the finding originates from a separate SAST tool whose suppression behavior we can't depend on. The only robust way to clear it is to remove MD5 from the code path entirely.

Doing this without breaking backward compatibility is essential: existing clusters have live VirtualMachineService resources named under the old MD5 scheme. Without a compatibility fallback, upgrading the cloud-provider-vsphere would cause the CPI to no longer find those resources, leading to duplicate VirtualMachineService creation, orphaned load balancers, and disrupted LB-type Services on upgrade.

12345lcr:
Remove the use of the cryptographically weak MD5 hash (crypto/md5)

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1783 

Test Verification Plan & Results

1. Baseline Setup: Provisioned a guest cluster using the paravirtual CPI without these changes applied.
2. Workload Deployment: Created a sample NGINX deployment and a LoadBalancer (LB) service on the guest cluster.
3. Supervisor Validation: Confirmed the sample service successfully initialized on the guest cluster, and verified that the corresponding VirtualMachineService and Service resources were correctly created on the Supervisor cluster.
4. Traffic Verification: Verified endpoint connectivity by executing a curl command against the LoadBalancer IP to confirm it was healthy.
5. Upgrade Trigger: Upgraded the CPI component to the build generated by this PR.
6. Post-Upgrade Stability: Confirmed that no duplicate or additional Service or VirtualMachineService resources were generated on the Supervisor. Verified existing resources continued to run as expected with no disruptions or changes to the LoadBalancer IP.
7. Teardown Validation: Deleted the sample NGINX service on the guest cluster and verified that the associated Service and VirtualMachineService resources on the Supervisor were cleanly deleted.
8. Migration Verification: Recreated the NGINX service on the guest cluster and confirmed that the newly generated Service and VirtualMachineService resources correctly adopted the new naming pattern.

**Release note**:

```release-note
Replaced MD5 with SHA-256 for VirtualMachineService names with a backward-compatible label fallback; upgrading requires adding the list verb to the CPI's virtualmachineservices RBAC.
```

### DIFF
--- a/pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
+++ b/pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
@@ -18,7 +18,7 @@ package vmservice
 
 import (
 	"context"
-	"crypto/md5" // #nosec
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"reflect"
@@ -29,6 +29,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	vmop "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphereparavirtual/vmoperator"
 	vmoptypes "k8s.io/cloud-provider-vsphere/pkg/cloudprovider/vsphereparavirtual/vmoperator/types"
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
```diff
@@ -106,15 +107,53 @@ func NewVMService(vmClient vmop.Interface, ns string, ownerRef *metav1.OwnerRefe
...
+
+	list, err := s.vmClient.VirtualMachineServices().List(ctx, s.namespace, vmoptypes.ListOptions{LabelSelector: selector})
+	if err != nil {
+		logger.Error(err, "failed to list VirtualMachineServices for legacy lookup")
+		return nil, err
+	}
+	if len(list) == 0 {
+		return nil, nil
+	}
+	if len(list) > 1 {
```
zhanggbj, line 152: In steady state, the labels should represent a 1:1 logical mapping under steady conditions, but during state transitions or failures or race conditions, the list can return multiple duplicates. However, simply returning list[0] is non-deterministic and may have risks.

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
```diff
@@ -106,15 +107,53 @@ func NewVMService(vmClient vmop.Interface, ns string, ownerRef *metav1.OwnerRefe
 }
 
 func (s *vmService) hashString(str string) string {
-	// #nosec
-	hash := md5.New()
+	// SHA-256 is used as a FIPS-approved, well-distributed hash to derive a
```
zhanggbj: nit:
```suggestion
	// SHA-256 is used as a FIPS-compliant, well-distributed hash to derive a
```

zhanggbj (changes requested):

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
12345lcr: Got it, updated to return an error instead.

pkg/cloudprovider/vsphereparavirtual/vmservice/vmservice.go
12345lcr: Updated, thanks

zhanggbj:
@12345lcr 
LGTM. Please help to add the release note in the PR, then we could merge it. Also highlight the RBAC and ClusterRole Requirements Change.

12345lcr:
> @12345lcr LGTM. Please help to add the release note in the PR, then we could merge it. Also highlight the RBAC and ClusterRole Requirements Change.

Sure, updated. Thanks.

Issue #1862, opened 2026-08-30
read InstanceType from instance metadata


k0da:
### What would you like to be added?

Have an ability to provide custom InstanceType rather than generating vsphere-vm.cpu-<>.mem-<>gb.os-<>.

CPI should read custom attribute of a vm or lookup for particular tag

### Why is this needed?

In karpenter provider for [VSphere](github.com/AbsaOSS/karpenter-provider-vsphere) we provide cloud like experience for onprem users. 

More friendly instance type would improve users understanding  instance options to choose when scheduling kubernetes workload

PR #1868, opened 2026-09-04
fix: parse instance types from VM config's annotation (#1)

dihlorfos:
* fix: parse instance types from VM config's annotation

Instance type must be parsed from VM config's annotation first.
If it is present and non-empty, it is used as NodeType; otherwise it fallsback to the existing vsphere-vm.cpu-%d.mem-%dgb.os-%s construction.  This fixes issue #1862.

**Release note**:

```release-note
NONE
```


zhanggbj:
Hi @dihlorfos,
Thanks for the PR, please sign the required CLA.

